PR #12, opened 2021-02-03
[HOTT-371] Quotas removal for XI service

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-371

### What?

Effectively removes quota functionality visibility for XI users

- [x] Quotas search page returns a 404 for XI
- [x] Quotas tools page link only shown for UK users
- [x] Quotas search page disables switching to the XI service

### Why?

I am doing this because:

- Quotas coming from the EU are currently disabled for the XI service so no data is shown

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,6 +5,7 @@ class ApplicationController < ActionController::Base
   include TradeTariffFrontend::ViewContext::Controller
   include ApplicationHelper
 
+  before_action :set_enable_service_switch_banner_in_action
   before_action :set_last_updated
   before_action :set_cache
   before_action :search_query
@@ -59,6 +60,10 @@ class ApplicationController < ActionController::Base
 
   def set_last_updated
     @tariff_last_updated ||= TariffUpdate.latest_applied_import_date
+  end
+
+  def set_enable_service_switch_banner_in_action
+    @enable_service_switch_banner_in_action = true
   end
 
   def search_invoked?
--- a/app/views/pages/tools.html.erb
+++ b/app/views/pages/tools.html.erb
@@ -14,7 +14,7 @@
     <h1 class="govuk-heading-xl govuk-!-margin-bottom-5">Tariff tools</h1>
     <%= service_switch_banner %>
     <ul class="govuk-list">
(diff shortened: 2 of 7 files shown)


spec/helpers/service_helper_spec.rb
```diff
@@ -114,19 +114,31 @@
...
+      assign(:enable_service_switch_banner_in_action, true)
+      allow(helper).to receive(:request).and_return(request)
     end
 
     context 'when on xi sections page' do
       let(:request) { double('request', filtered_path: '/xi/sections') }
 
       it 'returns the full banner that allows users to toggle between the services' do
-        expect(service_switch_banner).to include(t("service_banner.big.#{choice}", link: switch_service_link))
+        expect(helper.service_switch_banner).to include(t("service_banner.big.#{choice}", link: helper.switch_service_link))
```
willfish, line 125: If you don't invoke the helper and just rely on method_missing to send the message to the helper we don't invoke the assigns that scope the instance variables inside of the view. This was actually quite painful to debug and is clearly a bug in RSpec.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,7 @@ class ApplicationController < ActionController::Base
   include TradeTariffFrontend::ViewContext::Controller
   include ApplicationHelper
 
+  before_action :set_enable_service_switch_banner_in_action
```
willfish, line 8: Opted for globally setting the view instance variable to true rather than checking if the value is nil in the helper as this felt more obvious and didn't require treating nil (the default value of an instance variable that hasn't been set) as truthy. Happy to just treat nil as truthy in the helper and not introduce this hook if your prefer.

app/views/pages/tools.html.erb
```diff
@@ -14,13 +14,15 @@
...
-      </li>
+      <% if TradeTariffFrontend::ServiceChooser.uk? %>
+        <li>
+          <p>
+            <%= link_to t('breadcrumb.quotas'), quota_search_path %>
+            <br>
+            Search for tariff quotas, including daily updated balances.
+          </p>
+        </li>
+      <% end %>
```
tavyy, line 25: Could this be a decorator?

tavyy (approved):

PR #14, opened 2021-02-10
Fix static service links

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Fixes issues with static service links. We need to make sure that the typeahead search for geographical areas is pulling on the correct urls and preserving the current service
- [x] Fixes exclusion of GB for xi and uk service

### DIFF
--- a/app/models/geographical_area.rb
+++ b/app/models/geographical_area.rb
@@ -10,10 +10,15 @@ class GeographicalArea
   has_many :children_geographical_areas, class_name: 'GeographicalArea'
 
   def self.countries
-    excluded_geographical_area_ids = %w[GB]
+    excluded_geographical_area_ids =
+      if TradeTariffFrontend::ServiceChooser.xi?
+        []
+      else
+        %w[GB]
+      end
 
     all.sort_by(&:id)
-       .reject { |country| country.id.in?(excluded_geographical_area_ids) }
+      .reject { |country| country.id.in?(excluded_geographical_area_ids) }
   end
 
   def self.cached_countries
--- a/app/models/geographical_area.rb
+++ b/app/models/geographical_area.rb
@@ -3,7 +3,7 @@ require 'api_entity'
 class GeographicalArea
   include ApiEntity
 
-  collection_path '/geographical_areas/countries'
+  collection_path "/geographical_areas/countries"
 
   attr_accessor :id, :description, :geographical_area_id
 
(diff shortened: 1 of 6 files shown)


app/models/geographical_area.rb
```diff
@@ -3,46 +3,51 @@
...
   has_many :children_geographical_areas, class_name: 'GeographicalArea'
 
   def self.countries
-    excluded_geographical_area_ids = %w[GB]
+    excluded_geographical_area_ids =
+      if TradeTariffFrontend::ServiceChooser.xi?
+        []
+      else
+        %w[GB]
+      end
```
tavyy, line 18: This could be:

```ruby
excluded_geographical_area_ids = TradeTariffFrontend::ServiceChooser.xi? ? [] : %w[GB]
```

app/models/geographical_area.rb
```diff
@@ -3,46 +3,51 @@
...
 
     all.sort_by(&:id)
-       .reject { |country| country.id.in?(excluded_geographical_area_ids) }
+      .reject { |country| country.id.in?(excluded_geographical_area_ids) }
```
tavyy, line 21: or even better:

```ruby
dataset.reject { |country| country.id.in?(%w[GB]) } if TradeTariffFrontend::ServiceChooser.xi?
```

app/views/commodities/show.html.erb
```diff
@@ -34,3 +34,8 @@
 <% end %>
 
 <%= render 'declarables/declarable', declarable: @commodity %>
+
+<%= javascript_tag do %>
+  window.geographicalAreasPath = "<%= geographical_areas_path(format: :json) %>"
+  window.searchSuggestionsPath = "<%= search_suggestions_path(format: :json) %>"
+<% end %>
```
tavyy, line 41: Inline JS is not recommended.

app/views/search/quota_search.html.erb
```diff
@@ -42,3 +42,7 @@
     <h1 class="govuk-heading-l">There are no matching results</h1>
   </article>
 <% end %>
+
+<%= javascript_tag do %>
+  window.quotaSearchPath = "<%= quota_search_path(format: :json) %>"
```
tavyy, line 47: Inline JS is not recommended.

tavyy (approved):
Given the urgency I will approve but we need to refactor some bits tomorrow.

app/models/geographical_area.rb
willfish: I actually really don't like ternary operators.

PR #15, opened 2021-02-10
[HOTT-426] Make javascript links respect the service choice

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-426

### What?

I have added/removed/altered:

- [x] Forces the javascript links to respect the service choice
- [x] Extends the excluded_geographical_areas array logic to only exclude GB in the UK platform

### Why?

I am doing this because:

- GB only makes sense to be excluded as a country in the UK platform
- We plan on renaming GB's description in XI and hiding the XU geographical areas since GB has all of the preferences we want XU to have
- We want to make sure we don't get the wrong service in our calls to get geographical areas or commodities

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,9 +5,11 @@ class ApplicationController < ActionController::Base
   include TradeTariffFrontend::ViewContext::Controller
   include ApplicationHelper
 
+  before_action :set_cache
   before_action :set_enable_service_switch_banner_in_action
   before_action :set_last_updated
-  before_action :set_cache
+  before_action :set_path_info
+
   before_action :search_query
   before_action :maintenance_mode_if_active
   before_action :bots_no_index_if_historical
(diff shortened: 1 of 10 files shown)


app/controllers/application_controller.rb
```diff
@@ -5,9 +5,11 @@ class ApplicationController < ActionController::Base
   include TradeTariffFrontend::ViewContext::Controller
   include ApplicationHelper
 
+  before_action :set_cache
```
willfish, line 8: I'm alphabetically sorting these setters. Order won't matter for the Cache-Control header specified in the request I don't think

willfish:
When we're happy with this Octavian I'll squash the commits and give a better commit message when I merge

app/webpacker/src/javascripts/commodities.js.erb
```diff
@@ -746,7 +746,7 @@ import debounce from "./debounce";
                 source: debounce(function(query, populateResults) {
                   $.ajax({
                     type: "GET",
-                    url: window.geographicalAreasPath,
+                    url: $(".path_info").data("geographicalAreasPath"),
```
tavyy, line 749: The only comment would be to use a proper api endpoint here. Will the data attrib grow here?

tavyy (approved):

app/webpacker/src/javascripts/commodities.js.erb
willfish: Shouldn't grow beyond the extent of our current needs. I think it'll actually get removed. Not sure what you mean by a proper api endpoint? Maybe something we can discuss in person another time

PR #18, opened 2021-02-11
Temporarily comment sentry step

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Comment out sentry release until we have time to fix it


tavyy:
Can you add a TODO note on each file?

tavyy (approved):

willfish:
I don't think we need to? It's an extra 25 minutes of waiting :(

tavyy:
Fair fair

PR #23, opened 2021-03-12
[HOTT-463] Enable routing to exchange rates path in backend

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-463

### What?

I have added/removed/altered:

- [x] Extends the list of api resources that are routable via the frontend
- [x] Lints the routes and TradeTariffFrontend files
- [x] Adds a .gitignore entry for my editor

### Why?

I am doing this because:

We have added an exchange rate api for the duty calculator. Everything routes via the request forwarder in the frontend (for now) and it is configured to be strict about which resources are accessible.

### DIFF
--- a/lib/trade_tariff_frontend.rb
+++ b/lib/trade_tariff_frontend.rb
@@ -8,15 +8,41 @@ module TradeTariffFrontend
 
   # API Endpoints of the Tariff Backend API app that can be reached via Frontend
   def accessible_api_endpoints
-    %w[sections chapters headings commodities updates monetary_exchange_rates quotas
-       goods_nomenclatures search_references geographical_areas]
+    %w[
+      sections
+      chapters
+      headings
+      commodities
+      updates
+      monetary_exchange_rates
+      quotas
+      goods_nomenclatures
+      search_references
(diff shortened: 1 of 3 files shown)


lib/trade_tariff_frontend.rb
```diff
@@ -8,15 +8,41 @@ module TradeTariffFrontend
...
+      chapters
+      headings
+      commodities
+      updates
+      monetary_exchange_rates
+      quotas
+      goods_nomenclatures
+      search_references
+      geographical_areas
+    ]
```
tavyy, line 22: 👍

lib/trade_tariff_frontend.rb
```diff
@@ -8,15 +8,41 @@ module TradeTariffFrontend
...
+      search_references
+      additional_codes
+      certificates
+      footnotes
+      geographical_areas
+      chemicals
+      additional_code_types
+      certificate_types
+      footnote_types
+    ]
```
tavyy, line 45: 👍

tavyy (approved):

PR #24, opened 2021-03-15
Deploy production

mafonso:
### Jira link

HOTT-442

### What?

I have added/removed/altered:

- [ ] Allow deployment to production on manual gate.

### Why?

I am doing this because:
- Implementing workflow defined on a team chat.
-
-

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,8 +65,9 @@ commands:
             # Send "real" url to new version
             cf unmap-route "$CF_APP-<< parameters.environment_key >>-dark" london.cloudapps.digital -n "$CF_APP-<< parameters.environment_key >>-dark"
             cf map-route  "$CF_APP-<< parameters.environment_key >>-dark" london.cloudapps.digital -n "$CF_APP-<< parameters.environment_key >>"
-            cf map-route  "$CF_APP-<< parameters.environment_key >>-dark" "<< parameters.domain_prefix >>".trade-tariff.service.gov.uk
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -140,50 +142,64 @@ jobs:
...
           domain_prefix: "staging"
 
+  deploy_production:
+    docker:
+      - image: cimg/ruby:2.7-node
+    steps:
+      - cf_deploy:
+          space: "production"
+          environment_key: "production"
+          domain_prefix: ""
```
willfish: Isn't the domain prefix www?

mafonso: It is  yes!  🤦 
Which means I need to find another way of doing this.   `www`  vs `assets`

willfish:
I think we should remove the hold on deploying to staging since we've got a filter for branches == master

willfish:
We work a lot to make sure what we've merged has been reviewed, understood and tested and merged branches are highly stable

.circleci/config.yml
mafonso: To be fair, this is something that I wanted to have a look at. It does not make sense to me that the origin needs to know what's its external name. 
I know why it's happening, but I think it should not be like this. The origin can and should be agnostic, IMO.

mafonso:
> I think we should remove the hold on deploying to staging since we've got a filter for branches == master

Is that still in? Can't find it...

.circleci/config.yml
willfish: Yeah

willfish:
No it's not! Is my brain fart :)

willfish (approved):

PR #25, opened 2021-03-15
Removes pdf functionality

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Removes all PDF functionality from the frontend

### Why?

I am doing this because:

- We've removed the functionality from the UKTT gem
- We've removed the functionality from the backend


lib/trade_tariff_frontend.rb
```diff
@@ -93,10 +93,6 @@ def regulations_enabled?
     ENV.fetch('HIDE_REGULATIONS') != 'true'
   end
 
-  def download_pdf_enabled?
-    ENV.fetch('DOWNLOAD_PDF_ENABLED', 'false') == 'true'
```
tavyy, line 97: Will you remove it from the PaaS as well?

willfish: As in the env vars? Yeah , absolutely :)

tavyy (approved):
Nice

PR #27, opened 2021-03-16
Remove separate assets hostname

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Removes asset host configuration

### Why?

I am doing this because:

- We want to simplify our deployment configuration

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -77,9 +77,6 @@ Rails.application.configure do
     pool_size:  Integer(ENV['MAX_THREADS'] || 5)
   })
 
-  # Enable serving of images, stylesheets, and JavaScripts from an asset server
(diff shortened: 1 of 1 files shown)


config/environments/production.rb
```diff
@@ -77,9 +77,6 @@
     pool_size:  Integer(ENV['MAX_THREADS'] || 5)
   })
 
-  # Enable serving of images, stylesheets, and JavaScripts from an asset server
-  config.action_controller.asset_host = ENV['GOVUK_ASSET_ROOT']
```
willfish, line 81: This resolves to the same host so we don't need it

mafonso (approved):
LGTM

PR #28, opened 2021-03-16
Removes unused assets domain from ci deployment

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Removes now-unused subdomain for separate assets

### Why?

I am doing this because:

- This only ever resolves to the same domain as $HOST so is adding complexity without value


mafonso (approved):
LGTM

PR #29, opened 2021-03-18
Removes caching of geographical areas from the frontend

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Removes duplicating cache from the frontend for geographical areas

### Why?

I am doing this because:

- These are now cached in the backend for both the duty calculator and the
frontend


app/forms/quota_search_form.rb
```diff
@@ -33,7 +33,7 @@ def large_result?
   end
 
   def geographical_area
-    GeographicalArea.countries.find do |country|
+    GeographicalArea.all.find do |country|
```
tavyy, line 36: So we are not looking just in countries anymore?

willfish: GeographicalArea.all does call the `/geographical_areas/countries` endpoint bizarrely

tavyy (approved):

PR #30, opened 2021-03-19
Disable cdn caching of search results

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-451

### What?

I have added/removed/altered:

- [x] Adds before action in search controller that disables cdn caching
- [x] Adds test of Cache-Control header for SearchController
- [x] Refactor cache spec for the ApplicationController

### Why?

I am doing this because:

- This is costing us money without adding any value since they're cached once and often aren't reused

### DIFF
--- a/spec/controllers/search_controller_spec.rb
+++ b/spec/controllers/search_controller_spec.rb
@@ -3,7 +3,7 @@ require 'spec_helper'
 describe SearchController, 'GET to #search', type: :controller do
   include CrawlerCommons
 
-  controller(described_class) do
+  controller described_class do
     def index
       render plain: 'Hari Seldon'
     end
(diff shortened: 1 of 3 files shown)


spec/controllers/search_controller_spec.rb
```diff
@@ -3,6 +3,18 @@
 describe SearchController, 'GET to #search', type: :controller do
   include CrawlerCommons
 
+  controller(described_class) do
```
willfish:
```ruby
        # If you would like to spec a subclass of ApplicationController, call
        # controller like so:
        #
        #     controller(ApplicationControllerSubclass) do
        #       # ....
        #     end
```

willfish:
```suggestion
  controller described_class do
```

tavyy (approved):

PR #51, opened 2021-04-11
Handle commodity that has meursing codes for dc

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-486

### What?

I have added/removed/altered:

- [x] Hide link to duty calculator if a commodity has meursing code

### Why?

I am doing this because:

- This is needed for go live

### DIFF
--- a/app/models/concerns/declarable.rb
+++ b/app/models/concerns/declarable.rb
@@ -25,6 +25,10 @@ module Declarable
     meursing_code
   end
 
+  def no_meursing?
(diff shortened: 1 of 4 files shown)


willfish:
### Northern Ireland - with meursing

Screenshot from 2021-04-12 08-12-47

### United Kingdom - without meursing

Screenshot from 2021-04-12 08-13-01

tavyy (approved):

PR #68, opened 2021-04-28
Adds requirement getter method to force html_safe the requirement

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-647

### What?

I have added/removed/altered:

- [x] Adds a getter for the requirement attribute of a measure condition so that we can render the html embedded within the requirement string.
- [x] Adds a test for the new getter to validate the requirement is html safe

We trust that the data source is safe since the origin of the data is
one of the Taric (European) or CDS (HMRC/UK) databases

### Why?

I am doing this because:

- This is needed so that we can safely render the condition requirement
that comes from the json api resource in the backend
- This has been requested by a couple of stakeholders that use the tariff service

### DIFF
--- a/app/models/measure_condition.rb
+++ b/app/models/measure_condition.rb
@@ -3,5 +3,10 @@ require 'api_entity'
 class MeasureCondition
   include ApiEntity
 
-  attr_accessor :condition_code, :condition, :document_code, :requirement, :action, :duty_expression
+  attr_accessor :condition_code, :condition, :document_code, :action, :duty_expression
+  attr_writer :requirement
+
+  def requirement
+    @requirement&.html_safe
+  end
 end
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -131,10 +131,12 @@ FactoryBot.define do
   end
 
   factory :measure_condition do
-    document_code { Forgery(:basic).text }
-    action { Forgery(:basic).text }
+    condition_code { Forgery(:basic).text(exactly: 1) }
     condition { Forgery(:basic).text }
-    requirement_type { Forgery(:basic).text }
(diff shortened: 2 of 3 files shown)


app/models/measure_condition.rb
```diff
@@ -3,5 +3,10 @@
 class MeasureCondition
   include ApiEntity
 
-  attr_accessor :condition_code, :condition, :document_code, :requirement, :action, :duty_expression
+  attr_accessor :condition_code, :condition, :document_code, :action, :duty_expression
+  attr_writer :requirement
```
willfish, line 7: Having the requirement attribute as an accessor whilst duplicating the getter was triggering a rubocop violation so I've tidied this up to reflect that by adding the `attr_writer` here

app/models/measure_condition.rb
```diff
@@ -3,5 +3,10 @@
...
+
+  def requirement
+    @requirement.html_safe
```
willfish: We've talked about the extensive use of html_safe in this code base and agreed that there is a risk to users of the service if any of our downstream import processes are compromised. We should do a review of where we use html_safe and make sure we're sanitising script tags.

spec/factories.rb
```diff
@@ -131,10 +131,12 @@
   end
 
   factory :measure_condition do
```
willfish, line 133: This was just broken and couldn't be used so I've taken an example measure condition returned from the api and recreated this based on what form that took

app/models/measure_condition.rb
tavyy: Yeah, there's always that risk.

tavyy (approved):

PR #71, opened 2021-05-07
HOTT-265: On the chapter page, include both the chapter and heading number in the list of headings

krios82:
On the chapter page, include both the chapter and heading number in the list of headings

### Jira link

https://transformuk.atlassian.net/browse/HOTT-265

### What?

I have added/removed/altered:

- [x] Add chapter number in the list of headings
- [x] Add specs

### Why?

I am doing this because:

Traders could be confused by the omission of the chapter id from the list of headings


tavyy (approved):
Nice!

PR #82, opened 2021-05-18
Group measures by the following types

tavyy:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Group measures by the following types:

- VAT/EXCISE
- QUOTAS
- TRADE REMEDIES
- IMPORT CONTROLS

On the XI service, fetch the UK measures of type:
- EXCISE/VAT
- IMPORT UK CONTROLS

Only display the groups that are available for each service.

### Why?

I am doing this because:

- Business request


app/views/measures/_measures.html.erb
```diff
@@ -86,35 +86,14 @@
     <%= render partial: 'shared/country_picker', locals: { controller: @search, url: perform_search_path, anchor: 'import' } %>
 
     <% if declarable.import_measures.for_country(@search.country).any? %>
+      <%= render partial:'measures/grouped/navigation', locals: { uk_declarable: uk_declarable, xi_declarable: xi_declarable } %>
```
willfish:
```suggestion
      <%= render partial: 'measures/grouped/navigation', locals: { uk_declarable: uk_declarable, xi_declarable: xi_declarable } %>
```

app/views/measures/grouped/_xi_navigation.html.erb
```diff
@@ -0,0 +1,27 @@
...
+  <li class="govuk-!-margin-bottom-2">
+    <%= link_to('EU import controls', '#xi_import_controls', class: 'govuk-link') %>
+  </li>
+<% end %>
+<% if uk_declarable.import_measures.import_controls.for_country(@search.country).present? %>
+  <li class="govuk-!-margin-bottom-2">
+    <%= link_to('UK import controls', '#uk_import_controls', class: 'govuk-link') %>
+  </li>
+<% end %>
+
```
willfish, line 26: File termination char going crazzyy

willfish (approved):

app/models/measure.rb
```diff
@@ -95,4 +115,10 @@ def additional_code_sort
       additional_code.code.to_s
     end
   end
+
+  private
+
+  def declarable_types
+    @declarable_types ||= YAML.load_file('config/declarable_types.yml')
+  end
```
krios82, line 123: Maybe I'm wrong, but I think this will look it up every time the controller is reloaded (once per call).

Why not putting it into a custom configuration that will be loaded at startup, since this list will only change on deploy? (specifically using `Rails::Application.config_for` )

https://edgeguides.rubyonrails.org/configuring.html#custom-configuration

tavyy: Yep, we can do that!

config/initializers/cookies_serializer.rb
```diff
@@ -2,4 +2,4 @@
 
 # Specify a serializer for the signed and encrypted cookie jars.
 # Valid options are :json, :marshal, and :hybrid.
-Rails.application.config.action_dispatch.cookies_serializer = :hybrid
+Rails.application.config.action_dispatch.cookies_serializer = :json
```
tavyy: This was reverted in my last commit.

willfish (approved):

krios82 (approved):
LGTM

PR #85, opened 2021-05-19
HOTT-659: no error on landing

krios82:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-659

### What?

I have added/removed/altered:

- [x] Make sure the error message doesn't show when landing on the page the first time
- [x] Fill the date form with today's date if left blank

### Why?

I am doing this because:

- The error message was showing at the first load of the page


app/forms/quota_search_form.rb
```diff
@@ -23,6 +23,18 @@ def page
     @page || 1
   end
 
+  def year
+    @year || DEFAULT_YEAR_VALUE
```
willfish: We need to make sure this is dynamically worked out since a class constant is initialised when the application starts and we only deploy like once a week

willfish (approved):

PR #86, opened 2021-05-19
Run Cypress tests on CI

mafonso:
### Jira link

HOTT-692

### What?

I have added/removed/altered:

- [ ] Run Cypress smoke tests against staging on deployment to staging

### Why?

I am doing this because:

- We want to run a quick regression pack when we deploy to staging
-
-

### Have you? (optional)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -122,9 +122,9 @@ jobs:
     steps:
       - checkout
       - ruby/install-deps
-      - ruby/rubocop-check:
-          format: progress
-          label: Inspecting with Rubocop
+#      - ruby/rubocop-check:
+#          format: progress
+#          label: Inspecting with Rubocop
       - run:
           name: Inspecting with Brakeman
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -114,16 +114,17 @@ jobs:
...
     resource_class: small
     steps:
       - checkout
       - ruby/install-deps
-      - ruby/rubocop-check:
-          format: progress
-          label: Inspecting with Rubocop
+#      - ruby/rubocop-check:
+#          format: progress
+#          label: Inspecting with Rubocop
```
tavyy, line 127: Leaving Rubocop linting out?

mafonso: The agreement on the convo that followed in slack, is that we're better leaving these out for now, until we can address the linting violations, so that the test is more useful than being red every time until then.

willfish (approved):

PR #91, opened 2021-05-25
Bump ruby to latest patch version

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Bumps ruby

### Why?

I am doing this because:

- This makes the version we use in ci, in deployed applications and locally all consistent


mafonso (approved):
LGTM

PR #94, opened 2021-05-25
Fixes brakeman complaints

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Stop using permit! in application controller
- [x] Use json cookie serializer configuration


tavyy (approved):
I guess we can trust the unit tests!

PR #98, opened 2021-05-26
HOTT-696: Remove items of type "Commodity" from the A-Z

krios82:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-696

### What?

I have added/removed/altered:

- [ ] Filter out items of type commodity from the a-z list
- [ ] Added some specs
- [ ] Regenerated the relevant vcr file

### Why?

I am doing this because the ticket required the removal of commodities from the list.

There were two ways to do this: 

1. Filtering out commodities in the frontend using a select, which is slightly more expensive.
2. Add the relevant logic to the backend so that the filtering could be done using a query.

Considering that the backend endpoint doesn't currently allow filtering by `referenced_class` (or anything more than "letter" anyway) I considered that this functionality wasn't needed in the backend and went with solution 1), happy to change this if I'm mistaken.

### Have you? (optional)


app/controllers/search_references_controller.rb
```diff
@@ -3,6 +3,7 @@ def show
     @no_shared_switch_service_link = true
     @search_references = SearchReferencesPresenter.new(
       SearchReference.all(query: { letter: letter })
+                     .select { |ref| ref.attributes['referenced_class'] != 'Commodity' }
```
tavyy: Nice you could also do:

```ruby
.reject { |ref| ref.attributes['referenced_class']  ==  'Commodity'}
```

spec/controllers/search_references_controller_spec.rb
```diff
@@ -27,4 +27,9 @@
     expect(response.body).to include 'Machine Tools'
     expect(response.body).to include '/sections/16'
   end
+
+  it 'doesn\'t renders links to commodities' do
+    expect(response.body).not_to include 'Mascarpone'
+    expect(response.body).not_to include '/commodities/0406105090'
+  end
```
tavyy, line 54: I know it's a pattern in this codebase, but I think having multiple expectations inside the same block is not recommended. Did Rubocop not complain about this?

tavyy (approved):

willfish (approved):

willfish:
If you rebase the brakeman failures will disappear

PR #99, opened 2021-05-26
Quota changes

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-712

### What?

I have added/removed/altered:

- [x] Updates quota order number view to match new content requirements
- [x] Adds helper methods for quota measures and specs
- [x] Adds new Time format that is gov compliant

Screenshot from 2021-05-26 14-29-40

### Why?

I am doing this because:

- This is a business requirement

### DIFF
--- a/app/helpers/measures_helper.rb
+++ b/app/helpers/measures_helper.rb
@@ -1,12 +1,41 @@
 module MeasuresHelper
-  def order_number_info(record)
-    record.definition.try(:description) || "Order number #{record.number}"
+  def show_warning?(order_number)
+    order_number.definition.description && order_number.number.start_with?('05')
+  end
+
+  def start_and_end_dates_for(definition)
+    start_date = definition.validity_start_date
+    end_date = definition.validity_end_date
+
+    "#{start_date.to_formatted_s(:gov)} to #{end_date.to_formatted_s(:gov)}"
+  end
+
+  def suspension_period_dates_for(definition)
+    if definition.suspension_period_start_date.present? && definition.suspension_period_end_date.present?
+      start_date = definition.suspension_period_start_date
+      end_date = definition.suspension_period_end_date
+
+      "#{start_date.to_formatted_s(:gov)} to #{end_date.to_formatted_s(:gov)}"
+    else
+      'n/a'
+    end
+  end
+
+  def blocking_period_dates_for(definition)
+    if definition.blocking_period_start_date.present? && definition.blocking_period_end_date.present?
+      start_date = definition.blocking_period_start_date
+      end_date = definition.blocking_period_end_date
+
+      "#{start_date.to_formatted_s(:gov)} to #{end_date.to_formatted_s(:gov)}"
+    else
+      'n/a'
+    end
   end
 
   def filter_duty_expression(measure)
-    record = measure.duty_expression.to_s.html_safe
-    record = '' if record == 'NIHIL'
-    record = 'see conditions' if measure.measure_type.id.in? %w(DDA DDJ)
-    record
+    order_number = measure.duty_expression.to_s.html_safe
+    order_number = '' if order_number == 'NIHIL'
+    order_number = 'see conditions' if measure.measure_type.id.in? %w[DDA DDJ]
+    order_number
   end
 end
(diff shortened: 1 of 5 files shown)


app/helpers/measures_helper.rb
```diff
@@ -1,12 +1,41 @@
 module MeasuresHelper
-  def order_number_info(record)
-    record.definition.try(:description) || "Order number #{record.number}"
+  def show_warning?(order_number)
+    order_number.definition.description && order_number.number.start_with?('05')
```
krios82, line 3: Are we sure the `order_number.definition` and `order_number.number` will never be nil?

app/views/order_numbers/_order_number.html.erb
```diff
@@ -1,63 +1,55 @@
...
+            <th scope="col" class='govuk-table__header'>Blocking period</th>
+            <td class="numerical govuk-table__cell numerical"><%= blocking_period_dates_for(order_number.definition) %></td>
           </tr>
         </tbody>
       </table>
-      <p class="govuk-!-margin-top-4">In an effort to facilitate UK importers HMRC endeavours to keep quotas 'OPEN' where possible, therefore the information displayed above may differ slightly from the current UK position. If you require further information contact the Tariff Quota Management Team by email: <%= mail_to("tariff.management@hmrc.gsi.gov.uk", nil, subject: "Order number #{order_number.number}") %>.</p>
+      <p class="govuk-!-margin-top-4">In an effort to facilitate UK importers HMRC endeavours to keep quotas 'OPEN' where possible, therefore the information displayed above may differ slightly from the current UK position. If you require further information contact the Tariff Quota Management Team by email: <%= mail_to("tariff.managment@hmrc.gov.uk", nil, subject: "Order number #{order_number.number}") %>.</p>
```
krios82: The `managment` part of this address seems misspelled.
I've seen it's also misspelled in the ticket, is the official address that is misspelled or it's a typo from Matt?

app/helpers/measures_helper.rb
tavyy: Yeah, good point I think we need to guard against those cases.

app/helpers/measures_helper.rb
```diff
@@ -1,12 +1,41 @@
...
+
+  def suspension_period_dates_for(definition)
+    if definition.suspension_period_start_date.present? && definition.suspension_period_end_date.present?
+      start_date = definition.suspension_period_start_date
+      end_date = definition.suspension_period_end_date
+
+      "#{start_date.to_formatted_s(:gov)} to #{end_date.to_formatted_s(:gov)}"
+    else
+      'n/a'
+    end
```
tavyy, line 21: Fancy using:

```ruby
 return 'n/a' unless foo

 bar
```
?

Not a blocker.

app/helpers/measures_helper.rb
willfish: No no, we do in the view already so are already being defensive

app/views/order_numbers/_order_number.html.erb
willfish: I clarified with Matt and this should have the e in it. This is the spelling in the ticket

willfish:
```suggestion
      <p class="govuk-!-margin-top-4">In an effort to facilitate UK importers HMRC endeavours to keep quotas 'OPEN' where possible, therefore the information displayed above may differ slightly from the current UK position. If you require further information contact the Tariff Quota Management Team by email: <%= mail_to("tariff.management@hmrc.gov.uk", nil, subject: "Order number #{order_number.number}") %>.</p>
```

app/helpers/measures_helper.rb
willfish: I did originally and found it less readable :thinking:

willfish: I think rust and elixir have ruined me

app/helpers/measures_helper.rb
willfish: It's not possible for the number to be nil

krios82 (approved):

PR #102, opened 2021-06-01
HOTT-343: Presentational bug on commodity code pages - ghost tier in the breadcrumb

krios82:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-343

### What?

I have added/removed/altered:

- [ ] Remove the extra `li` tag that was causing the issue
- [ ] Fixed most of the linting issues in the page

### Why?

I am doing this because there was a presentational bug that needed to be fixed

### Have you? (optional)

### DIFF
--- a/app/helpers/commodities_helper.rb
+++ b/app/helpers/commodities_helper.rb
@@ -15,16 +15,16 @@ module CommoditiesHelper
 
   def commodity_tree(main_commodity, commodities)
     if commodities.any?
-      content_tag(:ul, class: 'commodities') do
-        content_tag(:li) do
-          tree_code(commodities.first.code) + 
-          content_tag(:p, commodities.first.to_s.html_safe) +
-          tree_node(main_commodity, commodities, commodities.first.number_indents)
+      tag.ul(class: 'commodities') do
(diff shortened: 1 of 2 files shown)


app/helpers/commodities_helper.rb
```diff
@@ -15,16 +15,16 @@ def commodity_level(commodity)
...
+          tree_code(commodities.first.code) +
+            tag.p(commodities.first.to_s.html_safe) +
+            tree_node(main_commodity, commodities, commodities.first.number_indents)
         end
       end
     else
-      content_tag(:ul, class: 'commodities') do
-        content_tag(:li, commodity_heading(main_commodity))
+      tag.ul(class: 'commodities') do
+        commodity_heading(main_commodity)
```
krios82, line 27: This is the fix: remove the external `li` tag: the `commodity_heading` function already returns a `li` wrapper, and this "external" `li` tag was displayed above the other since HTML doesn't allow nesting a `li` inside another `li` directly.

The rest are Rubocop fixes.

willfish (approved):
Seems reasonable to me

PR #109, opened 2021-06-03
Consolidates popups for the quota definition and order numbers

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-719

### What?

I have added/removed/altered:

- [x] Consolidate the quota popups in the measures table and the quota search results

### Why?

I am doing this because:

- We want these two views of quota definitions to look the same

#### Measures popup

Screenshot from 2021-06-03 15-43-25

#### Quota results popup

Screenshot from 2021-06-03 15-43-18


spec/factories.rb
```diff
@@ -63,12 +63,12 @@
     suspension_period_end_date { nil }
     blocking_period_start_date { nil }
     blocking_period_end_date { nil }
+    order_number { attributes_for(:order_number) }
   end
 
   factory :measure do
     transient do
       measure_type_description { Forgery(:basic).text }
-      ordernumber { Forgery(:basic).number(exactly: 5) }
```
krios82, line 71: Why removing this? 
Will measure mocks never be used on their own and always be initialised as children of something else that will pass down this value?

krios82 (approved):

spec/factories.rb
willfish: So this is just an unused transient attribute. Probably someone was tempted to recreate presenting a quota measure with an order number but failed to find a use for this.

PR #111, opened 2021-06-04
HOTT-665: Modifications to 'breadcrumbs' for commodity code

krios82:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-665

### What?

I have added/removed/altered:

- [ ] Modify the breadcrumbs structure
- [ ] Move each structure (section, header, chapter, "declarable") into a separate partial
- [ ] Extend the specs
- [ ] Fix some linting issues in the affected files

### Why?

I am doing this because users have complained that they find it hard to navigate back to the top of the catalogue hierarchy

### Have you? (optional)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -38,5 +38,7 @@ RSpec/DescribeClass:
     - tasks/**/*.rb
 RSpec/ExampleLength:
   Enabled: false
+RSpec/NestedGroups:
+  Max: 4
 Style/StringLiterals:
   EnforcedStyle: single_quotes
--- a/app/helpers/commodities_helper.rb
+++ b/app/helpers/commodities_helper.rb
@@ -62,7 +62,7 @@ module CommoditiesHelper
 
   def format_commodity_code(commodity)
     code = commodity.display_short_code.to_s
-    "#{code[0..1]}&nbsp;#{code[2..3]}&nbsp;#{code[4..-1]}".html_safe
+    "#{code[0..1]}&nbsp;#{code[2..3]}&nbsp;#{code[4..]}".html_safe
   end
 
   def format_commodity_code_based_on_level(commodity)
@@ -71,7 +71,7 @@ module CommoditiesHelper
 
     if commodity.number_indents > 1 || display_full_code
       # remove trailing pairs of zeros for non declarable
-      code = code.gsub(/[0]{2}+$/, '') if commodity.has_children?
+      code = code.gsub(/0{2}+$/, '') if commodity.has_children?
       tree_code(code, klass: nil)
     end
   end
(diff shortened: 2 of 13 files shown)


app/helpers/commodities_helper.rb
```diff
@@ -97,10 +97,10 @@ def tree_node(main_commodity, commodities, depth)
     if deeper_node.present? && deeper_node.number_indents < main_commodity.number_indents
       tag.ul do
         tag.li do
-          tree_code(deeper_node.code.gsub(/[0]{2}+$/, '')) +
+          tree_code(deeper_node.code.gsub(/0{2}+$/, '')) +
             tag.p(deeper_node.to_s.html_safe) +
             # if deeper_node.producline_suffix == '80'
-            #   tree_code(deeper_node.code.gsub(/[0]{2}+$/, ''))
+            #   tree_code(deeper_node.code.gsub(/0{2}+$/, ''))
             # end +
```
tavyy: Do we still need this code?

.rubocop.yml
```diff
@@ -38,5 +38,7 @@ RSpec/DescribeClass:
     - tasks/**/*.rb
 RSpec/ExampleLength:
   Enabled: false
+RSpec/NestedGroups:
+  Max: 4
```
tavyy, line 42: Will there not be lots of violations in the existing codebase?

tavyy (approved):
Great work really! It looks like it was painful.

spec/requests/chapter_spec.rb
```diff
@@ -1,60 +1,100 @@
...
+  context 'when requesting as JSON' do
+    context 'when requested with json format' do
+      before do
         VCR.use_cassette('chapters#show_01_api_json_format') do
           get '/chapters/01.json'
+        end
+      end
 
-          json = JSON.parse(response.body)
+      let(:json) { JSON.parse(response.body) }
```
willfish, line 72: Wondering whether this memoized helper is something we might want to introduce for all specs in the spec_helper.rb file or equivalent?

spec/requests/commodity_spec.rb
```diff
@@ -66,4 +66,36 @@
       end
     end
   end
+
+  context 'when commodity with whatever' do
+    before do
+      VCR.use_cassette('headings#show_0101') do
```
willfish, line 72: Our VCR cassettes are a total mess. So we've historically labelled the commodity response cassette for commodity 0101300000 `headings_show_0101`. At some point I'd love to relabel all of our cassettes based on the feature that needs them or something.

willfish (approved):
This is awesome

app/helpers/commodities_helper.rb
krios82: Probably not, it's been commented out for ten months now.

.rubocop.yml
krios82: There were a few in the tests I touched, and it was slightly easier to increase the allowed nesting by one.
If you think it's better leave it to 3, I'll go back and try to un-nest the specs instead, but my opinion is that it can be left for a later time

spec/requests/chapter_spec.rb
krios82: It's a good way to deal with multiple checks on the same result, I think we should gradually introduce it when we fix the linting issues in the specs

spec/requests/commodity_spec.rb
krios82: That's another large chunk of work... Not today, maybe? 😄

PR #115, opened 2021-06-10
Hott 736: UK service - Accessing commodities which are also headings (using URL) navigates to XI service

krios82:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-736

### What?

I have added/removed/altered:

- [ ] Add specs to highlight the issue
- [ ] Add exception handling in the `with_source` function that will reset the service choice to the original one
- [ ] Rework the CommoditiesController to fetch the xi commodity only for xi site, and to use the `with_source` only for the "other" source
- [ ] Rework the HeadingsController to fetch the xi heading only for xi site, and to use the `with_source` only for the "other" source

### Why?

I am doing this because on a commodity code not found in the UK app, the app was erroneously redirecting to the NI version of the page

### Have you? (optional)

### DIFF
--- a/app/controllers/commodities_controller.rb
+++ b/app/controllers/commodities_controller.rb
@@ -1,7 +1,8 @@
 class CommoditiesController < GoodsNomenclaturesController
-  before_action :fetch_commodity_from_xi,
-                :fetch_commodity_from_uk,
+  before_action :fetch_commodity,
                 only: %i[show]
+
+  helper_method :uk_commodity, :xi_commodity
 
   def show
     @heading = commodity.heading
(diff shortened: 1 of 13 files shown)


app/controllers/commodities_controller.rb
```diff
@@ -1,8 +1,9 @@
 class CommoditiesController < GoodsNomenclaturesController
-  before_action :fetch_commodity_from_xi,
-                :fetch_commodity_from_uk,
+  before_action :fetch_commodity,
```
willfish, line 2:
```suggestion
  before_action :fetch_commodities,
```

app/controllers/commodities_controller.rb
```diff
@@ -30,19 +31,29 @@ def set_prev_next(commodity)
     ]
   end
 
-  def fetch_commodity_from_xi
-    @xi_commodity = TradeTariffFrontend::ServiceChooser.with_source(:xi) do
-      CommodityPresenter.new(Commodity.find(params[:id], query_params))
+  def fetch_commodity
```
willfish, line 34:
```suggestion
  def fetch_commodities
```

willfish (approved):

app/controllers/commodities_controller.rb
krios82: I used the singular because it's technically only 1 commodity in up to 2 versions.

But, if you think it's more descriptive in the plural, should I also change `fetch_heading` to `fetch_headings` in the headings controller?

willfish: I'm not super picky :)

willfish (approved):

PR #126, opened 2021-07-01
HOTT-785: steel safeguards banner for UK site

krios82:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-785

### What?

I have added/removed/altered:

- [ ] Add steel safeguards banner for UK site and sections pages only

### Why?

I am doing this because:

UK government has made a decision on 30th June for implementation of steel safeguards. This has involved the creation of new steel safeguard quotas.

Notice of this was only given on 30th June and needs to be implemented the next day.

It takes a lot of time to create this data, so there is a need to put a temporary banner up while the data is loaded.

### Have you? (optional)

#### UK sections

Screen Shot 2021-07-01 at 09 34 02

#### XI sections

Screen Shot 2021-07-01 at 09 34 14

#### UK other than section (example)

Screen Shot 2021-07-01 at 09 34 26

### DIFF
--- a/app/helpers/service_helper.rb
+++ b/app/helpers/service_helper.rb
@@ -41,9 +41,16 @@ module ServiceHelper
           tag.p do
             banner_copy
           end
-        end
+        end +
+        steel_safeguards
       end
     end
+  end
+
+  def steel_safeguards
+    return t("service_banner.steel_safeguards").html_safe if uk_service_choice? && request.filtered_path == sections_path
+
+    ''
   end
 
   def search_label_text
--- a/app/helpers/service_helper.rb
+++ b/app/helpers/service_helper.rb
@@ -48,7 +48,7 @@ module ServiceHelper
   end
 
   def steel_safeguards
-    return t("service_banner.steel_safeguards").html_safe if uk_service_choice? && request.filtered_path == sections_path
+    return t("service_banner.steel_safeguards_html").html_safe if uk_service_choice? && request.filtered_path == sections_path
 
     ''
   end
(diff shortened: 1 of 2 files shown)


config/locales/en.yml
```diff
@@ -69,3 +69,24 @@ en:
     big:
       uk: From 1 January 2021, if you’re bringing goods into Northern Ireland from outside the UK and the EU, you will pay the UK duty rate <a href="https://www.gov.uk/guidance/check-if-you-can-declare-goods-you-bring-into-northern-ireland-not-at-risk-of-moving-to-the-eu-from-1-january-2021" class="govuk-link">if your goods are not ‘at risk’ of onward movement to the EU</a>. If they are at risk of onward movement to the EU, use the %{link}.
       xi: From 1 January 2021, if you’re bringing goods into Northern Ireland from outside the UK and the EU, you will pay the UK duty rate <a href="https://www.gov.uk/guidance/check-if-you-can-declare-goods-you-bring-into-northern-ireland-not-at-risk-of-moving-to-the-eu-from-1-january-2021" class="govuk-link">if your goods are not ‘at risk’ of onward movement to the EU</a>. If they are not at risk of onward movement to the EU, use the %{link}.
+    steel_safeguards: >- # Temporary warning regarding the steel safeguards
```
tavyy: you can say here `steel_safeguards_html` - Usually not a good idea to localise an html snipper with styles but given the urgency I am ok with it for now.

app/helpers/service_helper.rb
```diff
@@ -41,11 +41,18 @@ def service_switch_banner(optional_classes: 'govuk-!-margin-bottom-7')
           tag.p do
             banner_copy
           end
-        end
+        end +
+        steel_safeguards
```
willfish, line 45: Does this show for the XI service?

krios82: It doesn't show for xi, or for any other page than sections, as requested by Matt

willfish (approved):

config/locales/en.yml
krios82: I've changed it to `steel_safeguards_html`.

I've put it here because
* the same pattern is already used in the Duty Calculator footnotes, and
* it's supposed to go away at some point soon, so it's good to have it in a very specific and clearly flagged location

Would you have rather put it into the helper instead?

tavyy: Leave it here for now.

tavyy (approved):

PR #135, opened 2021-07-12
Fix asset loading of govuk frontend

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Removes images that come from govuk-frontend package
- [x] Use the assets from the govuk-frontend package
- [x] Fix webpacker config

### Why?

I am doing this because:

- We were copying images that were already loaded into the project by govuk-frontend. If these change, we want images to be updated via the assets in this package rather than having to hand crank changes all of the time.


tavyy (approved):
Nicer!

PR #136, opened 2021-07-13
Remove the Notes accordion component

tavyy:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [ ] Remove the Notes accordion component

### Why?

I am doing this because:

- It does not seem to be ever populated with data.

Screen Shot 2021-07-13 at 14 11 58


krios82 (approved):
LGTM

The `app/views/chapters/_chapter_guides.html.erb` partial seems to be almost empty and completely useless, would it be worth removing it from everywhere else is used and remove the partial?

willfish (approved):

tavyy:
Yeah, I didn't go down that route because there's this `_callout` partial as well that uses this partial and it looks like it's used on the headings.

PR #141, opened 2021-07-14
HOTT-805: On first accessing quota search, please hide the error message

krios82:
On first accessing quota search, please hide the error message "Sorry, there is a problem with the search query. Please specify one or more search criteria"

### Jira link

https://transformuk.atlassian.net/browse/HOTT-805

### What?

I have added/removed/altered:

- [ ] Add a guard in the presenter, so that the following line won't try `reject!` on nil, and the search result will be an array in any case

### Why?

I am doing this because when a search is not provided, the app tries to run a `reject!` on a nil object, and errors.

### Have you? (optional)

### DIFF
--- a/app/presenters/quota_search_presenter.rb
+++ b/app/presenters/quota_search_presenter.rb
@@ -4,7 +4,7 @@ class QuotaSearchPresenter
   def initialize(search_form)
     @with_errors = false
     @search_form = search_form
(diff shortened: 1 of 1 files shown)


krios82:
Screen Shot 2021-07-14 at 10 27 41

willfish (approved):
This class is so annoying that everything is in the constructor. I don't think you've made it any worse but I do think we should refactor it at some stage

PR #152, opened 2021-07-21
Removes unused exchange rates public endpoint

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-855

### What?

I have added/removed/altered:

- [x] Removes the monetary exchange rate api from the list of whitelisted endpoints

### Why?

I am doing this because:

- This has been removed

### DIFF
--- a/lib/trade_tariff_frontend.rb
+++ b/lib/trade_tariff_frontend.rb
@@ -29,7 +29,6 @@ module TradeTariffFrontend
       chapters
       headings
       commodities
-      exchange_rates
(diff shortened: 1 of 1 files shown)


jebw (commented):
Ties in with the backend PR - if we're removing the API should we be sending a 410, or doing something about the removal of the endpoint? Or is it not used

willfish:
Only used by us

jebw (approved):
LGTM :+1:

PR #153, opened 2021-07-26
HOTT-859 Fixed rendering issues with cookie banner

jebw:
Moved from using button_to and a lot of custom css to recommended markup from govuk design system.

Also changed to have a common controller action for accept and reject - this allows moving the form elements outside of the govuk-button-group container to solve rendering problems.

### Jira link

https://transformuk.atlassian.net/browse/HOTT-859

### What?

I have added/removed/altered:

- [x] Replaced the custom CSS/markup for the cookie banner with that recommended by the GovUK Design system
- [x] Replaced the separate accept/reject actions in cookie_consent_controller with a single action - this was because button_to wraps the buttons in form elements, which mucks up the flex box hierarchy causing layout issues with the design system button-group pattern

### Why?

I am doing this because:

- The existing code looked to be an amalgamation of an earlier design and a custom button group implementation - it exhibited a layout bug (buttons showing stacked up and centered). I dropped as much as possible of the custom code to solve the layout issue and ease maintenance.

### Have you? (optional)

- if a user loads a page before deploy, we deploy, then user clicks 'accept' or 'reject' - they will see a 404, reloading the page will resolve it. This probably be solved by adding a custom route to do nothing but reload the page for the 'old' accept/reject routes.

### Before

Screenshot from 2021-07-26 15-21-00 Screenshot from 2021-07-26 15-21-35

### After

Screenshot from 2021-07-26 15-22-27 Screenshot from 2021-07-26 15-21-58


spec/controllers/cookies_consent_controller_spec.rb
```diff
@@ -1,35 +1,39 @@
...
-      '{"settings":true,"usage":"true","remember_settings":"true"}'
-    end
+    describe 'with accept' do
+      let(:acceptance) { 'accept' }
 
-    it 'sets the cookie policy correctly' do
-      expect(response.cookies['cookies_policy']).to eq(expected_cookies)
-    end
+      let(:expected_cookies) do
+        '{"settings":true,"usage":"true","remember_settings":"true"}'
```
tavyy: Do you mind adding some spaces between the key/value pairs?

spec/controllers/cookies_consent_controller_spec.rb
```diff
@@ -1,35 +1,39 @@
...
-  describe 'POST reject_cookies' do
-    subject(:response) { post :reject_cookies }
+    describe 'with reject' do
+      let(:acceptance) { 'reject' }
 
-    let(:expected_cookies) do
-      '{"settings":true,"usage":"false","remember_settings":"false"}'
-    end
+      let(:expected_cookies) do
+        '{"settings":true,"usage":"false","remember_settings":"false"}'
```
tavyy: Do you mind adding some spaces between the key/value pairs?

spec/controllers/cookies_consent_controller_spec.rb
jebw: Yeh good shout - I was trying to minimize the LOC i was changing to keep the diff simple

spec/controllers/cookies_consent_controller_spec.rb
jebw: :+1:

app/controllers/cookies_consent_controller.rb
```diff
@@ -1,13 +1,9 @@
 require 'json'
 
 class CookiesConsentController < ApplicationController
-  before_action :set_cookie_policy, only: %i[accept_cookies reject_cookies]
+  before_action :set_cookie_policy, only: %i[create]
```
tavyy, line 4: 👍

tavyy (approved):
Very nice!

PR #155, opened 2021-07-27
Removes unused next/prev links on declarable view

willfish:
### Jira link

HOTT-<TODO>

### What?

The previous and next links on a commodity were added to go forwards and backwards between siblings in the commodity tree. This isn't helpful for users and wasn't actually asked for by HMRC so after a discussion with the team we've agreed to remove these links.

- [x] Removes unused prev/next on the declarable show page

### Why?

I am doing this because:

- This is unused


willfish:
After

Screenshot from 2021-07-27 10-16-34

Before

Screenshot from 2021-07-27 10-16-03

jebw (approved):

PR #156, opened 2021-07-27
HOTT-788 - Move country selector out of tabs

jebw:
### Jira link

[HOTT-788](https://transformuk.atlassian.net/browse/HOTT-788)

### What?

I have added/removed/altered:

- [x] Moved the country selector on the commodities page to outside of the tabs because we will soon have a third tab with a country selector on
- [x] Replaced the translation texts for the fields with a single version for XI and single version for UK
- [x] Dropped knowledge of the users tab from the country_selector_text
- [x] Added javascript to assign the hidden 'anchor' field prior to submitting the 'select country' form - this wont work for the non-js case, but tabs don't work for the non-js case either
- [x] Changed the search controller to sanitise the tab anchor it will pass through instead of white listing it - this was because it was redirecting from overview and footnotes tabs back to the export tab - the controller doesn't need to know which anchor its redirecting the user to, just needs to ensure its not an open redirect which sanitisation will achieve
- [x] [Drive by fix] - the submit button on the country selector in the non-js scenario was not correctly formatted so added the govuk-button class to it

### Why?

I am doing this because:

- There is a rules of origin tab coming which will also require a country selector.

### Have you? (optional)

- Non I'm aware of

### Before

Screenshot from 2021-07-27 16-46-16

### After

Screenshot from 2021-07-27 16-45-17

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -92,14 +92,7 @@ class SearchController < ApplicationController
   private
 
   def anchor
-    if params.dig(:search, :anchor).present?
-      if params[:search][:anchor] == 'import'
-        '#import'
-      else
-        '#export'
-      end
-    else
-      ''
-    end
+    safe_anchor = params.dig(:search, :anchor).to_s.gsub(/[^a-z]/, '')
+    safe_anchor.present? ? "##{safe_anchor}" : ''
   end
 end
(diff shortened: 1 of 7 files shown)


willfish:
You alright to add some screenshots @jebw?

jebw:
@willfish :point_up: I've updated the description with some :+1:

willfish (approved):

PR #158, opened 2021-07-28
Hott 754 dockerize frontend

mafonso:
### Jira link

HOTT-754

### What?

I have added/removed/altered:

- [ ] Dockerize and deploy app to dev environment.

### Why?

I am doing this because:

- This is a first step to move away from buildpacks.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,6 @@
 # Build compilation image
 FROM ruby:2.7.4-alpine3.13 as builder
 
-
-ENV API_SERVICE_BACKEND_URL_OPTIONS: {"uk":"http://localhost:8080","xi":"http://localhost:8080" }
-
 # The application runs from /app
 WORKDIR /app
 
@@ -32,8 +29,7 @@ COPY . /app/
 
 # Compile assets and run webpack
 # Run in rails test environment to avoid loading development gems
-RUN RAILS_ENV=test bundle exec rails assets:precompile
-ENV RAILS_SERVE_STATIC_FILES=true
+RUN RAILS_ENV=production bundle exec rails assets:precompile
 
 # Cleanup to save space in the production image
 RUN rm -rf node_modules log tmp && \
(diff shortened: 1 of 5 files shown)


config/webpacker.yml
```diff
@@ -83,7 +83,7 @@ test:
   compile: true
 
   # Compile test packs to a separate directory
-  public_output_path: packs-test
+  #public_output_path: packs-test
```
jebw: I think this will cause problems for the developers because any compilation in the test environment will end up in the 'normal' webpacker output path.

From yesterdays comments in Slack I don't think this should be needed anymore?

Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+
+RUN apk add --update --no-cache tzdata && \
+  cp /usr/share/zoneinfo/Europe/London /etc/localtime && \
+  echo "Europe/London" > /etc/timezone
+
+# The application runs from /app
+WORKDIR /app
+
+ENV RAILS_SERVE_STATIC_FILES=true
+ENV RAILS_ENV=production
```
jebw, line 57: I'm not sure how logging is setup on this app but you may also want `RAILS_LOG_TO_STDOUT=true`

jebw (changes requested):
Couple of really minor comments
- the env var may not be necessary depending on whether we are still using the default rails config for logging
- the webpacker change though is potentially problematic but I don't think is necessary anymore?

config/webpacker.yml
mafonso: Yes, we can now revert this since we are building the assets in `production` env.

Dockerfile
mafonso: It does log to stdout already with the current config, that was the case with buildpacks already.  
It could also be that particular config is injected in runtime by the environment, but it certainly goes to stdout, and in the production space it's forwarded to NewRelic

config/webpacker.yml
mafonso: I've reverted this change. This should keep the test environment unaffected, and assets are still built fine with the changes we made yesterday.

jebw (approved):
Just checked and realized the logging is always stdout in prod :+1:

PR #181, opened 2021-08-12
Don't try to deploy dependabot PRs

jebw:
### Jira link

????

### What?

I have added/removed/altered:

- [x] Added an exclusion rule to the deploy to DEV rule for dependabot PRs

### Why?

I am doing this because:

- This will cause a flood of conflicting deploys, and because there's likely to be multiple deploys to DEV they'll overwrite each other anyway.

- An incorrect circle change can impact our ability to deploy

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -250,6 +250,7 @@ workflows:
             branches:
               ignore:
                 - master
+                - /^dependabot\//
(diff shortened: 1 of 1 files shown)


mafonso (approved):
LGTM

willfish (approved):

PR #182, opened 2021-08-12
Removes regulations_enabled? and replaces modal with link to regulation url

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-867

### What?

I have added/removed/altered:

- [x] Remove hiding of regulations by removing the regulations_enabled? flag and references
- [x] Remove modal for rendering legal acts
- [x] Replace modal references with a simple target: '_blank' link to the regulation url if it is present
- [x] Remove unused specs that control the regulation flag

### Why?

I am doing this because:

- UK regulations are now constructed properly and can be show on the frontend safely


app/views/layouts/_base.html.erb
```diff
@@ -76,9 +76,7 @@
       <ul class="govuk-footer__list">
         <li class="govuk-footer__list-item"><a class="govuk-footer__link" target="_blank" href="https://www.gov.uk/government/collections/uk-trade-tariff-volume-1">UK Trade Tariff: Volume 1 &ndash; background information for importers and exporters</a></li>
         <li class="govuk-footer__list-item"><a class="govuk-footer__link" target="_blank" href="https://www.gov.uk/government/collections/uk-trade-tariff-volume-3">UK Trade Tariff: Volume 3 &ndash; procedures, codes and declaration entry details</a></li>
-        <%- if TradeTariffFrontend.regulations_enabled? -%>
-          <li class="govuk-footer__list-item"><a class="govuk-footer__link" target="_blank" href="http://ec.europa.eu/taxation_customs/dds2/taric/taric_consultation.jsp?Lang=en&Screen=0&Expand=true">Integrated tariff of the European Community (TARIC) database</a></li>
-        <% end %>
+        <li class="govuk-footer__list-item"><a class="govuk-footer__link" target="_blank" href="http://ec.europa.eu/taxation_customs/dds2/taric/taric_consultation.jsp?Lang=en&Screen=0&Expand=true">Integrated tariff of the European Community (TARIC) database</a></li>
```
willfish, line 79: Should we show this only on the xi service?

jebw: Think thats a question for matt though it seems odd on the UK service?

jebw (approved):
LGTM

jebw (changes requested):
Just realised the spec needs fixing.

For the regulation URL, I think that'll need a rubocop exception adding - the URL is autogenerated for the EU ones but the UK ones are actually data supplied from CDS (the weird packed information_text field)

Since the data source is CDS I guess we just trust it?

app/views/measures/_measure.html.erb
```diff
@@ -50,30 +50,18 @@
 
   <td><%= measure.excluded_country_list %></td>
 
-  <% if TradeTariffFrontend.regulations_enabled? && TradeTariffFrontend.uk_regulations_enabled? %>
-    <td class="numerical">
-      <% if measure.legal_acts.present? && measure.legal_acts.first.has_national_data? %>
```
willfish, line 55: We don't need any of these conditions since uk regulations and eu regulations all have the same api and so the url comes from the same field. This is since @jebw's improvements in using a standard presenter for all legal acts.

jebw (approved):
LGTM

PR #184, opened 2021-08-13
Adds missing specs for the measures helper

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Fixes additional attribute on duty expression factory
- [x] Removes reference to CHIEF excise measure types (we only pull excise out of
  UK/CDS data source)

### Why?

I am doing this because:

- I was adding other specs for the regulation urls and ended up shaving yaks


app/helpers/measures_helper.rb
```diff
@@ -2,7 +2,6 @@ module MeasuresHelper
   def filter_duty_expression(measure)
     duty_expression = measure.duty_expression.to_s.html_safe
     duty_expression = '' if duty_expression == 'NIHIL'
-    duty_expression = 'see conditions' if measure.measure_type.id.in? %w[DDA DDJ]
```
willfish, line 5: The measure type ids are excise measure type ids from CHIEF. CDS excise measures are always of the 306 type id

willfish: CHIEF is gone

jebw (approved):

PR #185, opened 2021-08-13
Removes service switching configurability

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Removes ability and tests around enabling/disabling service switching

### Why?

I am doing this because:

- We have long since not needed this configurability - since after January and is clogging up the code base


jebw (approved):
Nice to remove old code!

PR #188, opened 2021-08-16
HOTT-846 Rules of Origin tab

jebw:
### Jira link

### What?

I have added/removed/altered:

- [x] Added a Rules of Origin tab, currently hidden behind a feature flag
- [x] Amended the country_flag helper to ignore unrecognised country codes
- [x] Changed any instance of `Search` to only look up all GeographicalAreas once per request

### Why?

I am doing this because:

- We wish to show a Rules of Origin tab with relevant rules to the commodity being viewed 
- We have a mechanism to store the data on the backend but the API exposing that data will need to reflect the requirements of the frontend so this work is happening first. 
  - Next I'll provide an API client to return mock data
  - then I'll extend the backend to return mock data and change the API client to query the backend. 
  - Then I'll extend the backend to return real data.


app/views/measures/_measures.html.erb
```diff
@@ -135,6 +140,21 @@
     <% end %>
   </div>
 
+  <!-- Rules of Origin tab -->
+  <% if TradeTariffFrontend.rules_of_origin_enabled? %>
+  <div class="govuk-tabs__panel" id="rules-of-origin">
```
willfish: Should this be indented once more to indicate the if better?

app/helpers/country_flag_helper.rb
```diff
@@ -1,10 +1,15 @@
 module CountryFlagHelper
   def country_flag_emoji(two_letter_country_code)
+    return unless two_letter_country_code.match? %r{\A[A-Z][A-Z]\z}
```
willfish, line 3: Wondering whether this is unnecessarily defensive?

app/models/search.rb
```diff
@@ -41,11 +41,11 @@ def q=(term)
   end
 
   def countries
-    GeographicalArea.all.compact
+    @countries ||= GeographicalArea.all.compact
```
willfish, line 44: I don't remember countries ever coming back with nil entries :thinking: . I'm not convinced it's actually possible from a jsonapi relationship/include serializer perspective. Probably the compact that you didn't add is superfluous.

app/views/measures/_measures.html.erb
```diff
@@ -6,10 +6,15 @@
     <li class="govuk-tabs__list-item">
       <a class="govuk-tabs__tab" href="#import">Import</a>
     </li>
-    <li class="govuk-tabs__list-item"s>
+    <li class="govuk-tabs__list-item">
       <a class="govuk-tabs__tab" href="#export">Export</a>
     </li>
-    <li class="govuk-tabs__list-item"s>
+    <% if TradeTariffFrontend.rules_of_origin_enabled? %>
```
willfish, line 12: Nice

lib/trade_tariff_frontend.rb
```diff
@@ -91,6 +91,10 @@ def js_sentry_dsn
     ENV.fetch('JS_SENTRY_DSN', '')
   end
 
+  def rules_of_origin_enabled?
+    %w[1 true yes].include? ENV['RULES_OF_ORIGIN_ENABLED'].to_s.downcase
```
willfish: I reckon we can just pick one of these since we're in control of setting it and there are only two of us? I'd go for `true` and `false` and be consistent with things like RAILS_SERVE_STATIC_FILES, etc

willfish: Totally non-blocking

spec/helpers/country_flag_helper_spec.rb
```diff
@@ -1,23 +1,35 @@
...
     subject { helper.country_flag_tag(two_letter_country_code) }
 
-    context "with GB" do
+    context 'with GB' do
       let(:two_letter_country_code) { 'GB' }
 
       it { is_expected.to have_css('span.country-flag', text: '🇬🇧') }
     end
+
+    context 'with non country code' do
```
willfish, line 29: :+1:

willfish (approved):
Looks awesome. Would be good to see some screenshots when you've got the data plumbed together

app/helpers/country_flag_helper.rb
jebw: It was added in because I'd inadvertently called it with a geographical area id that wasn't a country code - it can't handle other format ids - (and they wouldn't correspond to flags anyway). 

If the format is not the expected one - we end up rendering 'unknown character' boxes into the page - just rendering nothing seems like a better way to go.

lib/trade_tariff_frontend.rb
jebw: Cool, will change

app/helpers/country_flag_helper.rb
willfish: Sounds good

willfish: Are we 100% sure that the geographical area ids all match the two letter country codes or do we need a mapping between the two? I'd be surprised if they didn't

jebw: After further discussion with Matt I'm going to switch this to using flags images anyway so I'll address this in that PR

PR #202, opened 2021-08-19
HOTT-846 responsive Rules of Origin tab

jebw:
### Jira link

### What?

I have added/removed/altered:

- [x] Added a `.govuk-table--responsive` class - this will switch to using 'vertical' tables on mobile, deriving column headings from `data-label` attributes on the cells
- [x] Added a `.responsive-full-width` cell modifier, to force certain cells to be 'stacked' - ie cell label above the cell content instead of side by side

### Why?

I am doing this because:

- The rules of origin tab uses govuk tables but these are not responsive - the 2 options were to switch to the existing responsive table pattern which would require a lot of custom theming to handle the embedded markdown content properly, or make the `.govuk-table` class work on mobile. I've made this a modifier css class so that it is opt in.
- There is an optional `.responsive-full-width` cell modifier class because the rule content is potentially long and out of our control so doubling the horizontal width will halve the vertical space taken up by it, at a cost of a predicable 1 additional row for the cell label.
-

### Have you? (optional)

- None - the css changes are opt-in only

### Before

Screenshot from 2021-08-19 10-57-35

### After

Screenshot from 2021-08-19 12-29-20

### DIFF
--- a/app/webpacker/src/stylesheets/_tables.scss
+++ b/app/webpacker/src/stylesheets/_tables.scss
@@ -223,3 +223,69 @@ table.section-browser {
     //}
   }
 }
+
+.govuk-table--responsive {
+  @include govuk-media-query($until: tablet) {
+    & > caption, & > thead, & > tbody {
+      display: block;
+    }
+
+    & > caption, & > thead, & > tbody {
+      display: block;
+
+      & > tr {
+        display: block;
+        border-width: 0 0 1px 0;
+        border-style: solid;
+        border-color: $govuk-border-colour;
+
+        &:first-of-type {
+          border-top-width: 2px;
+        }
+
+        & > th {
+          display: block;
+        }
+
+        & > td {
+          display: block;
+          position: relative;
+          padding-left: 45%;
+          padding-right: 0;
+          width: auto;
+
+          &:before {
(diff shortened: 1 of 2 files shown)


app/webpacker/src/stylesheets/_tables.scss
```diff
@@ -221,3 +221,60 @@ table.section-browser {
...
   }
 }
+
+.govuk-table--responsive {
+  @include govuk-media-query($until: tablet) {
+    caption, thead, tbody, th, td, tr {
+      display: block;
+    }
+    thead tr {
+      /* Hide table headers but keep them screenreader-accessible */
```
willfish: What's a screenreader?

willfish:
Looks great. Do we want to migrate to using this mechanism now in the sections/measures tables just for completeness? Otherwise we're introducing different approaches between the tables when we could be simplifying/consolidating/reducing the WTF's per minute

willfish (approved):

app/webpacker/src/stylesheets/_tables.scss
jebw: An accessibility tool for users - it reads the screen contents out. You can override what they read out using aria tags but its easier if the original markup 'just works'

PR #206, opened 2021-08-23
Manually pin the version of nokogiri since it errors and causes deploy to fail

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Pin to an older version of nokogiri

### Why?

I am doing this because:

- This new version outputs an error which is blocking deploys to dev (and other spaces)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -24,11 +24,10 @@ gem 'webpacker'
 
 # gov UK
 gem 'govspeak'
+gem 'nokogiri', '~> 1.11.7' # https://github.com/sparklemotion/nokogiri/issues/2205
 gem 'plek'
 
 gem 'connection_pool'
-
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -24,12 +24,11 @@ gem 'webpacker'
 
 # gov UK
 gem 'govspeak'
+gem 'nokogiri', '~> 1.11.7' # https://github.com/sparklemotion/nokogiri/issues/2205
```
willfish, line 27:
```suggestion
gem 'nokogiri', '~> 1.11.7' # This update was propagating a non zero exit code as part of a deploy causing deploy failures. https://github.com/sparklemotion/nokogiri/issues/2205.
```

PR #207, opened 2021-08-23
HOTT-909: Cookies banner reappears

jebw:
### Jira link

[HOTT-909](https://transformuk.atlassian.net/browse/HOTT-909)

### What?

I have added/removed/altered:

- [x] Added a `Cookies::Policy` model to control access to the policy cookie - this provides replacements for the various methods in `CookiesHelper` for reading and writing to the cookie
- [x] Added a `Cookies::PoliciesController` to manipulate that cookie, this takes over from `PagesController#tariff_cookies`, `PagesController#update_cookies`, and `CookiesConsentController`
- [x] Added a `Cookies::HideConfirmationsController` to manipulate the 'hide confirmation' cookie - this takes over from `CookiesConsentController#add_seen_confirmation`
- [x] Added a `Cookies` feature spec to verify the different paths a user can follow

### Why?

I am doing this because:

- We are getting inconsistent behaviour from the Cookies banner with it reappearing some of the time. I've not been able to identify the source of the bug but the existing code is hard to follow and there are multiple points reading and writing from the cookie. Hopefully this rewrite will solve the problem but if it doesn't, it should make it _much_ simpler to debug and resolve.

### Follow on PRs
[ ] There will be a follow on PR to drop the existing cookies controller+helper code but I'm rolling it out in two deploys to avoid causing 404s for users - see #208 
[ ] The tables listing cookies do not show correctly on mobile at present. the is an existing issue - I'll add the new responsive functionality from the Rules of Origin PRs once those are merged - see #216

### DIFF
--- a/app/controllers/cookies/policies_controller.rb
+++ b/app/controllers/cookies/policies_controller.rb
@@ -2,19 +2,16 @@ module Cookies
   class PoliciesController < ApplicationController
     before_action { @no_shared_search = true }
 
-    def show
-      @policy = cookies_policy
-    end
+    def show; end
 
     def create
-      @policy = cookies_policy
-      @policy.attributes = policy_params
+      cookies_policy.attributes = policy_params
 
       cookies[:cookies_policy] = {
-        value: @policy.to_cookie,
+        value: cookies_policy.to_cookie,
         expires: 1.year.from_now,
       }
-      @policy.mark_persisted!
+      cookies_policy.mark_persisted!
 
       if policy_params.key? :acceptance
         redirect_back(fallback_location: sections_path)
--- a/app/controllers/cookies/policies_controller.rb
+++ b/app/controllers/cookies/policies_controller.rb
@@ -3,17 +3,18 @@ module Cookies
     before_action { @no_shared_search = true }
 
     def show
-      @policy = Cookies::Policy.from_cookie(cookies[:cookies_policy])
+      @policy = cookies_policy
     end
 
     def create
-      @policy = Cookies::Policy.from_cookie(cookies[:cookies_policy])
+      @policy = cookies_policy
       @policy.attributes = policy_params
 
       cookies[:cookies_policy] = {
         value: @policy.to_cookie,
         expires: 1.year.from_now,
       }
+      @policy.mark_persisted!
 
       if policy_params.key? :acceptance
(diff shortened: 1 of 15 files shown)


app/controllers/cookies/policies_controller.rb
```diff
@@ -0,0 +1,36 @@
+module Cookies
+  class PoliciesController < ApplicationController
+    before_action { @no_shared_search = true }
+
+    def show
+      @policy = cookies_policy
```
willfish:
```suggestion
      @cookies_policy = cookies_policy
```

app/controllers/cookies/policies_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+
+    def create
+      @policy = cookies_policy
```
willfish:
```suggestion
      @cookies_policy = cookies_policy
```

app/controllers/cookies/policies_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+      @policy.attributes = policy_params
```
willfish:
```suggestion
      @cookies_policy.attributes = policy_params
```

app/controllers/cookies/policies_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+
+      cookies[:cookies_policy] = {
+        value: @policy.to_cookie,
```
willfish:
```suggestion
        value: @cookies_policy.to_cookie,
```

app/controllers/cookies/policies_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+        expires: 1.year.from_now,
+      }
+      @policy.mark_persisted!
```
willfish:
```suggestion
      @cookies_policy.mark_persisted!
```

app/controllers/cookies/policies_controller.rb
jebw: This will actually clash with the @cookies_policy set in `ApplicationController#cookies_policy` helper method - since this is duplicating that assignment anyway I've replaced references to `@policy` with usage of the helper_method `cookies_policy`

PR #218, opened 2021-08-26
Remove unused feature flags

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Remove unused Locking behaviour (for an old CDN config)
- [x] Removes unused feature flags (corresponding environment variable scripted cleanup also happening)

### Why?

I am doing this because:

- These all bloat up the application and make it appear more complex than it really is


jebw (approved):
LGTM - looks like the rack-attack gem becomes unused, can that be dropped now?

PR #219, opened 2021-08-27
Fixed cookie associations on form

jebw:
### Jira link

[HOTT-909](https://transformuk.atlassian.net/browse/HOTT-909)

### What?

I have added/removed/altered:

- [x] Corrected the labels on the old cookies page

### Why?

I am doing this because:

- it will allow me to run the same revised cypress cookie tests against both new and old cookie pages to compare/contrast

- This is not intended to be merged but is harmless if it is - the page will be getting dropped in #208


willfish (approved):

jebw:
This was just for comparison purposes

PR #225, opened 2021-09-01
Tidies

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Lint measure class


app/models/measure.rb
```diff
@@ -97,16 +103,6 @@ def additional_code
     @additional_code.presence || NullObject.new(code: '')
   end
 
-  def key
```
willfish, line 100: I couldn't see this used anywhere

PR #226, opened 2021-09-03
Surface declarable for headings and commodities

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-811

### What?

I have added/removed/altered:

- Surface declarable on search results (see https://github.com/trade-tariff/trade-tariff-backend/pull/195)
- Expose search endpoint to be used by api (see https://github.com/trade-tariff/trade-tariff-backend/pull/195)

### Why?

I am doing this because:

- We need to surface whether a search result is declarable (only ever true
for a commodity or heading) for the Single Trade Window team.
- We also need to make sure that the STW team use the actual api endpoint
to do search so we're exposing this ability in the public api endpoint
whitelist configuration.

### DIFF
--- a/lib/trade_tariff_frontend.rb
+++ b/lib/trade_tariff_frontend.rb
@@ -33,6 +33,7 @@ module TradeTariffFrontend
       quotas
       goods_nomenclatures
       search_references
+      search
       additional_codes
       certificates
       footnotes
--- a/spec/support/shared_examples/a_declarable.rb
+++ b/spec/support/shared_examples/a_declarable.rb
@@ -1,0 +1,3 @@
+RSpec.shared_examples 'a declarable' do
+  it { is_expected.to respond_to(:declarable) }
+end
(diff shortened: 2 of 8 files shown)


spec/support/shared_examples/a_declarable.rb
```diff
@@ -0,0 +1,3 @@
+RSpec.shared_examples 'a declarable' do
```
willfish, line 1: The idea is that we can easily backfill the behaviour for declarable concerns in future and have both heading and commodity implementations have conforming implementations

lib/trade_tariff_frontend.rb
```diff
@@ -33,6 +33,7 @@ def public_api_endpoints
       quotas
       goods_nomenclatures
       search_references
+      search
```
jebw, line 36: This is the only question mark I have - I'd thought the take away was that we'd do the minimal extension of the `/search.json` endpoint but not expose the 'old' null pattern search results on a new api endpoint

At some point in the future we'd expose the API we wanted onto `/v2/api/search` ? Ie with consistent search result formatting?

jebw (commented):
Only question is on exposing the current format of the results on a new endpoint?

lib/trade_tariff_frontend.rb
willfish: So I think it's fine to extend but not make obviously available for now via docs?

jebw (approved):

PR #231, opened 2021-09-13
HOTT-846: Lint fixes to spec controllers

jebw:
Lint fixes to controllers I'll be changing

### Jira link

### What?

I have added/removed/altered:

- [x] Linted spec files I will be changing 

### Why?

I am doing this because:

- to simplify the PR with the changes

### DIFF
--- a/app/models/commodity.rb
+++ b/app/models/commodity.rb
@@ -14,10 +14,12 @@ class Commodity < GoodsNomenclature
   has_many :ancestors, class_name: 'Commodity'
 
   delegate :goods_nomenclature_item_id, :display_short_code, to: :heading, prefix: true
-  alias :short_code :goods_nomenclature_item_id
+  alias_method :short_code, :goods_nomenclature_item_id
 
   def substring=(substring)
+    # rubocop:disable Naming/MemoizedInstanceVariableName
     @substring ||= substring.to_i
+    # rubocop:enable Naming/MemoizedInstanceVariableName
   end
 
   def leaf?
@@ -25,11 +27,11 @@ class Commodity < GoodsNomenclature
   end
 
   def has_children?
-    not(leaf?)
+    !leaf?
   end
 
   def display_short_code
-    code[4..-1]
+    code[4..]
   end
 
   def display_export_code
--- a/spec/controllers/commodities_controller_spec.rb
+++ b/spec/controllers/commodities_controller_spec.rb
@@ -2,7 +2,7 @@ require 'spec_helper'
 
 describe CommoditiesController, type: :controller do
   describe 'GET to #show' do
-    context 'with UK site' do
+    context 'with XI site' do
       before do
         allow(TradeTariffFrontend::ServiceChooser).to receive(:service_choice).and_return('xi')
         allow(TradeTariffFrontend::ServiceChooser).to receive(:with_source).with(:xi).and_call_original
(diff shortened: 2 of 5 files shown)


app/models/commodity.rb
```diff
@@ -14,22 +14,24 @@ class Commodity < GoodsNomenclature
   has_many :ancestors, class_name: 'Commodity'
 
   delegate :goods_nomenclature_item_id, :display_short_code, to: :heading, prefix: true
-  alias :short_code :goods_nomenclature_item_id
+  alias_method :short_code, :goods_nomenclature_item_id
 
   def substring=(substring)
+    # rubocop:disable Naming/MemoizedInstanceVariableName
```
alexdesi, line 20: really? 
I am wondering how we should Memoize in this case?
sometimes Rubocop is annoying

spec/controllers/commodities_controller_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 describe CommoditiesController, type: :controller do
   describe 'GET to #show' do
-    context 'with UK site' do
+    context 'with XI site' do
```
alexdesi, line 5: good catch :)

spec/controllers/commodities_controller_spec.rb
```diff
@@ -65,8 +64,7 @@
         end
 
         it 'redirects to actual version of the commodity page' do
-          expect(response.status).to eq 302
-          expect(response.location).to eq commodity_url(id: commodity_id.first(10))
+          expect(response).to redirect_to commodity_url(id: commodity_id.first(10))
```
alexdesi, line 67: `expect(response.status).to eq 302` seemed a genuine test, should we keep it?

spec/controllers/commodities_controller_spec.rb
```diff
@@ -88,16 +86,15 @@
 
         it 'redirects to heading page (strips exceeding commodity id characters)' do
           TradeTariffFrontend::ServiceChooser.service_choice = nil
-          expect(response.status).to eq 302
-          expect(response.location).to eq heading_url(id: commodity_id.first(4))
+          expect(response).to redirect_to heading_url(id: commodity_id.first(4))
```
alexdesi, line 89: same here

spec/controllers/commodities_controller_spec.rb
```diff
@@ -109,8 +106,7 @@
 
         it 'redirects to actual version of the commodity page' do
           TradeTariffFrontend::ServiceChooser.service_choice = nil
-          expect(response.status).to eq 302
-          expect(response.location).to eq commodity_url(id: commodity_id.first(10))
+          expect(response).to redirect_to commodity_url(id: commodity_id.first(10))
```
alexdesi, line 109: and here

alexdesi (approved):
It looks good!
the only thing I am not sure about is the removing of the 302 HTTP response tests (redirect)

spec/controllers/commodities_controller_spec.rb
jebw: My conclusion was we're probably not fussed about the particular status code, just that its in the 3xx range

That matcher chains through to the rails one, which checks a `3xx` status code, and checks the `response.location` ?

https://github.com/rails/rails/blob/83217025a171593547d1268651b446d3533e2019/actionpack/lib/action_dispatch/testing/assertions/response.rb#L53

alexdesi: I got your point, yes it sounds good to me! 👍

spec/controllers/commodities_controller_spec.rb
alexdesi: we've agreed to keep it (see the comment before)

spec/controllers/commodities_controller_spec.rb
alexdesi: as above

app/models/commodity.rb
jebw: Yeh, this left me stumped - hence giving up and using the `'rubocop:disable`

PR #195, opened 2021-08-18
HOTT-846 rules of origin api client

jebw:
### Jira link

[HOTT-846](https://transformuk.atlassian.net/browse/HOTT-846)

### What?

I have added/removed/altered:

- [x] Fixed the linting on various existing files I've modified
- [x] Added a ProductSpecificRule api class
- [x] Changed the GoodsNomenclature class to add `#rules` method for fetching associated rules
- [x] Overriden `#rules` in Heading model so it shorts out if the Heading is not declarable
- [x] Changed the ProductSpecificRule api class to return mock data till the endpoint on the backend is done

### Why?

I am doing this because:

- We need to query the backend for the rules, I wanted to clarify the required API since it influences the data load tickets

### Notes

This PR is against an earlier PR - I'll merge that after we've done a release to prod of the frontend some I'm not holding Mahdu up

### Deployment risks (optional)

- Mocked data is loaded but the actual Rules of Origin tab is feature flagged off

### Screenshots

Screenshot from 2021-08-18 12-41-20

------------------

Screenshot from 2021-08-18 12-41-44

PR #232, opened 2021-09-13
HOTT-846: Add Rules of Origin API client

jebw:
### Jira link

### What?

I have added/removed/altered:

- [x] Added an API client for the `/rules_of_origin_schemes` endpoint
- [x] Disabled feature flag in `Rails.env.development?` - the logic being other backend branches may not yet have latest api changes - this avoids `master` frontend having a hard dependency on tip of `master` backend. Developers can enable rules of origin by setting feature flag locally.

### Why?

I am doing this because:

- We should be querying the backend rather then showing mocked data from the frontend

### Notes

This supercedes #195


app/models/goods_nomenclature.rb
```diff
@@ -51,4 +51,14 @@ def id
   def code
     goods_nomenclature_item_id
   end
+
+  def rules_of_origin(*args, **kwargs)
+    return nil unless try(:declarable?)
```
willfish: Don't we always have this answer?

spec/models/goods_nomenclature_spec.rb
```diff
@@ -92,4 +92,72 @@
       end
     end
   end
+
+  describe '#rules_of_origin' do
+    before do
+      # Needed because goods nomenclature isn't declarable but #rules_of_origin
```
willfish, line 98: I think we can reflect that section, chapters, etc, all have declarable and just treat them as ducks/trust the data

willfish (approved):
Love it

app/models/goods_nomenclature.rb
jebw: Not on `Chapter` which also inherits from `GoodsNomenclature`

spec/models/goods_nomenclature_spec.rb
jebw: Ok, will change `Chapter` to always return false for `#declarable?`

jebw: This override will still be required though because we want to bypass the short-circuit

willfish (approved):

willfish (approved):

PR #238, opened 2021-09-16
Meursing lookup

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-618

### What?

I have added/removed/altered:

- [x] Adds meursing lookup wizard for taking meursing codes inputs
- [x] Adds meursing lookup results page for using inputs to generate an answer additional code that the user can copy and use to calculate their meursing code
- [x] Adds wizard gem to simplify population and handling of wizard elements
- [x] Adds dfe formbuilder gem to simplify writing GDS compliant forms
- [x] Adds view, controller and model specs for steps

### Why?

I am doing this because:

- This is needed for users to understand how to calculate their duties based off meursing measures and will be plumbed into/referenced by the duty calculator later


app/controllers/meursing_lookup/steps_controller.rb
```diff
@@ -0,0 +1,41 @@
+module MeursingLookup
+  class StepsController < ApplicationController
+    before_action do
+      @no_shared_search = true
+      @tariff_last_updated = nil
+      clear_session
```
jebw: can this be renamed to something less generic, eg `clear_meursing_wizard_session` - initially I'd thought this was calling a method in Rails till I read lower in the PR.

config/routes.rb
```diff
@@ -75,11 +79,11 @@
           )
   end
 
-  constraints(id: /[\d]{1,2}/) do
+  constraints(id: /\d{1,2}/) do
```
jebw: Semantically these look identical to me - has there been some testing around the change? Presumably Madhu's test suite covers all these end points?

app/models/meursing_lookup/wizard.rb
```diff
@@ -0,0 +1,25 @@
...
+      MeursingLookup::Steps::MilkProtein,
+      MeursingLookup::Steps::ReviewAnswers,
+      MeursingLookup::Steps::End,
+    ]
+
+    def self.step_level(key)
+      steps.index { |step| step.key == key }
+    end
+
+    def answers_by_step
```
jebw, line 13: You could add the following to `Steps::Start`, `Steps::End`, and `Steps::ReviewAnswers` which would avoid needing to special case those steps within the wizard?

```
def reviewable_answers; {}; end
```

app/models/meursing_lookup/steps/base.rb
```diff
@@ -0,0 +1,66 @@
...
+  module Steps
+    class Base < WizardSteps::Step
+      def self.inherited(child_class)
+        child_class.attribute child_class.key, :string
+        child_class.validates child_class.key, presence: true
+
+        super
+      end
+
+      def options
```
jebw: This method is doing several things and is quite hard to follow. Would it make sense to separate some of the concerns - so move the tree to its own class and add some methods that class for querying the dataset - that may help to simplify the logic here. 

```
class MeursingCodeTree
  def self.load_default
      new(File.read(....))
  end
end

module TradeTariffBackend
  def meursing_code_tree
    @meursing_code_tree ||= MeursingCodeTree.load_default
  end
end
```

app/models/meursing_lookup/steps/base.rb
```diff
@@ -0,0 +1,66 @@
...
+        @options ||= begin
+          current_tree = tree
+
+          previous_answers.each.with_index do |answer, index|
+            if index >= previous_answers.size - 1
```
jebw: This should be assigned to a variable name explaining what `index >= previous_answers - 1` is testing for - this would help explain what the conditional is for.

app/controllers/meursing_lookup/steps_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    def wizard_store_key
+      :meursing_lookup
+    end
+
+    def set_page_title
+      @title = "#{@current_step.key} step"
+    end
+
+    def back_link(step_path)
```
jebw: I don't think the controller should be generating HTML content - thats more the domain of the View layer. I can't see this being used elsewhere in the controller so maybe it could go in a view helper somewhere? Though I'm not sure that saves much over `link_to 'Back', step_path`?

app/models/meursing_lookup/steps/base.rb
jebw: Part of me wonders if this code is generic but without needing to be - if there's a generic 'query' option on a `MeursingCodeTree` - would that mean each of the steps could have their own `def options` passing in relevant params to the `MeursingCodeTree#lookup` ?

jebw (dismissed):
Looks great - I think some of the logic around Meursing lookup should be in its own class rather than the Wizard step models though - this will make it easier to write specs for as well.

Specs will also need adding for the rest before merging as well. Looks great :+1:

app/controllers/meursing_lookup/steps_controller.rb
willfish: Good idea. Feels like it could clear unrelated thinks

config/routes.rb
willfish: The square braces aren't required. None yet. I'll probably revert this

app/controllers/meursing_lookup/steps_controller.rb
willfish: I had intended on moving this to a view helper and view spec, actually

app/models/meursing_lookup/wizard.rb
willfish: I don't think this works since I'd still need to exclude the steps sadly

app/models/meursing_lookup/steps/starch.rb
```diff
@@ -0,0 +1,7 @@
+module MeursingLookup
+  module Steps
+    class Starch < AnswerStep
+      alias_method :current_tree, :tree
```
willfish: The top of the tree is synonymous with tree

willfish:
```suggestion
      # The top of the tree is synonymous with tree for the first input question
      alias_method :current_tree, :tree
```

app/helpers/steps_helper.rb
```diff
@@ -0,0 +1,9 @@
+module StepsHelper
```
alexdesi: we could be more specific here, 
what about naming it `MeursingStepHelper`,
in case we add more step forms in the project ... for example `CalculateBananasStepForms`

app/models/meursing_lookup/steps/end.rb
```diff
@@ -0,0 +1,18 @@
...
+
+      def meursing_code
+        missing_milk_protein_meursing_code || tree.dig(starch_answer, sucrose_answer, milk_fat_answer, milk_protein_answer)
+      end
+
+      private
+
+      # 'skip_milk_protein' keys mean the next tree level is skipped and we do not have answers for it so just retreive the 'skip_milk_protein' key value which is a valid meursing code
+      def missing_milk_protein_meursing_code
+        tree.dig(starch_answer, sucrose_answer, milk_fat_answer, 'skip_milk_protein').presence
```
alexdesi: maybe present?
are you checking the presence or it is important the returning value?

app/views/meursing_lookup/steps/_end.html.erb
```diff
@@ -0,0 +1,36 @@
...
+      </div>
+    </div>
+
+    <h2 class="govuk-heading-m">What next</h2>
+
+    <p class="govuk-body">
+      Use these four digits together with the ten-digit commodity code from Trade Tariff to work out duties applicable to certain complex agri-foods on the the Northern Ireland Tariff, when your import is considered to be 'at risk'.
+    </p>
+
+    <% if last_commodity_code.present? %>
```
alexdesi: I am thinking with we call this 
`current_commodity_code` 🤔 

last_commodity_code sounds a DB querry

spec/views/meursing_lookup/steps/show.html.erb_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+  context 'when the current step is a review_answer step' do
+    let(:current_step) { MeursingLookup::Steps::ReviewAnswers.new(wizard, store) }
+
+    it { expect(rendered).to render_template('meursing_lookup/steps/show') }
+    it { expect(rendered).to render_template('meursing_lookup/steps/_form') }
+    it { expect(rendered).to render_template('meursing_lookup/steps/_review_answers') }
+    it { expect(rendered).to render_template('meursing_lookup/steps/_review_answer') }
+  end
+
+  context 'when the current step is an end step' do
```
alexdesi: There is just one end, right?
"when the current step is **the** end step" do

spec/models/meursing_lookup/steps/milk_fat_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  it_behaves_like 'an answer step', 'milk_fat' do
+    describe '#current_meursing_code_level' do
+      let(:input_answers) do
+        {
+          'starch' => '0 - 4.99',
+          'sucrose' => '0 - 4.99',
+        }
+      end
+
+      let(:tree) do
```
alexdesi: maybe
`let(:meursing_codes) do`  ?
...
and 

`eq(meursing_codes['0 - 4.99']['0 - 4.99'])`

alexdesi (approved):
Amazing job! 🔥 🔥 🔥 

a couple of silly comments, you can just ignore them.

PR #248, opened 2021-09-21
HOTT-953: Switch to using images for flags

jebw:
On the Rules of Origin tab

We were previously relying on unicode characters but some countries flags were not included in the Unicode tables.

### Jira link

[HOTT-953](https://transformuk.atlassian.net/browse/HOTT-953)

### What?

I have added/removed/altered:

- [x] Added images for the various flags
- [x] Changed the country_flag_tag to use generate an image based flag instead of unicode character based

### Why?

I am doing this because:

- Some countries flags were not represented in the unicode tables

### Have you? (optional)


app/views/measures/_rules_of_origin.html.erb
```diff
@@ -1,8 +1,8 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h2 class="govuk-heading-l">
+    <h2 class="govuk-heading-l rules-of-origin-heading">
       Rules of origin for trading with <%= country_name %>
-      <%= country_flag_tag(country_code) %>
+      <%= country_flag_tag country_code, alt: "Flag for #{country_name}" %>
```
willfish, line 5: Helpful!

willfish (approved):

PR #256, opened 2021-09-24
HOTT-964: Intro text

jebw:
### Jira link

[HOTT-964](https://transformuk.atlassian.net/browse/HOTT-964)

### What?

I have added/removed/altered:

- [x] Added the intro text for countries we do not have a trade agreement with
- [x] Added the intro text for countries we have direct trade agreement
- [x] Added the intro text for countries we have an indirect trade agreement with, via a trade bloc - eg, EU
- [x] Moved all the Rules of Origin view templates to their own folder

### Why?

I am doing this because:

- We should be showing introducing the trading scheme we are operating with
- I moved the view templates because we are ending up with a few of them will probably have more in the future


app/helpers/rules_of_origin_helper.rb
```diff
@@ -0,0 +1,21 @@
+module RulesOfOriginHelper
+  def rules_of_origin_service_name
+    TradeTariffFrontend::ServiceChooser.uk? ? 'UK' : 'EU'
+  end
+
```
alexdesi, line 5: we use this logic in here
`app/views/rules_of_origin/_tab.html.erb#15`  too,
maybe we can reuse this method there

app/helpers/rules_of_origin_helper.rb
```diff
@@ -0,0 +1,21 @@
...
+  def rules_of_origin_schemes_intro(country_name, schemes)
+    if schemes.empty?
+      return render 'rules_of_origin/intro_no_scheme', country_name: country_name
```
alexdesi: I think the `( )` can make it a bit more tidy
return render('rules_of_origin/intro_no_scheme', country_name: country_name`)

up to you anyway, we do not have strictly rule for that 😄

app/helpers/rules_of_origin_helper.rb
```diff
@@ -0,0 +1,21 @@
...
+    end
+
+    schemes_intros = schemes.map do |scheme|
+      partial = scheme.countries.length > 1 ? 'bloc' : 'country'
```
alexdesi: with `bloc` do we mean  `bloc_of_countries` ?
we should consider renaming it

alexdesi (approved):
it sounds good to me ✅ 
a couple of comments, not blocking anyway

app/helpers/rules_of_origin_helper.rb
jebw: :) Just found that - I've changed it in #257

app/helpers/rules_of_origin_helper.rb
jebw: done

app/helpers/rules_of_origin_helper.rb
jebw: changed this to `trade_bloc` - I figured that was clearer

PR #261, opened 2021-09-27
HOTT-956: Expose introductory notes

jebw:
### Jira link

[HOTT-956](https://transformuk.atlassian.net/browse/HOTT-956)

### What?

I have added/removed/altered:

- [x] Added api support for introductory_notes field on schemes
- [x] Include the introductory_notes in the expanding explanation field below the rules on the Rules of Origin tab
- [x] Don't show the expanding notes if there are no rules for the chosen country

### Why?

I am doing this because:

- We want to show the introductory notes information on the Rules of Origin tab


app/views/rules_of_origin/_introductory_notes.html.erb
```diff
@@ -0,0 +1,60 @@
...
+    <p>
+      If the heading in column 1 starts with ‘ex’, the rules in column 3
+      <!--(and 4 if present)//--> will only apply to the product as described
+      in column 2.
+    </p>
+
+    <h2 class="govuk-heading-m">Introductory notes</h2>
+
+    <div class="tariff-markdown">
+      <%= govspeak introductory_notes.to_s %>
```
alexdesi: do we need `to_s` here?

app/views/rules_of_origin/_tab.html.erb
```diff
@@ -15,14 +15,15 @@
       <%= render 'rules_of_origin/rules_table',
                  country_name: country_name,
                  rules: rules %>
+
+      <%= render partial: 'rules_of_origin/introductory_notes',
+                 collection: rules_of_origin.map(&:introductory_notes) %>
```
alexdesi, line 20: maybe adding `introductory_notes` as a public method to get them, instead of map here? 🤔

alexdesi (approved):
it looks good 👍 
just added 2 minor comments

app/views/rules_of_origin/_introductory_notes.html.erb
jebw: safety net incase its nil and govsoeak doesn't handle nil, possible overkill - will change it to `if present?`

app/views/rules_of_origin/_tab.html.erb
jebw: Problem is rules of origin is an `Array` of `RulesOfOrigin::Scheme` not a specific class, so not sure where I would add it?

PR #272, opened 2021-09-28
Lock Rails version to our current release

jebw:
### Jira link

[HOTT-981](https://transformuk.atlassian.net/browse/HOTT-981)

### What?

I have added/removed/altered:

- [x] Restricted the rails version definition in the Gemfile

### Why?

I am doing this because:

- The current definition accepts _any_ version of Rails but upgrading the rails gem should be done in lockstep with running the `rails app:update` command so we want to ensure the developer is doing a dedicated rails upgrade for anything other than patch level updates.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ git_source(:github) do |repo_name|
   "https://github.com/#{repo_name}.git"
 end
 
-gem 'rails', '~> 6'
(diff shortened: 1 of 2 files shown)


willfish (approved):
This makes loads of sense. Good shout

PR #275, opened 2021-09-28
HOTT-878 Override content of conditions dialogs

alexdesi:
### What?

This commit replaces, augments or keeps the content of the conditions (modals)
based on the config file measure_condition_modal_config.yaml

### Jira link

https://transformuk.atlassian.net/browse/HOTT-878

### Why?
Because much of the content that comes back from the raw CDS data is pretty difficult to understand, we would like to be able to replace or augment it with curated, custom content in the conditions window.


app/helpers/measures_helper.rb
```diff
@@ -1,4 +1,40 @@
...
+  def measure_condition_config_for(goods_nomenclature_item_id, additional_code, measure_type_id)
+    config = TradeTariffFrontend.measure_condition_modal_config.find do |modal_config|
+      modal_config['goods_nomenclature_item_id'] == goods_nomenclature_item_id &&
+        modal_config['additional_code'] == additional_code &&
+        modal_config['measure_type_id'] == measure_type_id
+    end
+
+    config || DEFAULT_MODAL_CONFIG
+  end
+
```
alexdesi: During my sleep (really!) I realized that we should - if we have a couple - of hours, move these responsibilities into a class: 
`MeasureConditionAugmenter`, `MeasureConditionReplacer` 
that would simplify the naming of the public methods,
and we could place the config loader into the initializer.

app/models/measure_condition_augmenter.rb
```diff
@@ -0,0 +1,58 @@
+class MeasureConditionAugmenter
```
willfish: This is too specific in naming, now. It has one of 3 modes (default, replace, augment). It's also not a model in the sense that it's a verb, not a noun.

willfish:
```suggestion
class MeasureConditionDialogPartialOption
```

app/models/measure_condition_augmenter.rb
```diff
@@ -0,0 +1,54 @@
...
+        entry['goods_nomenclature_item_id'] == goods_nomenclature_item_id &&
+          entry['additional_code'] == additional_code &&
+          entry['measure_type_id'] == measure_type_id
+      end
+
+      measure_condition_to_augment || DEFAULT_CONFIG_ENTRY
+    end
+
+    def config
+      @config ||= YAML.load_file(Rails.root.join(CONFIG_FILE_NAME))
```
willfish: Much better method naming :)

app/models/measure_condition_augmenter.rb
```diff
@@ -0,0 +1,54 @@
+class MeasureConditionAugmenter
+  class << self
+    DEFAULT_CONFIG_ENTRY = {
+      content_file: nil,
+      overwrite: nil,
+    }.freeze
+
+    CONFIG_FILE_NAME = 'db/measure_condition_modal_config.yaml'.freeze
```
willfish: I'd always qualify file names just to be sure we're not reliant on the current working directory. It makes the application less flimsy/more portable

willfish (approved):
Looks got to me :)

spec/models/measure_condition_dialog_spec.rb
```diff
@@ -0,0 +1,75 @@
+require 'spec_helper'
+
+describe MeasureConditionDialog do
+  subject(:dialog) { described_class.build(declarable, measure) }
+
+  before do
+    stub_const('MeasureConditionDialog::CONFIG_FILE_NAME',
+               File.join('spec', 'fixtures', 'measure_condition_dialog_config.yaml'))
+  end
```
alexdesi, line 8: There was still a problem: our tests were linked to real data,
in fact, when I commented out 2 entries the tests failed.
So I created a fixture file just for testing.

PR #276, opened 2021-09-29
HOTT-984: Interpret placeholders

jebw:
### Jira link

[HOTT-984](https://transformuk.atlassian.net/browse/HOTT-984)

### What?

I have added/removed/altered:

- [x] Added a helper to substitute the tagged placeholders in the rule descriptions
- [x] Use this placeholder in the table for showing Rules of Origin

### Why?

I am doing this because:

- We should be showing these markers in a manner friendly to our users

### Have you? (optional)

### Notes

The implementation is decidly non optimal - the CSS needs work and using `<span>`s is not great - _but_ kramdown wont pass block level html elements through if we process before HTML conversion, and if we do it after we are stuck trying to fixup HTML nesting problems negating the advantage of markdown. 

This can be thought of as a strategic solution since it needs sorting for tomorrow but I'll iterate on it later.

### DIFF
--- a/app/helpers/rules_of_origin_helper.rb
+++ b/app/helpers/rules_of_origin_helper.rb
@@ -1,4 +1,6 @@
 module RulesOfOriginHelper
+  ROO_TAGGED_DESCRIPTIONS = %w[CC CTH CTSH EXW WO].freeze
+
   def rules_of_origin_service_name
     TradeTariffFrontend::ServiceChooser.uk? ? 'UK' : 'EU'
   end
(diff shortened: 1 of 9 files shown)


app/helpers/rules_of_origin_helper.rb
```diff
@@ -18,4 +20,16 @@ def rules_of_origin_schemes_intro(country_name, schemes)
 
     safe_join schemes_intros, "\n\n"
   end
+
+  def rules_of_origin_tagged_descriptions(content)
+    content.gsub(/\{\{([A-Z]+)\}\}/) do |_match|
+      matched_tag = Regexp.last_match(1)
```
willfish, line 26: Is this the same as match.last?

willfish (approved):

app/helpers/rules_of_origin_helper.rb
jebw: :eyes:

jebw: Unfortunately not - `String#gsub` returns a `String` as `_match`, not a regular expression `MatchData`  - the API docs are ambiguous but imply its MatchData but if I check in the debugger `_match` is definitely a String

jebw: Think I'll go with it as is - would have much preferred to not rely on `.last_match` - anything involving globals feels ugly

PR #277, opened 2021-09-30
HOTT-990: Removed the alerting for missing flags

jebw:
### Jira link

[HOTT-990](https://transformuk.atlassian.net/browse/HOTT-990)

### What?

I have added/removed/altered:

- [x] Removed the alerting for missing flag images

### Why?

I am doing this because:

- GeographicalArea's like XI don't actually have flags so this will just alert indefinitely without being fixable.

### DIFF
--- a/app/helpers/country_flag_helper.rb
+++ b/app/helpers/country_flag_helper.rb
@@ -1,10 +1,16 @@
 module CountryFlagHelper
+  COUNTRY_FLAG_IGNORE = %w[xi xc xl].freeze
+
   def country_flag_tag(two_letter_country_code, **kwargs)
-    image_pack_tag "flags/#{two_letter_country_code.downcase}.png",
+    two_letter_country_code = two_letter_country_code.downcase
+
+    image_pack_tag "flags/#{two_letter_country_code}.png",
                    **kwargs.merge(class: 'country-flag')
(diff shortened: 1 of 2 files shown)


app/helpers/country_flag_helper.rb
```diff
@@ -3,9 +3,6 @@ def country_flag_tag(two_letter_country_code, **kwargs)
     image_pack_tag "flags/#{two_letter_country_code.downcase}.png",
                    **kwargs.merge(class: 'country-flag')
   rescue Webpacker::Manifest::MissingEntryError
-    Raven.capture_message \
-      "Missing flag image file for #{two_letter_country_code.downcase}"
-
```
alexdesi, line 8: Not sure how often this happens, maybe should we just log it, as a info?

spec/helpers/country_flag_helper_spec.rb
```diff
@@ -35,5 +35,17 @@
         expect(Raven).to have_received(:capture_message)
       end
     end
+
+    context "with country which we don't have a flag which is in ignore list" do
```
alexdesi, line 39: this should be a bit more readable:
`context 'when country do not have a flag and is in the ignore list'`

alexdesi (approved):
👍

PR #288, opened 2021-10-06
Setup JS Framework Jest

alexdesi:
### What?
This PR setup Jest, in order to write unit tests for our JS codes.
It adds the first test for the commodities.js (former commodities.js.erb)

### Jira link

HOTT-991


app/webpacker/src/javascripts/commodities.js
```diff
@@ -2,13 +2,16 @@
 /*jslint
  white: true */
 
-import IMask from 'imask';
-import debounce from "./debounce";
+// import IMask from 'imask';
```
willfish, line 5: Do we not use this import?

babel.config.js
```diff
@@ -20,9 +20,14 @@ module.exports = function(api) {
       isTestEnv && [
         require('@babel/preset-env').default,
         {
-          targets: {
-            node: 'current'
-          }
+          // targets: {
```
willfish: Probably we should remove rather than comment this?

spec/javascript/setup.js
```diff
@@ -0,0 +1,3 @@
+var jsdom = require('jsdom');
+global.$ = require('jquery')(new jsdom.JSDOM().window);
+
```
willfish: https://forum.sublimetext.com/t/remove-any-extra-blank-lines-at-end-of-file-on-save/6175/3

app/webpacker/src/javascripts/commodities.js
alexdesi: that's a tricky topic,
the only way we were able to make Jest working was by using the commonjs module.
Also, the Jest docs states: 

> Jest ships with experimental support for ECMAScript Modules (ESM).

alexdesi: https://jestjs.io/docs/ecmascript-modules

willfish (approved):

PR #295, opened 2021-10-11
HOTT-1000 Add new top breadcrumbs "Home > Section"

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-1000

### What?
- [x] Add top breadcrumbs to section pages
- [x] Bring in the grey caption
- [x] Make the heading of the page the section number (+ colon) + section title
- [x] Give more spacing around the “Switch” mechanism
- [x] Set the caption for the table that lists the contained chapters and number of those.

### Why?

To improve usability

### DIFF
--- a/app/helpers/service_helper.rb
+++ b/app/helpers/service_helper.rb
@@ -38,18 +38,6 @@ module ServiceHelper
     link_to(t('service_banner.service_name.uk'), current_path)
   end
 
-  def service_switch_banner
-    if is_switch_banner_enabled?
-      tag.div(class: 'tariff-breadcrumbs js-tariff-breadcrumbs clt govuk-!-font-size-15') do
-        tag.nav do
-          tag.p do
-            banner_copy
-          end
-        end
-      end
-    end
-  end
-
   def search_label_text
     t('search.label', service_name: service_name)
   end
@@ -75,6 +63,12 @@ module ServiceHelper
     service_choice == 'uk'
   end
 
+  def switch_banner_copy
+    copy = request.filtered_path == sections_path ? "service_banner.big.#{service_choice}" : 'service_banner.small'
+
+    t(copy, link: switch_service_link).html_safe
+  end
+
 private
 
   def service_name
@@ -85,17 +79,7 @@ private
     t('title.service_description')
   end
 
-  def banner_copy
-    return t("service_banner.big.#{service_choice}", link: switch_service_link).html_safe if request.filtered_path == sections_path
-
-    t('service_banner.small', link: switch_service_link).html_safe
-  end
-
   def current_path
     request.filtered_path.sub("/#{service_choice}", '')
-  end
-
-  def is_switch_banner_enabled?
-    @enable_service_switch_banner_in_action
   end
 end
--- a/app/helpers/service_helper.rb
+++ b/app/helpers/service_helper.rb
@@ -3,8 +3,12 @@ module ServiceHelper
     t('title.default', service_name: service_name, service_description: service_description)
   end
 
-  def default_heading
-    t('h1.default', service_name: service_name, service_description: service_description)
+  def default_heading(section:)
+    section ? section_heading(section) : t('h1.service_description')
+  end
+
+  def section_heading(section)
+    "Section #{section.numeral}: #{section.title}"
   end
 
   def goods_nomenclature_title(goods_nomenclature)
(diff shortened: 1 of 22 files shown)


alexdesi:
New look:
Screenshot 2021-10-12 at 12 33 16

willfish (approved):

app/views/shared/_search.html.erb
```diff
@@ -25,7 +27,8 @@
   </header>
 <% elsif %w(search search_references exchange_rates pages errors feedback).exclude?(controller_name) && action_name != 'quota_search' %>
   <header>
-    <h1 class="govuk-heading-l"><a href="<%= sections_path %>"><%= default_heading %></a></h1>
+    <span class="govuk-caption-m"><%= trade_tariff_heading %></span>
+    <h1 class="govuk-heading-l"><a href="<%= sections_path %>"><%= default_heading(section: @section) %></a></h1>
```
jebw: This could use rails' `link_to` ? Then it will handle param escaping in the sections path

app/helpers/service_helper.rb
```diff
@@ -34,10 +38,17 @@ def switch_service_link
...
 
-  def service_switch_banner(optional_classes: 'govuk-!-margin-bottom-7')
+  def service_switch_banner
     # rubocop:disable Rails/HelperInstanceVariable
     if @enable_service_switch_banner_in_action
-      tag.div(class: "tariff-breadcrumbs js-tariff-breadcrumbs clt govuk-!-font-size-15 #{optional_classes}") do
+      css_classes = %w[tariff-breadcrumbs
+                       js-tariff-breadcrumbs
+                       clt
+                       govuk-!-font-size-15
```
jebw: Would it make sense to `@extend` these in the `tariff-breadcrumbs` css class ? Or better yet `@include govuk-typography-responsive($size: 15);`

app/models/section.rb
```diff
@@ -35,6 +36,12 @@ def to_param
     position.to_s
   end
 
+  delegate :count, to: :chapters, prefix: true
+
+  # TODO: Method deprecated!
+  # Why: ".to_s" should not be use to represent any
```
jebw: Personally I disagree with this - debugging, and 'inspecting' should be handled by `.inspect` - its there for that exact use case?

I'm not a fan of overriding `.to_s`, but in general it should be the 'stringified' version of an object - which may well be just the name ?

app/views/sections/show.html.erb
```diff
@@ -1,12 +1,13 @@
...
+
 <% content_for :head do %>
   <meta name="description" content="<%= @section %>">
   <link rel='alternate' type='application/json' href='<%= section_url(@section, format: :json) %>' title='Section information page in JSON format' />
 <% end %>
 
 <% content_for :bodyClasses, "page--wide" %>
 
-<h2 class="govuk-heading-m">Choose the chapter that best matches your goods</h2>
+<p>Section <span class="govuk-body govuk-!-font-weight-bold"><%= @section.numeral %></span> contains <span class="govuk-body govuk-!-font-weight-bold"><%= @section.chapters_count %></span> chapters. Choose the chapter that best matches your goods.</p>
```
jebw: Does this thist need `govuk-body` on it - `<p>` already has the `govuk-body` class applied automatically?

A `<strong>` tag would give semantic representation as well as visual implementation which may be useful for screen readers?

jebw (approved):
LGTM - I added a couple of suggestions around the CSS and href escaping

app/models/section.rb
willfish: So I didn't have strong views. I basically never use either since in pry you can just do `ls <object reference name>` and get a view on the internal state of the object

willfish: I guess the only advantage of to_s is when implicitly calling to_s on an object that we want to interpolate as part of a string?

willfish: Beyond that I can't see that it would be an issue overriding it.

willfish:
```ruby
section = ... # code to get a section

link_to "Section: #{section}", section_path(section)
```

alexdesi: What this
`link_to "Section: #{section}", section_path(section)`
is going to print?
the title of the section? or maybe the numeral? ...

What about this:
`link_to "Section: #{section.title}", section_path(section)`
it's explicit now.

Anyway, I'll remove the deprecation comment but I suggest to not rely on `to_s` to print the title.

app/views/sections/show.html.erb
alexdesi: 1. yes you're right, I've checked: Screenshot 2021-10-13 at 10 20 54

2.  good point, I saw `govuk-!-font-
weight-bold` in the guideline, but we are actually using the `strong` tag in other pages, I'll change it.

app/helpers/service_helper.rb
alexdesi: I did not change the typography style, only the top/bottom margins.
I checked and that's actually something it's better to fix globally, so I'll specify those in the tariff-breadcrumbs css

PR #308, opened 2021-10-12
Reuse http client

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Adds client builder service
- [x] Extract service chooser
- [x] Initialize client on application start
- [x] Make api entity and request forwarder use reusable client
- [x] Adds missing test coverage
- [x] Adds specs for request forwarder

### Why?

I am doing this because:

- Consolidate the initialisation of the client (easier substitution later).
- Massively reduce the process of instantiatiating and configuring an api client
- Enables the use of persistent connections

### DIFF
--- a/app/services/client_builder.rb
+++ b/app/services/client_builder.rb
@@ -1,0 +1,23 @@
+class ClientBuilder
+  DEFAULT_VERSION = 'v2'.freeze
+  DEFAULT_FORMAT = 'jsonapi'.freeze
+
+  def initialize(service)
+    @service = service
+  end
+
+  def call
+    Faraday.new(host) do |conn|
+      conn.request :url_encoded
+      conn.adapter Faraday.default_adapter
+      conn.response :json, content_type: /\bjson$/
+      conn.headers['Accept'] = "application/vnd.uktt.v#{DEFAULT_VERSION}"
+    end
+  end
+
+  private
+
+  def host
+    TradeTariffFrontend::ServiceChooser.public_send("#{@service}_host")
+  end
+end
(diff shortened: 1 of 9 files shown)


config/initializers/backend.rb
```diff
@@ -0,0 +1,5 @@
+require 'client_builder'
+
+Rails.application.config.http_client_uk = ClientBuilder.new('uk').call if TradeTariffFrontend::ServiceChooser.service_choices.present?
```
jebw: Rather than putting the logic for whether to build a client into the initialiser, how about making `.call` check the service chooser `.service_choices` (it already is for hostname) - and return nil if there aren't any choices?

jebw (commented):
Looks great as a refactor - there's quite a few failing tests that'll need fixing first

If we're refactoring this - does it make sense to make `ServiceChooser` into a class - we could instantiate it on boot (instead of the client builders), and it could in turn hold the references to the 'built clients' returned from client_builder ? Just a thought ?

app/services/client_builder.rb
```diff
@@ -0,0 +1,27 @@
...
+  DEFAULT_FORMAT = 'jsonapi'.freeze
+
+  def initialize(service, forwarding: false)
+    @service = service
+    @forwarding = forwarding
+  end
+
+  def call
+    return nil if TradeTariffFrontend::ServiceChooser.service_choices.blank?
+    return Faraday.new(host) if @forwarding
```
jebw, line 12: Would it make more sense to move the client create for the two scenarios to its own method? eg

```ruby
def call
  return nil if TradeTariff::ServiceChooser.service_choices.nil?

  @forwarding ? forwarding_client : internal_client
end
```

jebw (approved):
LGTM - splitting the `call` method might simplify things a bit?

PR #312, opened 2021-10-12
Rearrange measure table ordering

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1016

### What?

I have added/removed/altered:

- [x] Rearrange xi table and navigation partial ordering
- [x] Rearrange uk table and navigation partial ordering

image

### Why?

I am doing this because:

- The measure positions don't reflect their use/interest


jebw (approved):
LGTM

PR #314, opened 2021-10-13
HOTT-1007: Non breaking space

jebw:
### Jira link

[HOTT-1007](https://transformuk.atlassian.net/jira/software/projects/HOTT/boards/96)

### What?

I have added/removed/altered:

- [x] Added a helper to filter out `nbsp;` tags from the headings content (this will get removed from the source data, at which point this can be dropped)
- [x] Set an explicit width on the heading column, on the normal desktop use case this should avoid wrapping in almost all scenarios
- [x] Added a helper to mark `(word) (number)` sections of headings as unbreakable, which should help the browser to make smarter choices on the break points

### Why?

I am doing this because:

- Currently we have `&nbsp;` tags showing through on the headings for some commodities rules of origin
- These were added to avoid breaking headings at in appropriate points but are simply getting escaped by the frontend
- We do want to avoid breaking at inconvenient points but also want to avoid having presentational markup in the headings in the API response

### DIFF
--- a/app/helpers/rules_of_origin_helper.rb
+++ b/app/helpers/rules_of_origin_helper.rb
@@ -38,7 +38,7 @@ module RulesOfOriginHelper
     content.gsub('&nbsp;', ' ')
   end
 
-  def prevent_breaking_heading_components(content)
+  def restrict_wrapping(content)
     safe_content = html_escape(content)
 
     safe_content.gsub(ROO_NON_BREAKING_HEADING) { |match|
--- a/app/helpers/rules_of_origin_helper.rb
+++ b/app/helpers/rules_of_origin_helper.rb
@@ -1,5 +1,6 @@
 module RulesOfOriginHelper
   ROO_TAGGED_DESCRIPTIONS = %w[CC CTH CTSH EXW WO].freeze
+  ROO_NON_BREAKING_HEADING = /\w+\s+\d+/
 
   def rules_of_origin_service_name
     TradeTariffFrontend::ServiceChooser.uk? ? 'UK' : 'EU'
@@ -31,5 +32,17 @@ module RulesOfOriginHelper
         ''
       end
     end
+  end
+
+  def replace_non_breaking_space(content)
+    content.gsub('&nbsp;', ' ')
+  end
+
+  def prevent_breaking_heading_components(content)
+    safe_content = html_escape(content)
+
+    safe_content.gsub(ROO_NON_BREAKING_HEADING) { |match|
+      tag.span match, class: 'rules-of-origin__non-breaking-heading'
+    }.html_safe
   end
 end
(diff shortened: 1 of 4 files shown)


app/helpers/rules_of_origin_helper.rb
```diff
@@ -32,4 +33,16 @@ def rules_of_origin_tagged_descriptions(content)
       end
     end
   end
+
+  def replace_non_breaking_space(content)
+    content.gsub('&nbsp;', ' ')
+  end
```
alexdesi, line 39: Wondering if we need this as a public method 🤔 
- if I am not wrong `prevent_breaking_heading_components` is the only method that is using  it: `prevent_breaking_heading_components replace_non_breaking_space(rule.heading)`
- if we want to keep it public, since it is a generic method that works for a string, it should not be inside `RulesOfOriginHelper` but in a more generic helper.

app/helpers/rules_of_origin_helper.rb
```diff
@@ -32,4 +33,16 @@ def rules_of_origin_tagged_descriptions(content)
...
+
+  def prevent_breaking_heading_components(content)
+    safe_content = html_escape(content)
+
+    safe_content.gsub(ROO_NON_BREAKING_HEADING) { |match|
+      tag.span match, class: 'rules-of-origin__non-breaking-heading'
+    }.html_safe
+  end
```
alexdesi, line 47: maybe we can keep `replace_non_breaking_space` private and call it from 
`prevent_breaking_heading_components` ?

app/helpers/rules_of_origin_helper.rb
```diff
@@ -32,4 +33,16 @@ def rules_of_origin_tagged_descriptions(content)
       end
     end
   end
+
+  def replace_non_breaking_space(content)
+    content.gsub('&nbsp;', ' ')
+  end
+
+  def prevent_breaking_heading_components(content)
```
alexdesi: also, it  took me a while to understand the meaning of  breaking_heading_components ... but not sure about a better name🤔

app/helpers/rules_of_origin_helper.rb
jebw: _discussed in slack_ concluded we're better keeping them separate

app/helpers/rules_of_origin_helper.rb
jebw: renamed to `restrict_wrapping`

app/helpers/rules_of_origin_helper.rb
jebw: concluded we're best keeping this separate

alexdesi (approved):
🟢 🟢 🟢 🟢 🟢 🟢 🟢 🟢 🟢

PR #315, opened 2021-10-14
HOTT-1034: Upgrade Puma to 5.5.2

jebw:
### Jira link

[HOTT-1034](https://transformuk.atlassian.net/browse/HOTT-1034)

### What?

I have added/removed/altered:

- [x] Upgraded the puma gem to 5.5.2

### Why?

I am doing this because:

- we currently have 5.5.0 which has a minor security issue in it


alexdesi (approved):
✅

PR #318, opened 2021-10-18
Use net_http_persistent adapter

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1044

### What?

I have added/removed/altered:

- [x] Use net-http-persistent in backend clients

### Why?

I am doing this because:

- To reduce request overheads

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 
 orbs:
   aws-cli: circleci/aws-cli@2.0.3
-  ruby: circleci/ruby@1.1.2
+  ruby: circleci/ruby@1.2.0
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2.1
 
 orbs:
   aws-cli: circleci/aws-cli@2.0.3
-  ruby: circleci/ruby@1.1.2
+  ruby: circleci/ruby@1.2.0
```
willfish, line 5: We had some deprecation warnings in the bundling process that are fixed by using a newer version of the script that lives in this updated orb

alexdesi (approved):

PR #331, opened 2021-10-19
HOTT-1045: better error reporting from unparseable response

jebw:
### Jira link

[HOTT-1045](https://transformuk.atlassian.net/browse/HOTT-1045)

### What?

I have added/removed/altered:

- [x] Added a spec for the existing implementation of `TariffJsonapiParser`
- [x] Added a spec for the existing implementation of `ApiEntity`
- [x] Changed TariffJsonapiParser to catch missing keys and raise its own `ParsingError` with more debug info in
- [x] Changed ApiEntity to catch `ParsingError`s and raise `UnparseableResponseError` - this includes the URL which could not be parsed together with the request headers (to find out APIv1 vs APIv2) - the `ParsingError` will continue to be available in `UnparseableResponseError#cause` and Sentry should record this info.

### Why?

I am doing this because:

- I struggled to debug the issue this relates to yesterday and also found we were short of specs around this.

- Makes minor syntactic changes our `ApiEntity` client - which has historically been untested and unchanged accordingly

### DIFF
--- a/lib/tariff_jsonapi_parser.rb
+++ b/lib/tariff_jsonapi_parser.rb
@@ -6,9 +6,10 @@ class TariffJsonapiParser
   def parse
     return @attributes unless @attributes.is_a?(Hash) && @attributes.key?('data')
 
-    if data.is_a?(Hash)
+    case data
+    when Hash
       parse_resource(data)
-    elsif data.is_a?(Array)
+    when Array
       parse_collection(data)
     else
       data
@@ -18,6 +19,8 @@ class TariffJsonapiParser
   def errors
     @attributes['error'] || @attributes['errors']&.map { |error| error['detail'] }&.join(', ')
   end
+
+  class ParsingError < StandardError; end
 
   private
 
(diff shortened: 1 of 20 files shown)


lib/tariff_jsonapi_parser.rb
```diff
@@ -6,9 +6,10 @@ def initialize(attributes)
   def parse
     return @attributes unless @attributes.is_a?(Hash) && @attributes.key?('data')
 
-    if data.is_a?(Hash)
+    case data
```
willfish, line 9: We should be able to completely replace this parser with the one from uktt gem

spec/lib/api_entity_spec.rb
```diff
@@ -0,0 +1,188 @@
+require 'spec_helper'
+
+RSpec.describe ApiEntity do
```
willfish, line 3: Love it

spec/lib/tariff_jsonapi_parser_spec.rb
```diff
@@ -0,0 +1,257 @@
+require 'spec_helper'
+
+RSpec.describe TariffJsonapiParser do
+  describe '#parse' do
+    context 'with singlular resource' do
```
willfish:
```suggestion
    context 'with singular resource' do
```

spec/lib/tariff_jsonapi_parser_spec.rb
```diff
@@ -0,0 +1,257 @@
+require 'spec_helper'
+
+RSpec.describe TariffJsonapiParser do
```
willfish, line 3: Would it be easier to parse fixtures and validate their schema?

willfish: I wonder also if working from a real response would be better.

willfish (approved):

spec/lib/tariff_jsonapi_parser_spec.rb
jebw: Parsing fixtures sounds good, I'll change to that

I'm keen to have scenarios we can control rather than recorded responses. I tend to find VCR style fixtures just bit rot over time and end up not reflecting changing APIs. At least with specific scenarios its easier to tweak if circumstances change.

spec/lib/tariff_jsonapi_parser_spec.rb
jebw: :man_facepalming:

lib/tariff_jsonapi_parser.rb
jebw: In general I think we should definitely be looking to retire this code - was a bit shocked when I found it had no specs at all but this was more about just getting some specs and error reporting in there. I can pull out the `case` vs `if` change - that was just to shut rubocop up?

spec/lib/tariff_jsonapi_parser_spec.rb
willfish: Oh gosh. Yeah, VCR is super painful IMO

lib/tariff_jsonapi_parser.rb
willfish: Yeah

willfish (approved):

PR #332, opened 2021-10-19
Bumps gems - especially faraday

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Bumps faraday
- [x] Bumps other deps

### Why?

I am doing this because:

- We want to validate that faraday has more sensible overheads now that we're reusing persisted connections


jebw (approved):
LGTM

PR #333, opened 2021-10-20
Improve commodities table

alexdesi:
### Jira link
HOTT-Hott 687

### What?

Change the heading of the Start date (End date) column to Date(s)
Change the heading of the “Measure” column to “Measure type”
Change the heading of the “Conditions that apply” column to “Conditions”
Change the heading of the “Value” column to “Duty rate”
Dates should be displayed in a human readable format, i.e. "DD MMM YY"
Prevent the date cell from wrapping (even if it does sometimes take up an inordinate amount of room, e.g where there is a start date & end date

For dates with no end date, just show the start date, e.g. 01 Jan 2021
For dates with a start date and an end date,  show both the start date and the end date, e.g. 01 Jan 2021 to 31 Dec 2021

Optionality of columns: The value column is never needed on the import controls

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -27,10 +27,11 @@ module ApplicationHelper
     tag.div(class: 'govuk-breadcrumbs') { tag.ol(crumbs.join('').html_safe, class: 'govuk-breadcrumbs__list', role: 'breadcrumbs') }
   end
 
-  def govuk_header_navigation_item(active_class = false)
-    base_classname = 'govuk-header__navigation-item'
-    classname = "#{base_classname} #{active_class ? "#{base_classname}--active" : ''}"
-    tag.li(class: classname) { yield }
+  def govuk_header_navigation_item(active_class: '', &block)
+    base_class_name = 'govuk-header__navigation-item'
+    active_class_name = active_class.present? ? "#{base_class_name}--active" : ''
+
+    tag.li(class: "#{base_class_name} #{active_class_name}", &block)
   end
 
   def search_active_class
@@ -50,7 +51,7 @@ module ApplicationHelper
   end
 
   def currency_options
-    [%w[Pound\ sterling GBP], %w[Euro EUR]]
+    [['Pound sterling', 'GBP'], %w[Euro EUR]]
   end
 
   def chapter_forum_url(chapter)
@@ -62,14 +63,8 @@ module ApplicationHelper
   end
 
   def pretty_date_range(start_date, end_date)
-    pretty_end_date = end_date ? "to #{end_date.to_formatted_s(:rfc822)}" : ''
+    pretty_end_date = end_date ? "<br>to #{end_date.to_formatted_s(:rfc822)}" : ''
 
-    start_date.to_formatted_s(:rfc822) + pretty_end_date
-  end
-
-  private
-
-  def search_date_in_future_month?
-    @search&.date.date >= Date.today.at_beginning_of_month.next_month
+    (start_date.to_formatted_s(:rfc822) + pretty_end_date).html_safe
   end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -61,6 +61,12 @@ module ApplicationHelper
     request.base_url + request.path
   end
 
+  def pretty_date_range(start_date, end_date)
+    pretty_end_date = end_date ? "to #{end_date.to_formatted_s(:rfc822)}" : ''
+
+    start_date.to_formatted_s(:rfc822) + pretty_end_date
+  end
+
   private
 
   def search_date_in_future_month?
--- a/app/views/measures/grouped/_table.html.erb
+++ b/app/views/measures/grouped/_table.html.erb
@@ -2,12 +2,12 @@
   <thead>
     <tr>
       <th>Country</th>
-      <th>Measure</th>
-      <th>Value</th>
-      <th>Conditions that apply</th>
+      <th>Measure type</th>
+      <th>Duty rate</th>
+      <th>Conditions</th>
       <th>Exclusions</th>
       <th title="Opens in a new window">Legal base</th>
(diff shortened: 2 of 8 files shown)


app/views/measures/grouped/_table.html.erb
```diff
@@ -2,17 +2,21 @@
   <thead>
     <tr>
       <th>Country</th>
-      <th>Measure</th>
-      <th>Value</th>
-      <th>Conditions that apply</th>
+      <th>Measure type</th>
+
+      <% unless defined?(hide_duty_rate) && hide_duty_rate %>
```
alexdesi: I am aware this is a "double negation",
BUT if we use the direct form, for example: `<% if defined?(show_duty_rate) && show_duty_rate %>`

We would need to pass in the local `show_duty_rate: true` in every place table are displayed.

app/helpers/application_helper.rb
```diff
@@ -27,10 +27,11 @@ def generate_breadcrumbs(current_page, previous_pages)
     tag.div(class: 'govuk-breadcrumbs') { tag.ol(crumbs.join('').html_safe, class: 'govuk-breadcrumbs__list', role: 'breadcrumbs') }
   end
 
-  def govuk_header_navigation_item(active_class = false)
-    base_classname = 'govuk-header__navigation-item'
-    classname = "#{base_classname} #{active_class ? "#{base_classname}--active" : ''}"
-    tag.li(class: classname) { yield }
+  def govuk_header_navigation_item(active_class: '', &block)
```
alexdesi, line 30: `active_class` it's a string (check the caller!), so the default should be a string too, empty as default and not a boolean.

app/views/measures/grouped/_table.html.erb
willfish:
```suggestion
      <% unless try(:hide_duty_rate) %>
```

alexdesi: I tried, but it does not work, it returns nil even if hide_duty_rate is defined in the locals.
I guess `hide_duty_rate` is not considered a method, thus try(:hide_duty_rate) does not work.

Also, `try` is not used in any other views in the frontend, so I do not think that's a viable solution.

A different way would be `local_assigns`:
https://apidock.com/rails/v6.1.3.1/ActionView/Template/local_assigns

but I don't see any advantage on using it.

willfish: So I'd suggest either: 

1. Ask the measure we're rendering whether we should hide the duty rate (`measure.displays_duty?`) or;
2. Just use the same api when rendering every partial

willfish: Also, shouldn't `defined?` take a symbol rather than the result of a method call?

willfish: I think this is a smell that we either need two separate partials or that we need the view to ask the model about whether it should display the duty

app/helpers/application_helper.rb
```diff
@@ -50,7 +51,7 @@ def help_active_class
   end
 
   def currency_options
-    [%w[Pound\ sterling GBP], %w[Euro EUR]]
+    [['Pound sterling', 'GBP'], %w[Euro EUR]]
```
willfish, line 54: I don't think this method is actually used?

app/helpers/application_helper.rb
```diff
@@ -61,9 +62,9 @@ def current_url_without_parameters
     request.base_url + request.path
   end
 
-  private
+  def pretty_date_range(start_date, end_date)
+    pretty_end_date = end_date ? "to #{end_date.to_formatted_s(:rfc822)}" : ''
 
-  def search_date_in_future_month?
```
willfish, line 66: Did you mean to remove this?

app/webpacker/src/stylesheets/_tables.scss
```diff
@@ -118,7 +118,7 @@ table.small-table {
       td:nth-of-type(4) { width: 9% }
```
willfish, line 118: What happened to the option of removing all of these?

willfish: Do they not automatically size based on their contents?

app/views/measures/grouped/_table.html.erb
alexdesi: `defined?` take the name of the variable you want to check, I think if you want to check the existence of a method then you pass the symbol,
in our case, it's fine to pass the name of the local var, notice that it returns "local-variable" and not true/false.

app/helpers/application_helper.rb
alexdesi: it is used by  `/shared/_currency_picker.html.erb` , which is used by views/shared/_search.html.erb

app/helpers/application_helper.rb
alexdesi: yep, I could not find even a single caller in the project

app/webpacker/src/stylesheets/_tables.scss
alexdesi: I tried, but it did not work: https://future-borders.slack.com/archives/C01DXUP15M5/p1634660391069200?thread_ts=1634658227.061300&cid=C01DXUP15M5

image

alexdesi: also, the tables are going to have massive restiling, that's a WIP

app/views/measures/grouped/_table.html.erb
alexdesi: I am not sure if `measure` should know about the way it is going to be represented.
Not sure about "Just use the same api when rendering every partial",
you mean do not make the `hide_duty_rate` optional, so we can remove the `defined?`, I'd be happy with that, the downside is that we'd need to specify that param in every `render ".../table"` 

> So I'd suggest either:
> 
> Ask the measure we're rendering whether we should hide the duty rate (measure.displays_duty?) or;
> Just use the same api when rendering every partial

PR #335, opened 2021-10-20
HOTT-1060: Cache node modules for smoketest runs

jebw:
### Jira link

[HOTT-1060](https://transformuk.atlassian.net/browse/HOTT-1060)

### What?

I have added/removed/altered:

- [x] Cache npm modules in CI
### Why?

I am doing this because:

- because these change infrequently and this should speed up runs when yarn.lock has not changed

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,6 +97,36 @@ commands:
           event: pass
           template: basic_success_1
 
+  smoketest:
+    parameters:
+      url:
+        type: string
+    steps:
+      - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -97,6 +97,36 @@ commands:
           event: pass
           template: basic_success_1
 
+  smoketest:
```
willfish, line 100: I had this thought for a refactor, too

willfish (approved):

PR #337, opened 2021-10-22
HOTT-1064: Show different intro text for GSP countries

jebw:
### Jira link

[HOTT-1064](https://transformuk.atlassian.net/browse/HOTT-1064)

### What?

I have added/removed/altered:

- [x] Added an additional category of intro text - unilateral trade bloc - this will be used for GSP RoO schemes
- [x] Moved the intro partials into a separate subfolder to group them together

### Why?

I am doing this because:

- We need to show different intro text for GSP schemes, this isn't specifically because they are GSP schemes but because they are unilateral in nature. So I'm updating the backend data to mark these schemes as unilateral, and the frontend renders a different partial in this scenario.

- None - the extra partial is only shown if this data is present, if not its nil and the partial is not rendered. It will start showing the different partial once both backend and frontend are deployed.

### Before

Screenshot from 2021-10-22 12-39-31

### After

Screenshot from 2021-10-22 12-38-24


willfish (approved):
Clean

PR #338, opened 2021-10-22
GOV UK style for table.

alexdesi:
### Why?
This PR

-[x] adds the table Caption tag and remove the headers for the tables
-[x] adds many GOV UK styles to the table
-[x] remove zebra-stripes style to the rows
-[x] remove highlight to table on onmouseover


alexdesi:
Before:
Stripes and highlight:
Screenshot 2021-10-26 at 11 22 32


After:
No stripes/highlight AND new Heading
Screenshot 2021-10-27 at 12 04 51

app/webpacker/src/stylesheets/_tables.scss
```diff
@@ -1,59 +1,15 @@
...
-
 table.small-table {
-  @include tariff-zebra();
   margin: 0 0 30px 0;
   p {
     @include govuk-font($size: 14, $weight: normal);
   }
   th, td {
-    font-size: 12px;
+    /*font-size: 12px;*/
```
jebw: These shouldn't be commented out - if they are no-longer required they should be removed.

app/webpacker/src/stylesheets/_tables.scss
```diff
@@ -1,59 +1,15 @@
...
     padding: 0.63158em 0.526315em 0.47368em 0.526315em;
     text-align: left;
-    @media (min-width: $small-table-breakpoint) {
+    /*@media (min-width: $small-table-breakpoint) {
```
jebw: Again  - this commented out block should be removed

jebw (changes requested):
Looks good but commented out CSS code should be removed

One query is I thought the font size was meant to be staying small?

jebw (approved):
LGTM

PR #339, opened 2021-10-22
HOTT-687 Allign Export table to the import ones.

alexdesi:
### What?
Allign Export table to the import ones.
This also allow to reuse the existing partial for the table,
and fix the "Legal base" column (not visible on UK).

### Why?
This is something that was missing from the HOTT-687,
it was confirmed after I merged the PR, so this PR is complementary to the `HOTT-687-improve-commodities-table`.

### DIFF
--- a/app/views/measures/_measures.html.erb
+++ b/app/views/measures/_measures.html.erb
@@ -102,33 +102,16 @@
     <%= render partial: 'declarables/filtered', locals: { search: @search } %>
 
     <% if declarable.export_measures.for_country(@search.country).any? %>
-
       <%= render "shared/measure_types_help", anchor: "export" %>
 
-      <table class="small-table measures govuk-table">
-        <% if @search.filtered_by_country? %>
-          <caption class="govuk-table__caption">Measures for <%= @search.geographical_area %></caption>
-        <% else %>
-          <caption class="govuk-table__caption">Measures for all countries</caption>
-        <% end %>
+      <h3 class="govuk-heading-s govuk-!-margin-bottom-3">
+        <%= @search.filtered_by_country? ? "Measures for #{@search.geographical_area}" : "Measures for all countries" %>
+      </h3>
 
-        <thead>
-        <tr>
-          <th>Country</th>
-          <th>Measure</th>
-          <th>Value</th>
-          <th>Conditions that apply</th>
-          <th>Exclusions</th>
-          <th title="Opens in a new window">Legal base</th>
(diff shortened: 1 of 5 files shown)


app/views/measures/_measures.html.erb
```diff
@@ -104,33 +104,16 @@
...
-          <th>Measure</th>
-          <th>Value</th>
-          <th>Conditions that apply</th>
-          <th>Exclusions</th>
-          <th title="Opens in a new window">Legal base</th>
-          <th>Start date<br>(End date, if any)</th>
-          <th>Footnotes</th>
-        </tr>
-        </thead>
+      <h3 class="govuk-heading-s govuk-!-margin-bottom-3" id='xi_import_controls'>
```
willfish: Confused by xi_import_controls as an id of this component

app/views/measures/_measures.html.erb
```diff
@@ -104,33 +104,16 @@
...
+        <% @search.filtered_by_country? ? "Measures for #{@search.geographical_area}" : "Measures for all countries" %>
+      </h3>
 
-        <tbody>
-          <%= render partial: 'measures/measure', collection: declarable.export_measures.for_country(@search.country).sort_by(&:key) %>
-        </tbody>
-      </table>
+      <%= render partial: 'measures/grouped/table', locals: {
+        collection: declarable.export_measures.for_country(@search.country).sort_by(&:key),
+        hide_duty_rate: true }
```
willfish, line 113: I still think we should get rid of this. Feels inconsistent with the approach taken in the rest of the views for measures

willfish: Also, not all export measures are controls

app/views/measures/_measures.html.erb
alexdesi: 👍 copy/paste error 😄

alexdesi:
Before (on staging):
Screenshot 2021-10-22 at 16 18 08


After - Fixed, on local: all cols caption fixed, Date from/to fixed, "Legal base" fixed.

UK:
Screenshot 2021-10-25 at 11 17 58

XI:
Screenshot 2021-10-25 at 11 18 42

app/views/measures/_measures.html.erb
alexdesi: yes, I agree but it seems it's not as easy as we thought, we had some issues with the table heading (which was not accessible the same way as the table content).
Happy to try again in the coming PRs about the table.

willfish (approved):

PR #340, opened 2021-10-22
HOTT-923: raise errors for unexpected responses

jebw:
### Jira link

[HOTT-923](https://transformuk.atlassian.net/browse/HOTT-923)

### What?

I have added/removed/altered:

- [x] Added in Faradays RaiseError middlewhere - this replaces `ApiEntity`'s error handling with a more complete alternative provided by Faraday

### Why?

I am doing this because:

- I'd observed odd errors from the Frontend - these are almost always of the form that Nil is missing a method - the underlying cause is error responses from the Backend which have status codes the Frontend is not checking for - at which point it is parsed as a successful response even though it isn't and produces a misleading error. Since Faraday provides its own very robust error handling I'm dropping our version.

### Notes

There are still some parts of the codebase raising ApiError's so I've left the existing errors in place and will remove all references in a subsequent PR.

- This is a fundamental change to our API error handling so requires a full regression run in Staging prior to any deployment to Production

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -17,12 +17,13 @@ class ApplicationController < ActionController::Base
 
   layout :set_layout
 
-  rescue_from ApiEntity::Error, Errno::ECONNREFUSED do
+  rescue_from ApiEntity::Error, Faraday::ServerError, Errno::ECONNREFUSED do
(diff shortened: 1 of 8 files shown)


alexdesi (approved):
it looks good ✅

PR #342, opened 2021-10-25
Handle jsonapi metadata and update duty calculator business logic

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1067

### What?

I have added/removed/altered:

- [x] Hides duty calculator link for EPS commodities
- [x] Removes duplicating parse_record method
- [x] Adds support for metadata attribute as per jsonapi spec

### Why?

I am doing this because:

- These currently aren't supported in the duty calculator
- This brings us closer to the UKTT implementation and means we can access data about commodities that reflect aggregated questions on measures

### DIFF
--- a/lib/tariff_jsonapi_parser.rb
+++ b/lib/tariff_jsonapi_parser.rb
@@ -32,8 +32,8 @@ class TariffJsonapiParser
     result = {}
 
     parse_top_level_attributes!(resource, result)
-
     parse_relationships!(resource['relationships'], result) if resource.key?('relationships')
+    parse_meta!(resource, result) if resource.key?('meta')
 
     result
   end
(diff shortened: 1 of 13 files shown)


spec/models/commodity_spec.rb
```diff
@@ -102,43 +110,37 @@
     end
   end
 
-  describe '#no_meursing?' do
```
willfish, line 105: Removed this private method spec since we implicitly test it via the public method call `#calculate_duties?`

lib/tariff_jsonapi_parser.rb
```diff
@@ -73,19 +73,18 @@ def find_and_parse_multiple_included(name, records)
...
+
+    return {} if found_resource.blank?
+
+    parse_resource(found_resource)
   rescue NoMethodError
     raise ParsingError,
           "Error finding relationship - '#{name}', '#{id}', '#{type}': #{record.inspect}"
   end
 
-  def parse_record(record)
```
willfish, line 83: Duplicates parse_resource!

jebw (approved):
LGTM - is it worth including the reasoning for dropping `parse_record` in the relevant commit message?

PR #343, opened 2021-10-25
HOTT-1048: Add /browse path

jebw:
### Jira link

### What?

I have added/removed/altered:

- [x] Added a `/browse` page to eventually replace `/sections`

### Why?

I am doing this because:

- We are splitting up the browse and find commodity pages


app/helpers/application_helper.rb
```diff
@@ -44,6 +44,10 @@ def search_active_class
     'active' if params[:action] == 'search' || (params[:controller] == 'sections' && params[:action] == 'index')
   end
 
+  def browse_active_class
+    'active' if controller_name == 'browse_sections'
```
alexdesi, line 48: 🏅  for using `controller_name` instead of `params[:controller]`,
it'd be great if you can update the other occurrences in this file, to be consistent.

app/views/browse_sections/index.html.erb
```diff
@@ -0,0 +1,58 @@
+<% content_for :head do %>
+  <link rel='alternate' type='application/json' href='<%= sections_url(format: :json) %>' title='Section list in JSON format' />
```
alexdesi: usually, we use the `link_to` to create the links

app/views/browse_sections/index.html.erb
```diff
@@ -0,0 +1,58 @@
...
+<% end %>
+
+<%= page_header 'Browse the tariff' %>
+
+<p>
+  The UK goods classification
+  contains <%= @sections.length %> <%= 'section'.pluralize @sections.length %>,
```
alexdesi: feel free to add a helper for 
`<%= @sections.length %> <%= 'section'.pluralize @sections.length %>`
not very opinionated, it could be useful reused easily and simplify the test (you would not need to render the entire view).
Up to you.

app/views/layouts/_base.html.erb
```diff
@@ -27,6 +27,9 @@
         <%= govuk_header_navigation_item(active_class: search_active_class) do %>
           <%= link_to 'Search or browse the Tariff', sections_path, class: "govuk-header__link" %>
         <% end %>
+        <%= govuk_header_navigation_item(active_class: browse_active_class) do %>
+          <%= link_to 'Browse', browse_sections_path, class: "govuk-header__link" %>
+        <% end if TradeTariffFrontend.split_sections_page? %>
```
alexdesi: I was going to suggest a better name for `split_sections_page`, but I believe we are gonna get rid of it soon,  don't we?

lib/trade_tariff_frontend.rb
```diff
@@ -85,6 +85,11 @@ def revision
     `cat REVISION 2>/dev/null || git rev-parse --short HEAD`.strip
   end
 
+  # TODO: HOTT-1048 - remove once in production
```
alexdesi, line 88: indeed :)

alexdesi (approved):
good one! ✅ 
I suggested few optional improvements

app/helpers/application_helper.rb
jebw: Yeh, will do - I debated whether too or if it was just making work

app/views/browse_sections/index.html.erb
jebw: Yep, good call - this was a cut and paste from the existing page - should have looked at what I was cutting and pasting :+1:

app/views/layouts/_base.html.erb
jebw: Yes - but if we want a different name now's the time :smile: Any suggestions? I was struggling for a good name

app/helpers/application_helper.rb
```diff
@@ -41,23 +41,23 @@ def govuk_header_navigation_item(active_class: '', &block)
...
   end
 
   def a_z_active_class
-    'active' if params[:controller] == 'search_references'
+    'active' if controller_name == 'search_references'
   end
 
   def tools_active_class
-    'active' if params[:action] == 'tools'
+    'active' if action_name == 'tools'
```
alexdesi, line 56: nice one!

app/helpers/application_helper.rb
```diff
@@ -41,23 +41,23 @@ def govuk_header_navigation_item(active_class: '', &block)
   end
 
   def search_active_class
```
alexdesi, line 43: it's funny this class is called `search_active_class`, it seems it search for the active class 😄 
not an issue now, I've just notice it (and also you did not create it in here)

PR #350, opened 2021-10-26
Adds show action to geographical areas controller

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-873

### What?

I have added/removed/altered:

- [x] Adds show action to render a geographical area and its children geographical areas

### Why?

I am doing this because:

- We need this to simplify the show commodity view

image


app/controllers/geographical_areas_controller.rb
```diff
@@ -1,15 +1,23 @@
...
-        text: geographical_area.long_description
-      }
-    end
+    respond_to do |format|
+      format.json do
+        results = geographical_areas.map do |geographical_area|
+          {
+            id: geographical_area.id,
+            text: geographical_area.long_description,
+          }
```
alexdesi: it would be nice to have this outside the controller, maybe into the model or serializer 🤔

app/helpers/declarable_helper.rb
```diff
@@ -1,17 +1,38 @@
 module DeclarableHelper
+  def return_to_declarable_back_link
```
alexdesi: what about just 
`declarable_back_link` ?

app/views/geographical_areas/show.html.erb
```diff
@@ -0,0 +1,24 @@
+<% if current_declarable_code %>
+  <%= return_to_declarable_back_link %>
+<% end %>
```
alexdesi: a style suggestion:
`
<%= declarable_back_link if current_declarable_code %>
`
not important thought.

alexdesi (approved):
It looks good,
I just would avoid mapping in the controller.

PR #351, opened 2021-10-26
Tag slow specs

jebw:
This allows excluding the painfully slow specs from repeat runs where they are not relevant

`bundle exec rspec --tag=~slow`


alexdesi:
🤖 🤖 🤖 😱

alexdesi (approved):
👍

PR #352, opened 2021-10-26
HOTT-1048: refactored page header handling

jebw:
### Jira link

### What?

I have added/removed/altered:

- [x] Generalised the page header handling

### Why?

I am doing this because:

- I'll want to use this on pages which don't have the existing search section on

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -27,6 +27,12 @@ module ApplicationHelper
     tag.div(class: 'govuk-breadcrumbs') { tag.ol(crumbs.join('').html_safe, class: 'govuk-breadcrumbs__list', role: 'breadcrumbs') }
   end
 
+  def page_header(heading, &block)
+    extra_content = block_given? ? capture(&block) : nil
+
+    render 'shared/page_header', heading: heading, extra_content: extra_content
+  end
+
   def govuk_header_navigation_item(active_class: '', &block)
     base_class_name = 'govuk-header__navigation-item'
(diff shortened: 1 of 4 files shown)


spec/helpers/application_helper_spec.rb
```diff
@@ -12,7 +12,7 @@
       end
     end
 
-    context 'string contains Javascript code' do
+    context 'with string contains Javascript code' do
```
alexdesi: maybe it reads better with "when": 
"when string contains Javascript code"
("when string does not contains Javascript code")

app/helpers/application_helper.rb
```diff
@@ -27,6 +27,12 @@ def generate_breadcrumbs(current_page, previous_pages)
     tag.div(class: 'govuk-breadcrumbs') { tag.ol(crumbs.join('').html_safe, class: 'govuk-breadcrumbs__list', role: 'breadcrumbs') }
   end
 
+  def page_header(heading, &block)
+    extra_content = block_given? ? capture(&block) : nil
```
alexdesi, line 31: I've never used `capture`, but it seems appropriate for this case 👍

alexdesi (approved):
it looks good 👍

app/helpers/application_helper.rb
jebw: Its pretty much the exact use case it was designed for - its how a lot of the built in rails helpers work

spec/helpers/application_helper_spec.rb
jebw: Yep, will amend

PR #353, opened 2021-10-27
Move exclusions to country column

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-875

### What?

I have added/removed/altered:

- [x] Link to geographical areas controller to show specific area
- [x] Move exclusions to country column
- [x] Default to auto column widths on smaller windows
- [x] Lint table.scss

### Why?

I am doing this because:

- This is part of an ongoing refreshing/making GDS compliant of the tariff UI

### DIFF
--- a/app/webpacker/src/stylesheets/_tables.scss
+++ b/app/webpacker/src/stylesheets/_tables.scss
@@ -3,14 +3,15 @@ table.small-table {
   p {
     @include govuk-font($size: 14, $weight: normal);
   }
-  th, td {
+  th,
+  td {
     padding: 0.63158em 0.526315em 0.47368em 0.526315em;
     text-align: left;
   }
   td {
     min-height: $table-row-standard-height;
   }
-  a[rel='external']::after {
+  a[rel="external"]::after {
     display: none;
   }
   &.measures {
@@ -22,7 +23,12 @@ table.small-table {
     }
 
     @media (max-width: $small-table-breakpoint - 1) {
-      caption, thead, tbody, th, td, tr {
+      caption,
+      thead,
+      tbody,
+      th,
+      td,
+      tr {
         display: block;
       }
       thead tr {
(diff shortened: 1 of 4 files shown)


willfish:
Screenshot from 2021-10-27 15-55-05

app/webpacker/src/stylesheets/_tables.scss
```diff
@@ -53,33 +59,41 @@ table.small-table {
...
+
+      .legal-base-col::before {
+        content: "Legal base";
+      }
+
+      .dates-col::before {
+        content: "Date(s)";
+      }
 
-    @media (min-width: $small-table-breakpoint) {
```
willfish, line 66: These aren't necessary AFAIK

alexdesi: I tried to remove it, but it breaks the way the "Date(s)" columns wraps the dates.

For example
instead of 
`13 May 2021` (on single line)
you get:
```
13 may
2021
```
at least this is why I did not remove it, have you tested it? (also mobile view)

app/webpacker/src/stylesheets/_tables.scss
```diff
@@ -22,7 +23,12 @@ table.small-table {
     }
 
     @media (max-width: $small-table-breakpoint - 1) {
-      caption, thead, tbody, th, td, tr {
+      caption,
+      thead,
+      tbody,
+      th,
+      td,
+      tr {
```
alexdesi, line 31: I don't see many advantages here to be honest

app/webpacker/src/stylesheets/_tables.scss
```diff
@@ -53,33 +59,41 @@ table.small-table {
...
+        content: "Conditions";
+      }
+
+      .legal-base-col::before {
+        content: "Legal base";
+      }
+
+      .dates-col::before {
+        content: "Date(s)";
+      }
```
alexdesi, line 84: I like this!

alexdesi (commented):
Not sure if we can remove the 

td:nth-of-type(1) { width: 18%; }
...

I am happy to do it, just we should double check

app/webpacker/src/stylesheets/_tables.scss
willfish: This is just the linter doing this

alexdesi (approved):
✅

PR #358, opened 2021-10-28
HOTT-1048: Add the find commodity page

jebw:
### Jira link

### What?

I have added/removed/altered:

- [x] Added a find commodity page
- [x] Redirect the existing sections page to the find commodity page when the feature flag is enabled
- [x] Updated various specs to test with both the feature flag off and on

### Why?

I am doing this because:

- We want to provide separate Find and Browse pages whilst keeping all existing links working

### Notes

Switch service banner is currently missing because that is being worked on separately and this is hidden behind a feature flag

### Screenshot

Screenshot from 2021-10-28 16-25-20

### DIFF
--- a/spec/views/commodities/index.html.erb_spec.rb
+++ b/spec/views/commodities/index.html.erb_spec.rb
@@ -1,0 +1,39 @@
+require 'spec_helper'
+
+RSpec.describe 'commodities/index.html.erb', type: :view do
+  subject { render }
+
+  before { assign :search, search }
+
+  let(:as_of) { Time.zone.today }
+  let(:q) { nil }
+  let(:search) { Search.new q: q, as_of: as_of }
+
+  describe 'header' do
+    it { is_expected.to have_css 'header h1', text: /commodity codes, import duties/ }
+  end
+
+  describe 'search form' do
+    it { is_expected.to have_css 'form input[name="q"]' }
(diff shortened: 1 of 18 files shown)


spec/views/commodities/index.html.erb_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    it { is_expected.to have_css 'form input[name="q"]' }
+    it { is_expected.to have_css 'form details .govuk-details__text' }
+
+    context 'without preset date' do
+      it { is_expected.to have_css %(.govuk-details__text input[name="day"][value="#{as_of.day}"]) }
+      it { is_expected.to have_css %(.govuk-details__text input[name="month"][value="#{as_of.month}"]) }
+      it { is_expected.to have_css %(.govuk-details__text input[name="year"][value="#{as_of.year}"]) }
+    end
+
+    context 'with preset date' do
```
willfish: Is this context a duplicate of the one on line 20? We're setting a search date in both cases

willfish (approved):
Looks clean and awesome. Let's hold off merging until we've deployed the previous lot of features, though

spec/views/commodities/index.html.erb_spec.rb
jebw: It is, but with a different date - I can turn it into a `shared_behaviour` ?

PR #364, opened 2021-11-01
HOTT-1079 Fixes on switch banner

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-1079

### What?
- This fixes the double switch banner issue and hides the banner on the Meursing code finder page (and other pages).
- Also, this makes a small refactor on the shared/search partial

### Why?
Because we hate BUGS!!! :rage1:

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -7,7 +7,7 @@ class ApplicationController < ActionController::Base
   include CookiesHelper
 
   before_action :set_cache
-  before_action :set_enable_service_switch_banner_in_action
+  before_action :set_switch_banner
   before_action :set_last_updated
   before_action :set_path_info
 
@@ -73,8 +73,8 @@ class ApplicationController < ActionController::Base
     # rubocop:enable Naming/MemoizedInstanceVariableName
   end
 
-  def set_enable_service_switch_banner_in_action
-    @enable_service_switch_banner_in_action = true
+  def set_switch_banner
+    @switch_banner = true
   end
 
   def search_invoked?
@@ -116,7 +116,7 @@ class ApplicationController < ActionController::Base
   end
 
   def is_switch_banner_enabled?
-    @enable_service_switch_banner_in_action
+    @switch_banner == true
   end
 
   protected
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -7,7 +7,7 @@ class ApplicationController < ActionController::Base
   include CookiesHelper
 
   before_action :set_cache
-  before_action :set_switch_banner
+  before_action :enable_switch_service_banner
   before_action :set_last_updated
   before_action :set_path_info
 
@@ -49,7 +49,7 @@ class ApplicationController < ActionController::Base
 
   helper_method :cookies_policy,
                 :meursing_lookup_result,
-                :is_switch_banner_enabled?
+                :is_switch_service_banner_enabled?
 
   def render_500
     @no_shared_search = true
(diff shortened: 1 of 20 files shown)


app/controllers/application_controller.rb
```diff
@@ -73,8 +73,8 @@ def set_last_updated
     # rubocop:enable Naming/MemoizedInstanceVariableName
   end
 
-  def set_enable_service_switch_banner_in_action
-    @enable_service_switch_banner_in_action = true
+  def set_switch_banner
```
jebw: I know I'm going to seem really picky here - but `switch_banner` doesn't make any reference to what its switching - a potential point of confusion for future developers.

Can this be renamed to something like `service_switch_banner` which is a bit more specific but still short?

jebw (commented):
Only comment is around naming?

app/views/layouts/application.html.erb
```diff
@@ -76,11 +76,13 @@
       <%= render 'shared/phase_banner' if ENV['NOTICE_BANNER'] %>
       <%= yield :before_main_content %>
 
+      <%= render 'shared/top_breadcrumbs' %>
+      <%= render 'shared/switch_banner' %>
```
alexdesi, line 80: Those partials have been extracted from the `shared/search` partial.
(why did the search contain the breadcrumb and switch? 🤷 )

app/controllers/application_controller.rb
alexdesi: 👍

app/controllers/application_controller.rb
```diff
@@ -73,8 +73,16 @@ def set_last_updated
     # rubocop:enable Naming/MemoizedInstanceVariableName
   end
 
-  def set_enable_service_switch_banner_in_action
-    @enable_service_switch_banner_in_action = true
+  def set_switch_service_banner
```
willfish:
```suggestion
  def enable_switch_service_banner
```

app/controllers/application_controller.rb
```diff
@@ -7,7 +7,7 @@ class ApplicationController < ActionController::Base
   include CookiesHelper
 
   before_action :set_cache
-  before_action :set_enable_service_switch_banner_in_action
+  before_action :set_switch_service_banner
```
willfish:
```suggestion
  before_action :enable_switch_service_banner
```

app/controllers/pages_controller.rb
```diff
@@ -1,6 +1,7 @@
 class PagesController < ApplicationController
   before_action do
     @tariff_last_updated = nil
+
```
willfish: Do you want this extra line :)?

app/controllers/pages_controller.rb
```diff
@@ -15,8 +15,8 @@ def opensearch
   end
 
   def tools
+    set_switch_service_banner
```
willfish:
```suggestion
    enable_switch_service_banner
```

app/controllers/application_controller.rb
jebw: I'll add my +1 to will's suggestion

app/controllers/pages_controller.rb
willfish: Actually isn't this the default so I don't think you need this?

app/controllers/search_controller.rb
```diff
@@ -46,8 +46,6 @@ def quota_search
 
     form = QuotaSearchForm.new(params)
     @result = QuotaSearchPresenter.new(form)
-    @enable_service_switch_banner_in_action = false
```
willfish, line 49: Shouldn't you replace this with a call to `disable_switch_service_banner`

app/views/layouts/application.html.erb
willfish: So much better

app/controllers/search_controller.rb
alexdesi: Good catch! 
Actually here the banner was enabled here, so I removed that line.

(it has been enabled in various places too)

willfish (approved):

PR #372, opened 2021-11-04
Remove breadcrumbs from meursing lookup steps

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1078

### What?

I have added/removed/altered:

- [x] Removed breadcrumbs from meursing lookup form steps

### Why?

I am doing this because:

- This looks weird when combined with a back button


alexdesi (approved):
Looks good!

PR #373, opened 2021-11-04
Fixes the UK/XI context in the Home link in the breadcrumb

alexdesi:
### Jira link
HOTT-1086

### What?
Fixes the UK/XI context in the Home link in the breadcrumb


alexdesi:
Screenshot 2021-11-04 at 13 29 45

PR #374, opened 2021-11-04
HOTT-1087: Support calculating duties on commodities with meursing

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1087

### What?

I have added/removed/altered:

- [x] Removed block on meursing commodities being able to calculate duties

### Why?

I am doing this because:

- We now support this feature in the DC


alexdesi (approved):
nice one

PR #375, opened 2021-11-04
HOTT-1078: Remove breadcrumbs from meursing lookup end step

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1078

### What?

I have added/removed/altered:

- [x] Removes breadcrumbs from end step in meursing lookup

### Why?

I am doing this because:

- With the back link this looks strange


alexdesi (approved):
👍

PR #384, opened 2021-11-05
HOTT-923: Raise errors for unexpected responses

jebw:
### Jira link

[HOTT-923](https://transformuk.atlassian.net/browse/HOTT-923)

### What?

I have added/removed/altered:

- [x] Added in Faradays RaiseError middlewhere - this replaces `ApiEntity`'s error handling with a more complete alternative provided by Faraday

### Why?

I am doing this because:

- I'd observed odd errors from the Frontend - these are almost always of the form that Nil is missing a method - the underlying cause is error responses from the Backend which have status codes the Frontend is not checking for - at which point it is parsed as a successful response even though it isn't and produces a misleading error. Since Faraday provides its own very robust error handling I'm dropping our version.

### Notes

There are still some parts of the codebase raising ApiError's so I've left the existing errors in place and will remove all references in a subsequent PR.

- This is a fundamental change to our API error handling so requires a full regression run in Staging prior to any deployment to Production

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -21,7 +21,7 @@ class ApplicationController < ActionController::Base
     render_500
   end
 
-  rescue_from(ApiEntity::NotFound, Faraday::ResourceNotFound,
+  rescue_from(Faraday::ResourceNotFound,
               ActionView::MissingTemplate, ActionController::UnknownFormat,
               AbstractController::ActionNotFound, URI::InvalidURIError) do |_e|
     request.format = :html
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -16,12 +16,13 @@ class ApplicationController < ActionController::Base
 
   layout :set_layout
 
-  rescue_from ApiEntity::Error, Errno::ECONNREFUSED do
+  rescue_from ApiEntity::Error, Faraday::ServerError, Errno::ECONNREFUSED do
     request.format = :html
     render_500
   end
 
-  rescue_from(ApiEntity::NotFound, ActionView::MissingTemplate, ActionController::UnknownFormat,
+  rescue_from(ApiEntity::NotFound, Faraday::ResourceNotFound,
+              ActionView::MissingTemplate, ActionController::UnknownFormat,
               AbstractController::ActionNotFound, URI::InvalidURIError) do |_e|
     request.format = :html
     render_404
--- a/app/controllers/goods_nomenclatures_controller.rb
+++ b/app/controllers/goods_nomenclatures_controller.rb
@@ -1,5 +1,5 @@
 class GoodsNomenclaturesController < ApplicationController
-  rescue_from ApiEntity::NotFound, with: :find_relevant_goods_code_or_fallback
+  rescue_from ApiEntity::NotFound, Faraday::ResourceNotFound, with: :find_relevant_goods_code_or_fallback
 
   private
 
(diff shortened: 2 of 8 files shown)


app/controllers/application_controller.rb
```diff
@@ -17,12 +17,13 @@ class ApplicationController < ActionController::Base
 
   layout :set_layout
 
-  rescue_from ApiEntity::Error, Errno::ECONNREFUSED do
+  rescue_from ApiEntity::Error, Faraday::ServerError, Errno::ECONNREFUSED do
```
willfish, line 19: If we're intercepting these errors will Sentry receive them? I wonder if it's worth notifying sentry with the full error message?

app/controllers/goods_nomenclatures_controller.rb
```diff
@@ -1,5 +1,5 @@
 class GoodsNomenclaturesController < ApplicationController
-  rescue_from ApiEntity::NotFound, with: :find_relevant_goods_code_or_fallback
+  rescue_from ApiEntity::NotFound, Faraday::ResourceNotFound, with: :find_relevant_goods_code_or_fallback
```
willfish: Same here, really. Should we think about propagating the exception to Sentry?

app/controllers/healthcheck_controller.rb
```diff
@@ -1,7 +1,7 @@
 require 'api_entity'
 
 class HealthcheckController < ActionController::Base
-  rescue_from ApiEntity::Error do |_e|
+  rescue_from ApiEntity::Error, Faraday::ServerError do |_e|
```
willfish, line 4: And here

lib/api_entity.rb
```diff
@@ -98,19 +98,11 @@ def collection(collection_path, opts = {})
       retries = 0
       begin
         resp = api.get(collection_path, opts)
-        case resp.status
-        when 404
-          raise ApiEntity::NotFound, TariffJsonapiParser.new(resp.body).errors
```
willfish, line 103: Can we remove the defined ApiEntity::NotFound error, now?

willfish: And the other ones we've removed?

app/controllers/application_controller.rb
willfish: I think you can remove the reference to ApiEntity::Error, here?

app/controllers/application_controller.rb
```diff
@@ -17,12 +17,13 @@ class ApplicationController < ActionController::Base
...
     request.format = :html
     render_500
   end
 
-  rescue_from(ApiEntity::NotFound, ActionView::MissingTemplate, ActionController::UnknownFormat,
+  rescue_from(ApiEntity::NotFound, Faraday::ResourceNotFound,
```
willfish: I think you can remove the reference to ApiEntity::NotFound, here?

app/controllers/goods_nomenclatures_controller.rb
willfish: I think you can remove the reference to ApiEntity::NotFound, here?

app/controllers/healthcheck_controller.rb
willfish: I think you can remove the reference to ApiEntity::Error, here?

app/presenters/chemical_search_presenter.rb
```diff
@@ -5,7 +5,7 @@ def initialize(search_form)
     @with_errors = false
     @search_form = search_form
     @search_result = Chemical.search(search_form.to_params) if search_form.present?
-  rescue ApiEntity::NotFound
+  rescue ApiEntity::NotFound, Faraday::ResourceNotFound
```
willfish: I think you can remove the reference to ApiEntity::NotFound, here?

app/controllers/application_controller.rb
jebw: There are still some locations which are using ApiEntity::Error - it can be removed when those are.

Re: Sentry - yeh, I don't much like this but was following the existing behaviour - I'd be much happier removing the global rescue though

willfish (approved):

lib/api_entity.rb
jebw: Yeh - ApiEntity::Error is still referenced elsewhere but I'll remove the others

willfish (approved):

PR #386, opened 2021-11-05
HOTT-1048: Add find commodity page

jebw:
### Jira link

### What?

I have added/removed/altered:

- [x] Added a find commodity page
- [x] Redirect the existing sections page to the find commodity page when the feature flag is enabled
- [x] Updated various specs to test with both the feature flag off and on

### Why?

I am doing this because:

- We want to provide separate Find and Browse pages whilst keeping all existing links working

### Notes

Switch service banner is currently missing because that is being worked on separately and this is hidden behind a feature flag

### Screenshot

Screenshot from 2021-10-28 16-25-20


app/controllers/commodities_controller.rb
```diff
@@ -4,6 +4,10 @@ class CommoditiesController < GoodsNomenclaturesController
   before_action :fetch_commodities, only: %i[show]
   before_action :set_session, only: %i[show]
 
+  def index
```
willfish: I'm not sure about this as the action for the find commodity page :thinking:

app/helpers/application_helper.rb
```diff
@@ -41,7 +41,8 @@ def govuk_header_navigation_item(active_class: '', &block)
   end
 
   def search_active_class
-    'active' if controller_name == 'search' || (controller_name == 'sections' && action_name == 'index')
+    'active' if action_name == 'search' ||
```
willfish, line 44: This feels a bit weird/horrible. I know it's what we're inheriting

app/webpacker/src/stylesheets/_feature-panel.scss
```diff
@@ -9,4 +9,8 @@ $feature-panel-border-colour: #333;
   &.last-feature-panel {
     border: 0;
   }
+
+  &--shaded {
+    background-color: #f0f4f5;
```
willfish: Should we explicitly name this color?

app/webpacker/src/javascripts/commodities.js
```diff
@@ -798,7 +798,10 @@
                         data: obj
                       }
                     }]);
-                    $('.js-commodity-picker-select').parents('form:first').trigger("submit");
+
```
willfish, line 801: Do we want to write a test for this change?

app/helpers/application_helper.rb
willfish: Chatted in slack about this. We've got quite a few dependency issues around layouts and controller action options that we can tidy up separately

app/webpacker/src/javascripts/commodities.js
jebw: In principle yes, but the existing spec is skeletal at best. I'll need write a more complete spec for existing behaviour in order to add a spec for this small change. I can look at it this afternoon

app/webpacker/src/stylesheets/_feature-panel.scss
jebw: tbh - we shouldn't be using a colour outside the official palette - it adds nothing for accessibility. Following the designs though

willfish (approved):

PR #392, opened 2021-11-08
HOTT-1097 test loading of accessible autocomplete

jebw:
### Jira link

HOTT-1097

### What?

I have added/removed/altered:

- [x] Added a spec to check initialization of the commodity picker
- [x] Fixed integration between JSDOM context and the test suite context - before document and window didn't seem to work in the documents

### Why?

I am doing this because:

- we want to increase our javascript test coverage

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,9 +38,11 @@
     "webpack-cli": "^3.3.12"
   },
   "scripts": {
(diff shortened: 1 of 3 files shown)


alexdesi (approved):
it looks good!

PR #398, opened 2021-11-09
New context selector

alexdesi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1052

### What?
This adds the new context selector table.

NOTE: Two of the requirements on the tickets has not been implemented, since they are related to other tickets and it's better not to implement those now.

Date selector and Country selector we'll be moved to a different page.

### Why?
to improve UI


app/helpers/commodities_helper.rb
```diff
@@ -33,6 +33,16 @@ def commodity_tree(main_commodity, commodities)
     end
   end
 
+  def classification_description(commodity)
```
willfish: In a declarable helper?

app/helpers/declarable_helper.rb
```diff
@@ -1,4 +1,14 @@
 module DeclarableHelper
+  def classification_description(commodity)
```
willfish, line 2: We should add a test for this guy

app/models/concerns/declarable.rb
```diff
@@ -57,6 +57,22 @@ def all_footnotes
     ).uniq(&:code)
   end
 
+  def supplementary_unit(country)
```
willfish: I think we need a spec for this guy

app/models/measure_component.rb
```diff
@@ -0,0 +1,6 @@
+class MeasureComponent
+  include ApiEntity
+
+  has_one :duty_expression # (p/st)
```
willfish: We don't need these comments

app/presenters/declarable_presenter.rb
```diff
@@ -8,11 +8,11 @@ def initialize(declarable)
   end
 
   def to_s
-    declarable.formatted_description
+    declarable.formatted_description.html_safe
```
willfish: I wonder if we should sanitize :thinking:

app/views/headings/show.html.erb
```diff
@@ -1,22 +1,26 @@
...
 
 <% content_for :bodyClasses, "page--wide" %>
 
+  <%= page_header heading_for(section: @section,chapter: @chapter,heading: @heading, commodity: @commodity) %>
+
+  <%= render 'shared/search/date_picker_form' %>
+  <%= render 'shared/context_tables/heading' %>
+  <%= render 'shared/callouts/heading'%>
+
+<!-- TODO: what's the new UI when the heading is declarable ? -->
```
willfish: Remove me

app/views/headings/show.html.erb
```diff
@@ -37,5 +41,7 @@
     </div>
     <%= render 'footnote', footnotes: @heading.all_footnotes %>
   </article>
-  <%= render 'shared/notes', section_note: @heading.section.section_note, chapter_note: @heading.chapter.chapter_note %>
 <% end %>
+
+<%= render 'shared/notes', section_note: @heading.section.section_note, chapter_note: @heading.chapter.chapter_note %>
```
willfish: Is this correct?

app/views/shared/context_tables/_commodity.html.erb
```diff
@@ -0,0 +1,24 @@
+<table class="govuk-table">
+  <tbody class="govuk-table__body">
+    <tr class="govuk-table__row ">
```
willfish, line 3: I feel like we've all fallen into the trap of treating declarable headings and commodities differently when they should be treated exactly the same. I'm wondering if we should rename this to _declarable.html.erb and just render the declarable or the non-declarable version of this partial.

I'm not picky about it as this is an idiom we're not always using

app/views/shared/search/_date_picker_form.html.erb
```diff
@@ -0,0 +1,13 @@
+<%= form_tag perform_search_path, method: :get, class: "tariff-search #{@section_css}", id: "new_search" do |f| %>
+  <%= yield %>
```
willfish, line 2: What are we yielding?

app/helpers/declarable_helper.rb
alexdesi: 👍

willfish (approved):

PR #393, opened 2021-11-08
Refactor search

alexdesi:
### What?
@jebw and I paired on this to simplify the complex conditions in the _search (now called search_form).

### Why?
This is necessary to work on the coming stories, which require the heading to be outside of the search form.

PR #399, opened 2021-11-10
HOTT-1076 Show commodity code in 4 4 2 pattern

jebw:
### Jira link

[HOTT-1076](https://transformuk.atlassian.net/browse/HOTT-1076)

### What?

I have added/removed/altered:

- [x] Added a helper for format the commodity code into a 4 4 2 split, this will also handle heading or chapter codes if handed those
- [x] Changed the _search partial to use the helper when formatting the heading

### Why?

I am doing this because:

- We want to visually break up the commodity code

### Have you? (optional)

- None

### Notes

This conflicts with the refactor of the `_search` partial in #393 - it should be rebased and merged after that PR.

### Before

Screenshot from 2021-11-10 11-10-49

### After

Screenshot from 2021-11-10 11-10-10

### DIFF
--- a/app/helpers/commodities_helper.rb
+++ b/app/helpers/commodities_helper.rb
@@ -144,4 +144,14 @@ module CommoditiesHelper
         tag.p(commodity.to_s.html_safe)
     end
   end
+
+  def four_four_two_commodity_code(code)
+    return if code.blank?
+
+    parts = code.to_s.gsub(/[^\d]/, '').split('').each_slice(4).map(&:join)
+
+    tag.span class: 'commodity-code-4-4-2' do
+      safe_join parts.map(&tag.method(:span)), "\n"
+    end
+  end
 end
--- a/app/helpers/commodities_helper.rb
+++ b/app/helpers/commodities_helper.rb
@@ -145,12 +145,12 @@ module CommoditiesHelper
     end
   end
 
-  def four_four_two_commodity_code(code)
+  def segmented_commodity_code(code)
     return if code.blank?
 
     parts = code.to_s.gsub(/[^\d]/, '').split('').each_slice(4).map(&:join)
 
-    tag.span class: 'commodity-code-4-4-2' do
+    tag.span class: 'segmented-commodity-code' do
       safe_join parts.map(&tag.method(:span)), "\n"
     end
   end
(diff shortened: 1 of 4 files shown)


app/helpers/commodities_helper.rb
```diff
@@ -144,4 +144,14 @@ def declarable_heading_full(commodity)
         tag.p(commodity.to_s.html_safe)
     end
   end
+
+  def four_four_two_commodity_code(code)
```
alexdesi: the name is a bit cryptic, what about 
`pretty_commodity_code` or `readable_commodity_code`?

app/helpers/commodities_helper.rb
```diff
@@ -144,4 +144,14 @@ def declarable_heading_full(commodity)
...
+    return if code.blank?
+
+    parts = code.to_s.gsub(/[^\d]/, '').split('').each_slice(4).map(&:join)
+
+    tag.span class: 'commodity-code-4-4-2' do
```
alexdesi: I think this should be renamed too: the `4-2-2` part resembles some kind of formatting we are gonna use.

app/views/shared/_search.html.erb
```diff
@@ -2,7 +2,7 @@
   <% if (controller_name == 'commodities' && action_name == 'show') || (controller_name == 'headings' && action_name == 'show' && @heading&.declarable?) %>
     <header>
       <h1 class="govuk-heading-l commodity-header" data-comm-code="<%= commodity_code %>">
-        Commodity information for <%= commodity_code %>
+        Commodity information for <%= four_four_two_commodity_code commodity_code %>
```
alexdesi: this is not the only place to apply prettify the code, is it?

alexdesi (approved):
Nice one!
I am just not sure about the naming of the method.

app/helpers/commodities_helper.rb
jebw: Yeh - naturally I'd call it something like `format_commodity_code` but we're already formatting it several different ways - hence trying to incorporate the format into the name.

How about `segmented_commodity_code` ?

app/views/shared/_search.html.erb
jebw: Ticket requests we do it just in the heading first - then it will be rolled out elsewhere later?

app/helpers/commodities_helper.rb
jebw: Renamed to `segmented_commodity_code`

app/helpers/commodities_helper.rb
jebw: renamed to `.segmented-commodity-code`

alexdesi (approved):
It looks good to me!

PR #403, opened 2021-11-10
Refactor measure collection

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-978

### What?

I have added/removed/altered:

- [x] Refactor the measure collection tests to be more compliant with rspec idioms
- [x] Simplify the measure collection methods
- [x] Avoids mutating the measure collection object everytime we generate a new scope

### Why?

I am doing this because:

- The scope mutation change helps us memoize an instance of the measure collection without having to reinstatiate the root collection everytime. This is especially helpful when improving on the law of demeter violations
- The rest is just common practice improvements


app/models/measure_collection.rb
```diff
@@ -1,90 +1,56 @@
...
-
-  def present?
-    measures.any?
-  end
-
-  private
-
-  def present_with(presenter_klass)
-    measures.map { |entry| presenter_klass.new(entry) }
+    measures.map { |entry| MeasurePresenter.new(entry) }
```
willfish, line 54: I couldn't see any situation in the code where we were **not** using the MeasurePresenter class when interacting with the Measure model via a collection. AFAIK we look at measures on a declarable via:

1. declarable#overview_measures
2. declarable#import_measures
3. declarable#export_measures

app/models/measure_collection.rb
```diff
@@ -1,90 +1,56 @@
...
-  def initialize(measures, presenter_klass = MeasurePresenter)
+  def initialize(measures)
     @measures = measures.clone
-    @presenter_klass = presenter_klass
   end
 
   def for_country(country)
-    self.measures = measures.select { |m| m.relevant_for_country?(country) }
-    self
+    self.class.new(measures.select { |m| m.relevant_for_country?(country) })
```
alexdesi, line 14: this should be the same as 
`new(...)`

alexdesi: and all the other occurrences too.

app/views/measures/grouped/_uk.html.erb
```diff
@@ -1,5 +1,7 @@
+<%- uk_import_measures = uk_declarable.import_measures.for_country(@search.country) %>
+
 <!-- UK IMPORT CONTROLS -->
-<% if (collection = uk_declarable.import_measures.import_controls.for_country(@search.country)).present? %>
+<% if (collection = uk_import_measures.import_controls).present? %>
```
alexdesi: so much cleaner! 👍

app/views/measures/grouped/_uk.html.erb
```diff
@@ -1,5 +1,7 @@
+<%- uk_import_measures = uk_declarable.import_measures.for_country(@search.country) %>
+
 <!-- UK IMPORT CONTROLS -->
-<% if (collection = uk_declarable.import_measures.import_controls.for_country(@search.country)).present? %>
+<% if (collection = uk_import_measures.import_controls).present? %>
   <%= render partial: 'measures/grouped/table', locals: {
     caption: TradeTariffFrontend::ServiceChooser.uk? ? 'Import controls' : 'UK import controls',
     collection: collection.sort_by(&:key),
```
alexdesi: at this point,
the `collection` var seems to be a boolean 🤔 
(from `if (collection = uk_import_measures.import_controls).present?` above)

app/models/measure_collection.rb
willfish: I'll give it a go

app/models/measure_collection.rb
alexdesi: that's reasonable, I should work

app/presenters/measure_presenter.rb
```diff
@@ -36,26 +35,8 @@ def grouped_measure_conditions
                         'quantity'
                       else
                         'default'
-                      end
+                      end,
       }
     end
   end
-
-  def has_references?
```
willfish, line 44: This method is unused so I removed it

app/presenters/measure_presenter.rb
```diff
@@ -1,28 +1,27 @@
...
-  attr_reader :measure
-
-  def initialize(measure)
-    @measure = measure
+class MeasurePresenter < SimpleDelegator
+  def has_children_geographical_areas?
+    geographical_area.children_geographical_areas.any?
   end
 
-  def geo_class
```
willfish, line 8: The second branch of this method is impossible to hit so I've removed it and updated downstream

alexdesi (approved):
it looks good to me

app/models/measure_collection.rb
```diff
@@ -1,90 +1,56 @@
...
 
-  attr_accessor :measures, :presenter_klass
+  attr_accessor :measures
 
+  delegate :present?, to: :measures
   delegate :size, :length, :collect, :map, :each, :all?, :include?, :to_ary, to: :to_a
 
-  def initialize(measures, presenter_klass = MeasurePresenter)
+  def initialize(measures)
     @measures = measures.clone
```
willfish, line 10: We never mutate the measures, here, so we should be able to kill the heavy lifting of cloning the individual measures at some stage

PR #417, opened 2021-11-18
HOTT-334 Commodity picker: prevent duplicate entries

jebw:
### Jira link

[HOTT-334](https://transformuk.atlassian.net/browse/HOTT-334)

### What?

I have added/removed/altered:

- [x] Refactored the commodity picker auto complete to only add the typed query if its not already in the list

### Why?

I am doing this because:

- If the user types a search term which we find an exact match for, they will see duplicate results for the query they typed. This is both confusing for the user and unnecessary.

### DIFF
--- a/app/webpacker/src/javascripts/commodities.js
+++ b/app/webpacker/src/javascripts/commodities.js
@@ -814,13 +814,6 @@
                       let exactMatch = false;
                       options = [];
 
-                      newSource.push(query.toLowerCase());
-                      options.push({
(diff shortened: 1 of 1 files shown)


willfish (approved):
I think it makes sense to test this change but I also realise that's a bit of a PITA. Maybe it'll be easier when we're using stimulus :tada:

jebw:
Yeh, I think rewriting some of these js widgets using stimulus would really improve maintainability

PR #418, opened 2021-11-19
HOTT-1094 Add Browse in to the breadcrumbs

jebw:
### Jira link

[HOTT-1094](https://transformuk.atlassian.net/browse/HOTT-1094)

### What?

I have added/removed/altered:

- [x] Added a browse item to the breadcrumbs on the commodities, headings, chapters, and sections pages

### Why?

I am doing this because:

- Browse sections is no longer the same as home, under the new updated navigation 'home' points to the find_commodity page


alexdesi (approved):
It looks good!

PR #419, opened 2021-11-19
HOTT-1077 Moved breadcrumbs on non-commodity pages

jebw:
### Jira link

[HOTT-1077](https://transformuk.atlassian.net/browse/HOTT-1077)

### What?

I have added/removed/altered:

- [x] Added a new `content_for` location in the page template for breadcrumbs under the top banner  
- [x] Changed breadcrumbs on all content pages to render into that location 

### Why?

I am doing this because:

- When I'd updated the commodities pages earlier I'd not realised we had multiple mechanisms for rendering the breadcrumbs and this hadn't moved other breadcrumbs up under the banner

### Before

Screenshot from 2021-11-19 10-10-15

### After

Screenshot from 2021-11-19 11-09-55

### DIFF
--- a/app/views/browse_sections/index.html.erb
+++ b/app/views/browse_sections/index.html.erb
@@ -3,7 +3,7 @@
                                href: sections_url(format: :json),
                                title: 'Section list in JSON format') %>
 
-<% content_for :before_main_content do %>
+<% content_for :top_breadcrumbs do %>
   <%= generate_breadcrumbs(
       t('breadcrumb.browse'),
       [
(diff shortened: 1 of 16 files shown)


app/views/browse_sections/index.html.erb
```diff
@@ -3,7 +3,7 @@
                                href: sections_url(format: :json),
                                title: 'Section list in JSON format') %>
 
-<% content_for :before_main_content do %>
+<% content_for :top_breadcrumbs do %>
```
alexdesi, line 6: that's a better name 👍

alexdesi (approved):
It looks good! 💚

PR #420, opened 2021-11-19
HOTT-493 Avoid duplicate ids on the search fields

jebw:
### Jira link

[HOTT-493](https://transformuk.atlassian.net/browse/HOTT-493)

### What?

I have added/removed/altered:

- [x] Changed the id of the hidden field in the commodity search forms

### Why?

I am doing this because:

- Currently we have duplicate ids - the hidden field is never referred to by id and the duplicate id may cause accessibility problems

- This does change the code around one of the key entry points to our site - but the change should be safe from a 'logical' perspective and still works correctly in manual testing on both new and old commodity search forms

### DIFF
--- a/app/views/find_commodities/show.html.erb
+++ b/app/views/find_commodities/show.html.erb
@@ -37,11 +37,11 @@
               You must use the right commodity code.
             </div>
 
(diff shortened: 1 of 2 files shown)


alexdesi (approved):
It looks good ✅

PR #421, opened 2021-11-19
Changes logic on the supplementary unit in context tables.

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-1054

### What?
This PR changes the Business logic for the supplementary unit in context tables.

## New logic:
Depending on whether or not a country is selected in filter by country
- If a country is selected …

    Look for any current measures of type 109 or 110 applicable to that country (or a group that contains that country)

    All these measures will be brought back in the API anyway

    - If there is a measure of type 109, then show: in column 1, the words “Supplementary unit”
    - If there is a measure of type 110, then show: in column 1, the words “Supplementary unit (import)”

    In either case, in column 2, show the applicable description from the table measuremet_unit_descriptions that applies to the measurement unit

    If there is also a measurement unit qualifier, then also bring the qualifier back

        this in turn needs a fix making to the API, which is covered in ticket xyz

    Matt to complete this as and when possible

- If a country is not selected …

As above, but look for measures that are assigned to 1011 only - all other logic is identical
Where there is no supplementary unit

If there is no supplementary unit measure (109 or 110) on a commodity, then show this instead

### DIFF
--- a/spec/models/measure_spec.rb
+++ b/spec/models/measure_spec.rb
@@ -159,4 +159,32 @@ RSpec.describe Measure do
       it { is_expected.not_to be_trade_remedies }
     end
   end
+
+  describe '#import?' do
+    context 'when the measure is an import measure' do
+      subject(:measure) { build(:measure, :import) }
+
+      it { is_expected.to be_import }
+    end
+
+    context 'when the measure is an export measure' do
+      subject(:measure) { build(:measure, :export) }
+
+      it { is_expected.not_to be_import }
+    end
+  end
+
+  describe '#export?' do
+    context 'when the measure is an export measure' do
+      subject(:measure) { build(:measure, :export) }
+
+      it { is_expected.to be_export }
+    end
+
+    context 'when the measure is an import measure' do
+      subject(:measure) { build(:measure, :import) }
+
+      it { is_expected.not_to be_export }
+    end
+  end
 end
--- a/spec/models/measure_spec.rb
+++ b/spec/models/measure_spec.rb
@@ -2,71 +2,48 @@ require 'spec_helper'
 
 RSpec.describe Measure do
   describe '#relevant_for_country?' do
-    it 'returns true when no geographical area is specified' do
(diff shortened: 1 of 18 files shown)


willfish:
I'm a little confused by all this. I'll have a read of the ticket when I'm back. I feel like we might be trying to marry supplementary measures that are both import and export (have a direction code on their measure type of 2) and supplementary measures that are just import (have a measure type direction code of 0). So really we have a couple of things here:

```ruby
# measure type id, measure type description
[["110", "Supplementary unit import"],
 ["111", "Supplementary unit export"],
 ["109", "Supplementary unit"]]
```

So I think we need to handle import/export, 0, 1, or multiple and I'm confused about why we're special snowflaking erga omnes

As far as the descriptions go, we're returning everything we need as far as I know unless there's an extra qualifier description which - if I'm honest - I never understood/read about the meaning of the qualifiers and I've done a bunch of work around measure units

alexdesi:
I got your point, but that's what's the ticket says, maybe we should check with Matt. 

About the country not being selected, the ticket states to look for measures that are assigned to 1011.
That's what I tried to do here:
```
def supplementary_measures(country)
    @supplementary_measures ||= begin
      country_or_erga_omnes = country.present? ? country : Measure::DEFAULT_GEOGRAPHICAL_AREA_ID
      import_measures.for_country(country_or_erga_omnes).supplementary
    end
  end
```
Anyway, I think it's better we pair on this when you come back.

willfish (approved):

spec/models/measure_spec.rb
```diff
@@ -2,71 +2,48 @@
...
+      it { expect(measure.relevant_for_country?(nil)).to eq true }
+    end
+
+    context 'when a national measure that is erga omnes' do
+      subject(:measure) { build(:measure, :national, :erga_omnes) }
+
+      it { expect(measure.relevant_for_country?('br')).to eq true }
+    end
+
+    context 'when an excluded geographical area' do
```
alexdesi: the description sounds incomplete:
`when an excluded geographical area`

PR #422, opened 2021-11-19
HOTT-1107: Adds date page

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1107

### What?

**The change dates page defaults to todays date when no query date is supplied**

Screenshot from 2021-11-25 09-45-32

**Basic date validations feed into the form correctly**
Screenshot from 2021-11-25 09-45-49

**Query date populates the form input values**
Screenshot from 2021-11-25 09-46-17

**Commodity context now with link**
Screenshot from 2021-11-25 09-46-53

**Declarable heading context now with link**
Screenshot from 2021-11-25 09-47-15

**Non-declarable heading context now with link**
Screenshot from 2021-11-25 09-47-32

**Chapter context now with link**
Screenshot from 2021-11-25 09-47-46

**Section has no context menu and cannot change the date**
Screenshot from 2021-11-25 09-48-00

I have added/removed/altered:

- [x] Adds change date controller, model, routes and views to enable making changes to the import date
- [x] Migrates breadcrumbs to use home_path
- [x] Adds setting of the goods nomenclature item id in chapter page
- [x] Removes now-unused date forms/partials and their references
- [x] Adds specs covering all of the
- [x] Adds links in context summary lists to change the date
- [x] Lints some touched files

### Why?

I am doing this because:

- This is a feature request and is required to do dates in a compliant way


app/helpers/declarable_helper.rb
```diff
@@ -40,23 +40,28 @@ def declarable_stw_link(declarable, search)
...
+    case current_goods_nomenclature_code&.size
+    when nil
+      home_path
+    when Chapter::SHORT_CODE_LENGTH
+      chapter_path(goods_nomenclature_path_opts)
+    when Heading::SHORT_CODE_LENGTH
+      heading_path(goods_nomenclature_path_opts)
     else
-      commodity_path(declarable_path_opts)
+      commodity_path(goods_nomenclature_path_opts)
```
alexdesi: it looks the path can be a method inside
Chapter, Heading and commodity
and we'd not need the case

app/views/dates/show.html.erb
```diff
@@ -0,0 +1,98 @@
...
+                      ) %>
+              </div>
+            </div>
+          </div>
+        </fieldset>
+      </div>
+
+      <div class="govuk-button-group">
+        <%= submit_tag 'Update date', class: 'govuk-button', name: nil %>
+
```
alexdesi: I am not wrong, Madhu said that this is not working

app/controllers/change_dates_controller.rb
```diff
@@ -0,0 +1,59 @@
+class ChangeDatesController < ApplicationController
```
alexdesi: just calling it `DatesController`  sounds better,
but maybe there is a reason to call it `ChangeDatesController`

app/models/change_date.rb
```diff
@@ -0,0 +1,49 @@
...
+    import_date&.day
+  end
+
+  def month
+    import_date&.month
+  end
+
+  def year
+    import_date&.year
+  end
```
alexdesi: maybe you want to delegate these 3 methods to  import date using allow_nil?

alexdesi (commented):
Just a couple of comments.

app/models/change_date.rb
willfish: Good idea :)

app/controllers/change_dates_controller.rb
willfish: I just picked what Jeremy suggested offline. I originally had DatesController

willfish: I really don't have strong opinions

willfish: So semantically ImportDate would make more sense

app/models/change_date.rb
```diff
@@ -0,0 +1,39 @@
+class ChangeDate
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+  include ActiveRecord::AttributeAssignment
+
+  attribute :import_date, :date
+
+  delegate :day, :month, :year, to: :import_date, allow_nil: true
```
alexdesi: nice 👍

app/controllers/import_export_dates_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+                                    'import_date(2i)': params[:month],
+                                    'import_date(1i)': params[:year],
+                                  }
+                                else
+                                  {
+                                    'import_date(3i)': today.day.to_s,
+                                    'import_date(2i)': today.month.to_s,
+                                    'import_date(1i)': today.year.to_s,
+                                  }
+                                end
```
alexdesi, line 47: this is kind of equivalent:
```
import_export_date_params = {
                                    'import_date(3i)': params[:day] || today.day.to_s,
                                    'import_date(2i)': params[:month] || today.month.to_s,
                                    'import_date(1i)': params[:year] || params[:year],
                                  }
```

alexdesi (approved):

app/controllers/import_export_dates_controller.rb
willfish: I need all three of the query params to be set sadly - I had thought of doing something similar

PR #423, opened 2021-11-22
HOTT-1050: Headings switcher

jebw:
### Jira link

[HOTT-1050](https://transformuk.atlassian.net/browse/HOTT-1050)

### What?

I have added/removed/altered:

- [x] Fixed the size of the caption in headings generated by the `page_header` helper
- [x] Refactored the switch service button to allow for reuse in the header
- [x] Added the switch service button into headings generated by the `page_header` helper method
- [x] Switch all applicable pages to use the `page_header` helper rather than coding their own page headings
- [x] Removed use of the old switch service banner
- [x] Refactored the switch_service_banner configuration to default to on, but still allow turning it off

### Why?

I am doing this because:

- We want to use the new style switch service link and want it to be part of the page heading

### Have you? (optional)

- Low, potentially there may be a page missing a switch service link but I can't find any

### Before

Screenshot from 2021-11-26 15-03-49

### After

Screenshot from 2021-11-26 15-08-46

### DIFF
--- a/app/controllers/find_commodities_controller.rb
+++ b/app/controllers/find_commodities_controller.rb
@@ -1,5 +1,5 @@
 class FindCommoditiesController < ApplicationController
-  before_action :disable_switch_service_banner, only: [:show]
+  before_action :disable_switch_service_banner!, only: [:show]
 
   def show
     @no_shared_search = true
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -34,7 +34,10 @@ module ApplicationHelper
   def page_header(heading = nil, &block)
     extra_content = block_given? ? capture(&block) : nil
 
-    render 'shared/page_header', heading: heading, extra_content: extra_content
+    render 'shared/page_header',
+           heading: heading,
+           show_switch_service: is_switch_service_banner_enabled?,
+           extra_content: extra_content
   end
 
   def govuk_header_navigation_item(active_class: '', &block)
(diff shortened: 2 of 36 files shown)


willfish (approved):
This looks good to me. I'd add a couple of before/after screenshots just to make the changes plain to future devs but otherwise :tada:

app/controllers/find_commodities_controller.rb
```diff
@@ -1,5 +1,5 @@
 class FindCommoditiesController < ApplicationController
-  before_action :disable_switch_service_banner, only: [:show]
+  before_action :disable_switch_service_banner!, only: [:show]
```
alexdesi, line 2: I am happy to use the bang `!` notation, but we haven't used it for the other `disable_...` methods,
should we change it for the others  too?

app/helpers/application_helper.rb
```diff
@@ -34,7 +34,10 @@ def generate_breadcrumbs(current_page, previous_pages)
   def page_header(heading = nil, &block)
     extra_content = block_given? ? capture(&block) : nil
 
-    render 'shared/page_header', heading: heading, extra_content: extra_content
+    render 'shared/page_header',
+           heading: heading,
+           show_switch_service: is_switch_service_banner_enabled?,
```
alexdesi, line 39: do you need to pass `show_switch_service` in?
we use `is_switch_service_banner_enabled?` as global helper, don't we?

alexdesi (approved):
it looks good, 
just a couple of minor comments.

app/helpers/application_helper.rb
jebw: This is rendering a partial - it shouldn't need to rely on global state - hence passing in the state

PR #425, opened 2021-11-22
HOTT-1136 Show XI content on find commodity page

jebw:
Previously we were referring to the UK service whilst on the XI version of the find commodity page

### Jira link

[HOTT-1136](https://transformuk.atlassian.net/browse/HOTT-1136)

### What?

I have added/removed/altered:

- [x] Updated the copy on the find commodity page for the XI service
- [x] Added a `region_name` helper for the above
- [x] Added some missing specs to the existing `service_name` helper

### Why?

I am doing this because:

- We should be referring to Northern Ireland on the XI service

### DIFF
--- a/app/helpers/service_helper.rb
+++ b/app/helpers/service_helper.rb
@@ -82,6 +82,10 @@ module ServiceHelper
     t("service_banner.bottom.#{service_choice}")
   end
 
+  def region_name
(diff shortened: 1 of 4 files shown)


alexdesi (approved):
simple and clean 👍

PR #440, opened 2021-11-25
Show the Headings that are declarable using commodity View

alexdesi:
### What?
This PR shows the Headings that are declarable using commodity View

For example: URL /headings/0903 -> URL commodities/0903000000

### Jira link
https://transformuk.atlassian.net/browse/HOTT-1129

### Why?
To improve User usability and navigation of the commodities and headings.


alexdesi:
When heading is declarable: 
(note: heading is not present in the breadcrumb)
Screenshot 2021-11-26 at 09 20 12
---
When heading is NOT declarable:
Screenshot 2021-11-26 at 09 20 40
:

app/controllers/commodities_controller.rb
```diff
@@ -26,6 +25,17 @@ def commodities_by_code
...
+      heading_id = params[:id].slice(0, 4)
+      heading = HeadingPresenter.new(Heading.find(heading_id, query_params))
+
+      if heading.declarable? # stay in commodity page
+        @commodities[:uk] = heading
+
+        return
+      end
+    end
+
```
alexdesi: This is like a guard_clause, I am thinking to move this to a different `before_action`

app/controllers/commodities_controller.rb
```diff
@@ -26,6 +25,17 @@ def commodities_by_code
   def fetch_commodities
     @commodities ||= {}
 
+    if is_heading_id?
+      heading_id = params[:id].slice(0, 4)
```
willfish: The last 6 digits should be 0 for a heading :+1:

app/controllers/goods_nomenclatures_controller.rb
```diff
@@ -1,6 +1,10 @@
 class GoodsNomenclaturesController < ApplicationController
   rescue_from Faraday::ResourceNotFound, with: :find_relevant_goods_code_or_fallback
 
+  def is_heading_id?
```
jebw: This feels like business logic? Should this be a class method on either the `Heading` or the `GoodsNomenclature models`

jebw: If not, it should probably be in the `private` section

jebw: Just looked further - this is a duplication of the logic in `GoodsNomenclature#heading?` - I think this should definitely become a class method at `GoodsNomenclature.is_heading_id?(id)`, and `GoodsNomenclature#heading?` needs changing to delegate to logic in the class method

jebw (changes requested):
Looks great - only query is the duplicating business logic out the `GoodsNomenclature` model - instead that should instead be refactored/reused

app/controllers/goods_nomenclatures_controller.rb
alexdesi: I agree, I'll dry the logic.

app/controllers/commodities_controller.rb
```diff
@@ -23,17 +22,43 @@ def commodities_by_code
     Commodity.by_code(search_term).sort_by(&:code)
   end
 
+  def fetch_headings_or_commodities
+    if heading_from_param&.declarable?
+      fetch_headings
+    else
+      fetch_commodities
+    end
+  end
```
alexdesi, line 30: I added this to avoid adding another conditional before_action filter, and try to be more explicit.

alexdesi:
The `fetch_commodities` method was getting too complex,
so I added a similar method `fetch_headings` to support it (and I simplified it too).

jebw (approved):
LGTM :+1:

app/helpers/declarable_helper.rb
```diff
@@ -1,6 +1,14 @@
 module DeclarableHelper
   def classification_description(commodity)
```
willfish:
```suggestion
  def classification_description(goods_nomenclature)
```

app/helpers/declarable_helper.rb
```diff
@@ -1,6 +1,14 @@
...
-    descriptions = [commodity.heading, commodity.ancestors, commodity]
+    tree_items = if commodity.heading?
```
willfish:
```suggestion
    tree_items = if goods_nomenclature.heading?
```

app/helpers/declarable_helper.rb
```diff
@@ -1,6 +1,14 @@
...
+                   [commodity]
```
willfish:
```suggestion
                   [goods_nomenclature]
```

app/helpers/declarable_helper.rb
```diff
@@ -1,6 +1,14 @@
...
+                 else
+                   [commodity.heading,
```
willfish:
```suggestion
                   [goods_nomenclature.heading,
```

app/helpers/declarable_helper.rb
```diff
@@ -1,6 +1,14 @@
...
+                    commodity.ancestors,
```
willfish:
```suggestion
                    goods_nomenclature.ancestors,
```

app/helpers/declarable_helper.rb
```diff
@@ -1,6 +1,14 @@
...
+                    commodity]
```
willfish:
```suggestion
                    goods_nomenclature]
```

app/helpers/declarable_helper.rb
willfish: Or declarable

app/controllers/commodities_controller.rb
```diff
@@ -21,6 +21,33 @@ def commodities_by_code
...
+      @commodities[:uk] = heading_from_param
+      @commodities[:xi] = nil
+    else
+      @commodities[:xi] = heading_from_param
+      @commodities[:uk] = TradeTariffFrontend::ServiceChooser.with_source(:uk) { heading_from_param }
+    end
+  end
+
+  def heading_from_param
+    @heading_from_param ||= if GoodsNomenclature.is_heading_id?(params[:id])
```
willfish, line 45: Separately, we should update the places that ask this question for consistency

app/helpers/declarable_helper.rb
willfish: So most of what's being asked here is a bit greedy on the goods_nomenclature model. I wonder if tree_items should be pushed down to the individual model?

willfish: e.g. Heading#tree_items Commodity#tree_items

willfish: Then you don't have to know about whether the declarable is a heading

app/models/goods_nomenclature.rb
```diff
@@ -65,4 +65,8 @@ def rules_of_origin(*args, **kwargs)
   def rules_of_origin_rules(...)
     rules_of_origin(...).flat_map(&:rules)
   end
+
+  def self.is_heading_id?(goods_code_id)
```
willfish:
```suggestion
  def self.is_heading_id?(goods_nomenclature_item_id)
```

app/views/commodities/_header.html.erb
```diff
@@ -1,4 +1,8 @@
 <header>
+  <span class="govuk-caption-m">
+    <%= trade_tariff_heading %>
```
willfish, line 3: I guess this was missing :+1:

willfish (approved):

PR #455, opened 2021-11-29
HOTT-1133: Adds page to pick the filtering country

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1133

### What?

I have added/removed/altered:

- [x] Adds trading partner controller, model, routes and views to enable
      picking a trading partner and redirecting back to the goods
      nomenclature
- [x] Plumb accessible-autocomplete together for the trading partner page
- [x] Adds missing coverage for context table views
- [x] Adds rows to the declarable heading and commodity context tables
- [x] Removes check for declarable on commodity - this is something we
  handle upstream on non-declarable commodities.

**Trading partners no query country**

Screenshot from 2021-11-29 11-45-55

**Trading partners with query country defaults to country**

Screenshot from 2021-11-29 11-46-13

**Chapter gets no filter by country row**

Screenshot from 2021-11-29 11-46-33

**Non declarable heading gets no filter by country row**

Screenshot from 2021-11-29 11-46-40

**Commodity gets filter by country row**

Screenshot from 2021-11-29 11-46-48

**Declarable heading gets filter by country row**

Screenshot from 2021-11-29 11-47-07

### Why?

I am doing this because:

- This is a feature requirement that makes us more compliant and is part of the new ui

### Have you? (optional)

### DIFF
--- a/app/controllers/import_export_dates_controller.rb
+++ b/app/controllers/import_export_dates_controller.rb
@@ -1,7 +1,9 @@
 class ImportExportDatesController < ApplicationController
   include DeclarableHelper
 
-  before_action :disable_search_form
+  before_action :disable_search_form, :disable_switch_service_banner do
+    @tariff_last_updated = nil
+  end
 
   def show
     @import_export_date = ImportExportDate.new(show_import_export_date_params)
--- a/app/models/trading_partner.rb
+++ b/app/models/trading_partner.rb
@@ -1,0 +1,25 @@
+class TradingPartner
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  attribute :country, :string
+
+  delegate :options, to: :class
+
+  validate :validate_country
+
+  def validate_country
+    country.in?(options.map(&:id)).tap do |valid_country|
+      errors.add(:country, :invalid_country) unless valid_country
+    end
+  end
+
+  def self.options
+    GeographicalArea.all.map do |geographical_area|
+      OpenStruct.new(
+        name: geographical_area.long_description,
+        id: geographical_area.id,
+      )
+    end
+  end
+end
--- a/app/views/trading_partners/show.html.erb
+++ b/app/views/trading_partners/show.html.erb
@@ -1,0 +1,40 @@
+<%= page_header %>
+
+<% content_for :before_main_content do %>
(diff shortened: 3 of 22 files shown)


app/controllers/import_export_dates_controller.rb
```diff
@@ -1,7 +1,9 @@
 class ImportExportDatesController < ApplicationController
   include DeclarableHelper
 
-  before_action :disable_search_form
+  before_action :disable_search_form, :disable_switch_service_banner do
```
jebw, line 4: :smile: This is a new trick I didn't know about - symbols and block, nice! definitely one to remember

app/views/trading_partners/show.html.erb
```diff
@@ -0,0 +1,40 @@
...
+        <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
+          <h1 class="govuk-fieldset__heading"><%= t('trading_partner.form.legend_text') %></h1>
+        </legend>
+
+        <%= f.govuk_error_summary %>
+
+        <%= f.govuk_collection_select :country,
+          TradingPartner.options,
+          :id,
+          :name,
```
jebw, line 23: Rather than `:name` - if `TradingPartner.options` was to just return the GeographicalAreas - this could be `:long_description` ?

app/models/trading_partner.rb
```diff
@@ -0,0 +1,25 @@
...
+  validate :validate_country
+
+  def validate_country
+    country.in?(options.map(&:id)).tap do |valid_country|
+      errors.add(:country, :invalid_country) unless valid_country
+    end
+  end
+
+  def self.options
+    GeographicalArea.all.map do |geographical_area|
```
jebw, line 18: If `options` is only going to be used for the form builder, its probably easier to change it to call the `#long_description` method rather than wrapping the array elements in OpenStruct's ?

jebw (approved):
Looks great, the autocomplete code is a world apart from the search box, much much nicer - the only suggestion would be to avoid wrapping the countries array?

app/views/trading_partners/show.html.erb
willfish: I initially tried this. I'll have a gander

PR #457, opened 2021-11-29
Adds GoodsNomenclatureHelper

willfish:
### Jira link

HOTT-<TODO>

### What?

The DeclarableHelper is meant for headings and commodities that are
declarable. Some of the functionality in there was really built for all
goods nomenclature so this extracts that behaviour into its own helper

### Why?

I am doing this because:

- This makes way more sense conceptually

### DIFF
--- a/app/helpers/goods_nomenclature_helper.rb
+++ b/app/helpers/goods_nomenclature_helper.rb
@@ -1,0 +1,45 @@
+module GoodsNomenclatureHelper
+  def goods_nomenclature_back_link
+    link_to('Back', goods_nomenclature_path, class: 'govuk-back-link')
+  end
+
+  def goods_nomenclature_path(path_opts = {})
+    path_opts = goods_nomenclature_path_opts.merge(path_opts) if current_goods_nomenclature_code.present?
+
+    case current_goods_nomenclature_code&.size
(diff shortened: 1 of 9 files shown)


app/helpers/goods_nomenclature_helper.rb
```diff
@@ -0,0 +1,45 @@
+module GoodsNomenclatureHelper
+  def goods_nomenclature_back_link
+    link_to('Back', goods_nomenclature_path, class: 'govuk-back-link')
```
alexdesi, line 3: since all methods have a namespace I am wondering if we need all methods here to be named `goods_nomenclature_...`

app/helpers/goods_nomenclature_helper.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def goods_nomenclature_path(path_opts = {})
```
alexdesi, line 6: maybe we still need it, otherwise, this method should be called only `path`,
which could not make much sense or be in conflict with other methods

alexdesi (approved):
it makes sense,
I added 2 comments, but those are some considerations, a bit like thinking out loud.

PR #458, opened 2021-11-29
HOTT-1176 Faster autocomplete on home page

jebw:
### Jira link

[HOTT-1176](https://transformuk.atlassian.net/browse/HOTT-1176)

### What?

I have added/removed/altered:

- [x] Made the frontend caching of search results more granular - now we have separate cache entries for each of the starting letters - this drops the time to get a search result from ~120ms locally to a consistent 17ms
- [x] Reduced the debounce timeout on the autocomplete field from 300ms to 100ms
- [x] Combined effect is to drop the time from finishing typing, to seeing autocomplete results from approx 420ms down to 120ms

### Why?

I am doing this because:

- The autocomplete search results feel slow and this is an quick win.

- Changes the caching behaviour on the search
- Will likely increase queries to the search suggestions endpoint - probably double (but they're taking quarter the time)

### DIFF
--- a/app/models/search_suggestion.rb
+++ b/app/models/search_suggestion.rb
@@ -2,21 +2,46 @@ require 'api_entity'
 
 class SearchSuggestion
   include ApiEntity
+  CACHE_VERSION = 'v1'.freeze
 
   collection_path '/search_suggestions'
 
   attr_accessor :value
 
-  def self.cached_suggestions
-    TradeTariffFrontend::ServiceChooser.cache_with_service_choice(
-      'cached_search_suggestions', expires_in: 24.hours
-    ) do
-      all
-    end
-  end
+  class << self
+    def cached_suggestions_for(term)
+      return [] if term.blank?
 
-  def self.start_with(term)
-    cached_suggestions.select { |s| s.value =~ /^#{term}/i }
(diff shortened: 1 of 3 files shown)


app/models/search_suggestion.rb
```diff
@@ -2,21 +2,46 @@
...
+
+      if results["#{cache_prefix}loaded"]
+        return results["#{cache_prefix}#{first_letter}"] || []
+      end
+
+      grouped_results = all.group_by { |r| r.value.to_s.downcase.slice(0, 1) }
+      grouped_results.transform_keys! { |letter| "#{cache_prefix}#{letter}" }
+      grouped_results["#{cache_prefix}loaded"] = true
+
+      Rails.cache.write_multi grouped_results, expires_in: 24.hours
```
willfish, line 38:
```suggestion
      Rails.cache.write_multi grouped_results
```

A day is the default

willfish (approved):

app/models/search_suggestion.rb
jebw: Bit split on this one - elsewhere we've been explicit on the expiry, and its replacing code which was explicit about the expiry - I think I'm going to leave it as is for the moment. We do have a specific requirement for how long this is cached for since it ties in with the data sync on the backend.

PR #471, opened 2021-11-30
HOTT-1132 Service tags

jebw:
### Jira link

[HOTT-1132](https://transformuk.atlassian.net/browse/HOTT-1132)

### What?

I have added/removed/altered:

- [x] Renamed the `region_name` helper to `service_region` to clarify usage
- [x] Added a `replace_service_tags` helper to allow inserting the service name/region into the News Item content
- [x] Used the helper on the various places news items are rendered

### Why?

I am doing this because:

- We want a way to insert the Service name into news item content shown on the frontend

### DIFF
--- a/app/helpers/service_helper.rb
+++ b/app/helpers/service_helper.rb
@@ -90,7 +90,7 @@ module ServiceHelper
     t("service_banner.bottom.#{service_choice}")
   end
 
-  def region_name
+  def service_region
     t("title.region_name.#{service_choice}")
   end
 
--- a/app/helpers/service_helper.rb
+++ b/app/helpers/service_helper.rb
@@ -94,6 +94,21 @@ module ServiceHelper
     t("title.region_name.#{service_choice}")
   end
 
+  def replace_service_tags(content)
+    content.gsub %r{\[\[SERVICE_[A-Z_]+\]\]} do |match|
+      case match
+      when '[[SERVICE_NAME]]'
+        service_name
+      when '[[SERVICE_PATH]]'
+        TradeTariffFrontend::ServiceChooser.xi? ? '/xi' : ''
+      when '[[SERVICE_REGION]]'
+        service_region
+      else
+        match
+      end
+    end
+  end
+
 private
 
   def service_name
--- a/app/views/news_items/index.html.erb
+++ b/app/views/news_items/index.html.erb
@@ -17,7 +17,7 @@
         </h2>
 
         <div class="tariff-markdown">
-          <%= govspeak replace_service_tags news_item.paragraphs.first %>
+          <%= govspeak replace_service_tags(news_item.paragraphs.first) %>
         </div>
 
         <%= link_to 'Show more ...', news_item if news_item.paragraphs.many? %>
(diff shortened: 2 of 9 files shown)


app/helpers/service_helper.rb
```diff
@@ -90,10 +90,25 @@ def switch_banner_bottom_copy
...
 
-  def region_name
+  def service_region
     t("title.region_name.#{service_choice}")
   end
 
+  def replace_service_tags(content)
+    content.gsub %r{\[\[SERVICE_[A-Z_]+\]\]} do |match|
+      case match
+      when '[[SERVICE_NAME]]'
```
alexdesi, line 100: I am thinking if we add the `[[` and `]]` in the regexp,
```
when 'SERVICE_NAME'
...
when 'SERVICE_PATH'
...
```
which is more readable

app/views/news_items/index.html.erb
```diff
@@ -17,7 +17,7 @@
         </h2>
 
         <div class="tariff-markdown">
-          <%= govspeak news_item.paragraphs.first %>
+          <%= govspeak replace_service_tags news_item.paragraphs.first %>
```
alexdesi: maybe we can use ( ) here, just for readability:
`govspeak replace_service_tags(news_item.paragraphs.first)`
VS
`govspeak replace_service_tags news_item.paragraphs.first`

app/views/news_items/show.html.erb
```diff
@@ -16,7 +16,7 @@
   </h3>
 
   <div class="tariff-markdown">
-    <%= govspeak @news_item.content %>
+    <%= govspeak replace_service_tags @news_item.content %>
```
alexdesi: same here

spec/helpers/service_helper_spec.rb
```diff
@@ -222,6 +222,106 @@
...
+            * [Find commodity](/find_commodity)
+            * [Browse](/browse)
+          END_OF_EXPECTED
+        end
+
+        it { is_expected.to eql expected }
+      end
+    end
+
+    context 'with XI service' do
```
alexdesi, line 276: The tests in this context are almost a identical of the previous ones, wondering if we can dry this a bit. 
If not it's fine, don't stress 😄

app/views/news_items/index.html.erb
jebw: sure - will look at that

spec/helpers/service_helper_spec.rb
jebw: Yeh - wondered about that _but_ the input and output strings are different each time - I'll see if I can come up with something

app/helpers/service_helper.rb
jebw: After discussion decided this actually ends up more complicated (requires use of RegExp.last_match) so leaving as is

app/views/news_items/show.html.erb
jebw: done

app/views/news_items/index.html.erb
jebw: done

spec/helpers/service_helper_spec.rb
jebw: wrote a `shared_behaviour` but it ended up being 1 line shorter per block and with another layer of indirection so I've left it as is

alexdesi (approved):
✅

PR #476, opened 2021-12-02
Updates home path to find_commodity globally

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Altered references to sections_path to be browse_sections_path

### Why?

I am doing this because:

- Avoids unnecessary redirects
- Cleans up references to an unused path we might get rid of later

### DIFF
--- a/app/helpers/goods_nomenclature_helper.rb
+++ b/app/helpers/goods_nomenclature_helper.rb
@@ -8,7 +8,7 @@ module GoodsNomenclatureHelper
 
     case current_goods_nomenclature_code&.size
     when nil
-      find_commodity_path(path_opts)
+      home_path(path_opts)
     when Chapter::SHORT_CODE_LENGTH
       chapter_path(path_opts)
     when Heading::SHORT_CODE_LENGTH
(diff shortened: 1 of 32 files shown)


app/helpers/goods_nomenclature_helper.rb
```diff
@@ -8,7 +8,7 @@ def goods_nomenclature_path(path_opts = {})
 
     case current_goods_nomenclature_code&.size
     when nil
-      sections_path(path_opts)
+      find_commodity_path(path_opts)
```
jebw: Not sure on this - should this one be going to `browse_sections_path`?

willfish: So the fallback path here should be the home page, yeah, so this is consistent with previous behaviour.

willfish: Unfortunately we can't use root_path

willfish:
Add home_path helper to make it easier to replace essentially the root_path that is hijacked in production

jebw (approved):
LGTM

PR #477, opened 2021-12-02
HOTT-1028: Don't show duplicate RoO links

jebw:
Added `id` generation to allow identifying duplicates

### Jira link

[HOTTT-1028](https://transformuk.atlassian.net/browse/HOTT-1028)

### What?

I have added/removed/altered:

- [x] Added an autogenerated `id` to RulesOfOrigin::Link models
- [x] Use the `RulesOfOrigin::Link#id` to remove duplicates from the sidebar

### Why?

I am doing this because:

- We were showing duplicate links on the side for countries which multiple Rules of Origin Schemes apply to


app/models/rules_of_origin/link.rb
```diff
@@ -1,7 +1,13 @@
 require 'api_entity'
+require 'digest'
```
willfish, line 2: Don't think you need to require this?

willfish (approved):

app/models/rules_of_origin/link.rb
jebw: I'll try it without

jebw: I'm going to leave it - from the look of it we're not requiring elsewhere in the site but Rails does require it so I _think_ thats why it still works. Relying on an implicit require within rails, that wont be autoloaded, feels a bit fragile to me so I'll leave it here. It does no harm and makes the dependency explicit.

willfish:
I'll end up removing it with the zeitwerk change at some stage so all good

PR #478, opened 2021-12-02
HOTT-1143: Migrate to sentry-rails

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1143

### What?

I have added/removed/altered:

- [x] Added monkey patch for sentry rails
- [x] Added config initializer
- [x] Added sentry-rails gem
- [x] Fixed broken specs

### Why?

I am doing this because:

- The sentry-raven gem is now out of support

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -53,7 +53,7 @@ gem 'govuk_design_system_formbuilder', github: 'DFE-Digital/govuk-formbuilder'
 gem 'wizard_steps'
 
 # Sentry
-gem 'sentry-raven'
+gem 'sentry-rails'
 
 group :development do
   gem 'letter_opener'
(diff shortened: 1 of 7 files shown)


jebw (approved):
Looks good - only query would be around the monkey patch

This overrides the monkey-patch included in `sentry-rails` _but_ that does check for `defined(ActiveRecord)` before applying its monkey patch - would we be better not having ActiveRecord defined, presumably this will potentially cause problems in other gems.

https://github.com/getsentry/sentry-ruby/blob/master/sentry-rails/lib/sentry/rails/railtie.rb#L63

Or is this necessary for the date type casting stuff?

willfish:
You've extended my understanding there, yeah. So our issue is that the ActiveRecord constant is defined in the duty calculator and the frontend because of the insane way multiparamter assignment is half-implemented  in ActiveModel::Attributes, yeah. So if we had no need to do the following to enable multiparameters, we'd not require this monkey patch:

```ruby
  include ActiveRecord::AttributeAssignment
```

PR #480, opened 2021-12-06
HOTT-1171 Global search

jebw:
### Jira link

[HOTT-1171](https://transformuk.atlassian.net/browse/HOTT-1171)

### What?

I have added/removed/altered:

- [x] Added a global search banner to all pages
- [x] Added feature flag to toggle between the existing in-page search forms and the header bar based search banner

### Why?

I am doing this because:

- because we want to provide a search feature accessible from all pages whilst also keeping other pages focused on their core purpose


willfish:
This seems to be breaking service switching on dev bizarrely

app/webpacker/src/stylesheets/_search_banner.scss
```diff
@@ -0,0 +1,82 @@
+.hott-search-banner {
```
willfish:
```suggestion
.tariff-search-banner {
```

app/views/layouts/_base.html.erb
```diff
@@ -41,6 +41,23 @@
...
         <% end %>
       </ul>
+
+      <% if TradeTariffFrontend.search_banner? %>
+      <button class="hott-search-banner__toggle" type="button" aria-label="Show or hide the search banner">
+        <span class="govuk-visually-hidden">
+          Search the <%= service_name %>
+        </span>
+
+        <svg class="hott-search-banner__search-icon" width="27" height="27" viewBox="0 0 27 27" fill="none" xmlns="http://www.w3.org/2000/svg" aria-hidden="true" focusable="false">
```
willfish:
```suggestion
        <svg class="tariff-search-banner__search-icon" width="27" height="27" viewBox="0 0 27 27" fill="none" xmlns="http://www.w3.org/2000/svg" aria-hidden="true" focusable="false">
```

app/views/layouts/_base.html.erb
```diff
@@ -41,6 +41,23 @@
...
+          <circle cx="12.0161" cy="11.0161" r="8.51613" stroke="currentColor" stroke-width="3"></circle>
+          <line x1="17.8668" y1="17.3587" x2="26.4475" y2="25.9393" stroke="currentColor" stroke-width="3"></line>
+        </svg>
+
+        <span aria-hidden="true" class="hott-search-banner__close-icon" focusable="false">
```
willfish:
```suggestion
        <span aria-hidden="true" class="tariff-search-banner__close-icon" focusable="false">
```

willfish:
I'll not keep suggesting changes. We can fix hott namespacing after we merge :)

willfish:
So this looks fine to me. I think it would be worth adding a screen shot or gif for easier review/future developers to quickly understand the change. I'd also love to see stimulus being rolled out for this if feasible :tada:

willfish (approved):

PR #481, opened 2021-12-06
HOTT-1066 Review urls

jebw:
### Jira link

### What?

I have added/removed/altered:

- [x] Added a CI step to deploy a review branch version of an app
- [x] Added a CI step to remove orphaned review apps

### Why?

I am doing this because:

- we would like to have separate URLs to view different PRs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,6 +97,77 @@ commands:
           event: pass
           template: basic_success_1
 
+  cf_deploy_review:
+    steps:
+      - run:
+          name: "Setup CF CLI"
+          command: |
+            curl -L -o cf.deb --retry 3 'https://packages.cloudfoundry.org/stable?release=debian64&version=7.2.0&source=github-rel'
+            file cf.deb
+            sudo dpkg -i cf.deb
+            cf -v
+            cf api "$CF_ENDPOINT"
+            cf auth "$CF_USER" "$CF_PASSWORD"
+            cf add-plugin-repo CF-Community https://plugins.cloudfoundry.org
+            cf install-plugin app-autoscaler-plugin -r CF-Community -f
+            cf target -o "$CF_ORG" -s development
+      - run:
+          name: "Fetch existing manifest"
+          command: |
+            cf create-app-manifest "$CF_APP-dev" -p deploy_manifest.yml
+      - run:
+          name: "Update app manifest"
+          command: |
+            sed -i  "s/instances: [0-9]\+/instances\: 1/" deploy_manifest.yml
+      - run:
+          name: "Push review app"
+          command: |
+            export DOCKER_IMAGE=tariff-frontend
+            export DOCKER_TAG=dev-${CIRCLE_SHA1}
+            export PR_NUMBER=$(echo $CIRCLE_PULL_REQUEST | sed 's|https://github.com/||' | cut -d '/' -f 4)
+            export REVIEW_URL_ID="pr${PR_NUMBER}"
+
+            CF_DOCKER_PASSWORD=$AWS_SECRET_ACCESS_KEY cf push "$CF_APP-${REVIEW_URL_ID}" -f deploy_manifest.yml --no-route --docker-image "$ECR_REPO/$DOCKER_IMAGE:$DOCKER_TAG" --docker-username "$AWS_ACCESS_KEY_ID"
+            cf map-route "$CF_APP-$REVIEW_URL_ID" london.cloudapps.digital -n "$CF_APP-$REVIEW_URL_ID"
+
+            # Enable routing from this service to the backend applications which are private
+            cf add-network-policy "$CF_APP-$REVIEW_URL_ID" "$CF_BACKEND_APP_XI-dev" --protocol tcp --port 8080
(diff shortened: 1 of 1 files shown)


alexdesi:
Should we document this in some way?

.circleci/config.yml
```diff
@@ -97,6 +97,77 @@ commands:
           event: pass
           template: basic_success_1
 
+  cf_deploy_review:
+    steps:
+      - run:
+          name: "Setup CF CLI"
+          command: |
+            curl -L -o cf.deb --retry 3 'https://packages.cloudfoundry.org/stable?release=debian64&version=7.2.0&source=github-rel'
+            file cf.deb
```
willfish, line 106: Is this for debug?

.circleci/config.yml
```diff
@@ -97,6 +97,77 @@ commands:
...
+            sudo dpkg -i cf.deb
+            cf -v
```
willfish, line 108: And this

.circleci/config.yml
```diff
@@ -97,6 +97,77 @@ commands:
...
+          name: "Update app manifest"
+          command: |
+            sed -i  "s/instances: [0-9]\+/instances\: 1/" deploy_manifest.yml
+      - run:
+          name: "Push review app"
+          command: |
+            export DOCKER_IMAGE=tariff-frontend
+            export DOCKER_TAG=dev-${CIRCLE_SHA1}
+            export PR_NUMBER=$(echo $CIRCLE_PULL_REQUEST | sed 's|https://github.com/||' | cut -d '/' -f 4)
+            export REVIEW_URL_ID="pr${PR_NUMBER}"
```
willfish, line 128: Do we want to include the branch for semantic helpfuls?

.circleci/config.yml
```diff
@@ -97,6 +97,77 @@ commands:
...
+
+            CF_DOCKER_PASSWORD=$AWS_SECRET_ACCESS_KEY cf push "$CF_APP-${REVIEW_URL_ID}" -f deploy_manifest.yml --no-route --docker-image "$ECR_REPO/$DOCKER_IMAGE:$DOCKER_TAG" --docker-username "$AWS_ACCESS_KEY_ID"
+            cf map-route "$CF_APP-$REVIEW_URL_ID" london.cloudapps.digital -n "$CF_APP-$REVIEW_URL_ID"
+
+            # Enable routing from this service to the backend applications which are private
+            cf add-network-policy "$CF_APP-$REVIEW_URL_ID" "$CF_BACKEND_APP_XI-dev" --protocol tcp --port 8080
```
willfish, line 134: This is all crazy simples :tada:

.circleci/config.yml
```diff
@@ -97,6 +97,77 @@ commands:
...
+          command: |
+            REPO_API_URL=$(echo "${CIRCLE_REPOSITORY_URL}/" | sed 's|.git/$||' | sed 's|git@github.com:|https://api.github.com/repos/|')
+            curl "${REPO_API_URL}/pulls?state=open" | jq '.[].number' > open_prs.txt
+            cat open_prs.txt | sed "s/^/${CF_APP}-pr/" | sort > expected_apps.txt
+            cat expected_apps.txt
+      - run:
+          name: Find old review apps
+          command: |
+            cf apps | cut -d ' ' -f 1  | grep "${CF_APP}-pr" | sort > current_apps.txt
+            (comm -1 --output-delimiter="|" <(cat expected_apps.txt) <(cat current_apps.txt) | egrep -v '^\|' || true) > orphaned_apps.txt
```
willfish, line 162: Awesome

.circleci/config.yml
```diff
@@ -97,6 +97,77 @@ commands:
...
+      - run:
+          name: Remove orphaned apps
+          command: |
+            cat orphaned_apps.txt | while read ORPHANED_APP
```
willfish:
```suggestion
            for ORPHANED_APP in $(cat orphaned_apps.txt)
```

willfish: Opinionated nonsense but I do find ^^ a bit more readable

willfish (approved):
I'll take 5 :dollar:

.circleci/config.yml
jebw: Just duplicated what was already there, but yes - we were getting odd errors from curl - and this was to keep track - though I can come up with a better solution - I'd like to standardise the 'setup CF' step if at all possible, I'll do that as part of a follow up PR

.circleci/config.yml
jebw: As above - just re-using what we already had

.circleci/config.yml
jebw: My take would be a definite no - ultimately we've already got a unique identifier (the PR number) - adding something else in just makes the URL much more complicated so that leads to the question what do we gain by it?

.circleci/config.yml
jebw: Part of me feels this is too much like 'magic' and might be better as Ruby but it works for now

.circleci/config.yml
jebw: :smile:

.circleci/config.yml
jebw: Yeh - totally agree - I'll get that changed - (some my shell scripting definitely gives my age away :laughing: )

PR #487, opened 2021-12-07
Add section to link DIT CHEG service in Export Tab.

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-1158

### What?
Add a section to link DIT CHEG service in Export Tab.

### Why?
Go give precise info to the users.

### Export to EU Country

Screenshot 2021-12-07 at 12 00 53

### Export to non EU Country
Screenshot 2021-12-07 at 12 01 38

### DIFF
--- a/app/views/measures/_measures.html.erb
+++ b/app/views/measures/_measures.html.erb
@@ -98,6 +98,8 @@
   <!-- Export tab -->
   <div class="govuk-tabs__panel" id="export">
     <h2 class="govuk-heading-m"><%= measures_heading(tab: 'export') %></h2>
+    <p>The commodity code for exporting and <%= link_to 'Intrastat reporting', 'https://www.gov.uk/intrastat', target: "_blank" %> is <%= declarable.display_export_code %></p>
+
     <%= render partial: 'declarables/consigned', locals: { declarable: declarable } %>
     <%= render partial: 'declarables/filtered', locals: { search: @search } %>
 
@@ -115,6 +117,12 @@
     <% else %>
       <p>There are no export measures for this commodity on this date.</p>
     <% end %>
+
+    <%= render 'measures/export_tab_check_duties',
+                commodity_code: declarable.code,
+                country_code: @search.country,
+                country_name: @search.geographical_area&.description,
+                european: @search.geographical_area&.eu_member? %>
   </div>
 
   <!-- Rules of Origin tab -->
--- a/spec/models/geographical_area_spec.rb
+++ b/spec/models/geographical_area_spec.rb
@@ -49,4 +49,20 @@ RSpec.describe GeographicalArea do
       it { is_expected.not_to be_erga_omnes }
     end
   end
+
+  describe '#eu_member?', vcr: { cassette_name: 'geographical_areas#1013' } do
+    subject(:geographical_area) { build(:geographical_area, id: geographical_area_id) }
+
+    context 'when the country is part of EU' do
+      let(:geographical_area_id) { 'IT' } # Italy
+
+      it { is_expected.to be_eu_member }
+    end
+
+    context 'when the country is NOT part of EU' do
+      let(:geographical_area_id) { 'AF' } # Afghanistan
+
+      it { is_expected.not_to be_eu_member }
+    end
+  end
 end
--- a/app/models/geographical_area.rb
+++ b/app/models/geographical_area.rb
@@ -1,6 +1,8 @@
 require 'api_entity'
 
 class GeographicalArea
+  EUROPEAN_UNION_ID = '1013'.freeze
+
   include ApiEntity
 
   collection_path '/geographical_areas/countries'
@@ -44,6 +46,10 @@ class GeographicalArea
     end
   end
 
+  def eu_member?
+    id.in?(self.class.eu_members_ids)
+  end
+
   def long_description
     "#{description} (#{id})"
   end
@@ -54,5 +60,11 @@ class GeographicalArea
 
   def erga_omnes?
     id == ERGA_OMNES
+  end
+
+  def self.eu_members_ids
+    find(EUROPEAN_UNION_ID)
+      .children_geographical_areas
+      .map(&:id)
   end
 end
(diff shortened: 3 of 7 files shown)


app/models/search.rb
```diff
@@ -1,6 +1,8 @@
 require 'api_entity'
 
 class Search
+  EUROPEAN_COUNTRIES = %w[AT BE BG CY CZ DE DK EE ES FI FR GR HR HU IE IT LT LU LV MT NL PL PT RO SE SI SK]
```
jebw: Linters complaining about this line - I believe general policy is linters not enforced because of historical code but anything new should pass the linter

app/models/search.rb
```diff
@@ -50,6 +52,10 @@ def geographical_area
     end
   end
 
+  def is_country_european?
```
jebw: This method and its associate constant should be on `GeographicalArea` rather than the Search model

app/views/measures/_measures.html.erb
```diff
@@ -115,6 +117,12 @@
     <% else %>
       <p>There are no export measures for this commodity on this date.</p>
     <% end %>
+
+    <%= render 'measures/export_tab_check_duties',
+                commodity_code: declarable.code,
+                country_code: @search.country,
+                country_name: @search.geographical_area&.description,
+                european: @search.is_country_european? %>
```
jebw:
```
european: @search.geographical_area&.is_country_european? %>
```

jebw (changes requested):
Looks good but I think the logic relates to `GeographicalArea` and not `Search`

app/models/search.rb
willfish: 100% agree with this one

willfish (changes requested):
This all makes sense but we should:

1. Fetch the EU members from the EU geographical area and not use a static constant
2. Move all of the geographical area behaviour to the geographical area model
3. Rename european to something the matches what we're actually describing (e.g. european_union_member?) - technically the UK is still European but not part of the EU.

app/models/search.rb
alexdesi: I agree, here is it missing the `.freeze`, I'll fix it

willfish:
```ruby
# In the backend
GeographicalArea.all.map { |ga| [ga.geographical_area_id, ga.description] }.sort_by(&:first).select { |ga| ga.last == 'European Union' }
=> [["1013", "European Union"], ["EU", "European Union"]]

# In the frontend
class GeographicalArea
  EUROPEAN_UNION = '1013'.freeze

  def eu_member?
    id.in?(eu_members.map(&:id))
  end

  private

  def eu_members
    find(EUROPEAN_UNION).children_geographical_areas
  end
end
```

spec/models/geographical_area_spec.rb
```diff
@@ -49,4 +49,25 @@
       it { is_expected.not_to be_erga_omnes }
     end
   end
+
+  describe '#eu_member?' do
+    let(:geographical_area) { build(:geographical_area, id: geographical_area_id) }
```
willfish:
```suggestion
    subject(:geographical_area) { build(:geographical_area, id: geographical_area_id) }
```

alexdesi: that is just the geographical_area and not the subject under test, `.eu_member?` is 

I tried it, but we need to stub the eu_members_ids before calling the eu_member.
(Alternatively, we should use VRC, which is overkill.)

alexdesi:
thanks for the comments, guys, I've refactored this.

spec/models/geographical_area_spec.rb
willfish: The truth is the subject is not usually a verb (a method call), it's something you interface and send messages too (a noun). We tighten up our tests a bunch by treating the subject as a verb and using `is_expected.to eq('foo')` - we probably shouldn't use this approach as it breaks the standard meaning of subject. Most of the time a subject is either the class or the instance of that class as configured by the constructor.

I'd lean on just using VCR since it's easier/doesn't require mocking the subject

willfish: It's also really quick to do, I've used the `model#id` pattern in the past for the vcr cassette name (`geographical_area#it`)

willfish: https://github.com/trade-tariff/trade-tariff-frontend/blob/5caceda749c2304865a3cf744a47790fb7dc57cd/spec/controllers/geographical_areas_controller_spec.rb#L4

willfish: I've actually already loaded the cassette for the EU, even. So all you need to do is reuse that, here

spec/models/geographical_area_spec.rb
```diff
@@ -49,4 +49,25 @@
...
+    end
+
+    context 'when the country is part of EU' do
+      let(:geographical_area_id) { 'IT' } # Italy
+
+      it { expect(geographical_area.eu_member?).to eq(true) }
+    end
+
+    context 'when the country is NOT part of EU' do
+      let(:geographical_area_id) { 'AF' } # Afganistan
```
willfish:
```suggestion
      let(:geographical_area_id) { 'AF' } # Afghanistan
```

spec/models/geographical_area_spec.rb
```diff
@@ -49,4 +49,25 @@
...
+
+      it { expect(geographical_area.eu_member?).to eq(false) }
```
willfish:
```suggestion
      it { is_expected.not_to be_eu_member }
```

spec/models/geographical_area_spec.rb
```diff
@@ -49,4 +49,25 @@
...
+
+    before do
+      european_countries = %w[AT BE BG CY CZ DE DK EE ES FI FR GR HR HU IE IT LT LU LV MT NL PL PT RO SE SI SK]
+      allow(geographical_area).to receive(:eu_members_ids).and_return(european_countries)
+    end
+
+    context 'when the country is part of EU' do
+      let(:geographical_area_id) { 'IT' } # Italy
+
+      it { expect(geographical_area.eu_member?).to eq(true) }
```
willfish:
```suggestion
      it { is_expected.to be_eu_member }
```

willfish: Not sure how you feel about this but it reads better to me

willfish (approved):
Amazing :tada:

PR #490, opened 2021-12-07
HOTT-1066 Notify slack when we create a review app

jebw:
### Jira link

### What?

I have added/removed/altered:

- [x] Added a slack notification to the Review URLs feature to notify slack when a Review App is deployed

### Why?

I am doing this because:

- We need to know the URLs for the apps being created

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -120,11 +120,15 @@ commands:
           command: |
             sed -i  "s/instances: [0-9]\+/instances\: 1/" deploy_manifest.yml
       - run:
+          name: "Find PR number"
+          command: |
+            PR_NUMBER=$(echo $CIRCLE_PULL_REQUEST | sed 's|https://github.com/||' | cut -d '/' -f 4)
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -119,12 +119,16 @@ commands:
           name: "Update app manifest"
           command: |
             sed -i  "s/instances: [0-9]\+/instances\: 1/" deploy_manifest.yml
+      - run:
+          name: "Find PR number"
+          command: |
+            PR_NUMBER=$(echo $CIRCLE_PULL_REQUEST | sed 's|https://github.com/||' | cut -d '/' -f 4)
+            echo "export PR_NUMBER=${PR_NUMBER}" > $BASH_ENV
```
willfish, line 126: This feels a bit weird to me :tada:

willfish (approved):

.circleci/config.yml
jebw: Yep - definitely - since found you can modify parameters if you define them on a step but since this is well tested I'll stick with this

PR #491, opened 2021-12-08
HOTT-1066 Update JIRA

jebw:
### Jira link

### What?

I have added/removed/altered:

- [x] Added steps to the Deploy Review App to add a JIRA comment if a ticket number is included in the PR

### Why?

I am doing this because:

- We need a way to find the URLs for review apps

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,7 +100,7 @@ commands:
   cf_deploy_review:
     steps:
       - run:
-          name: "Setup CF CLI"
+          name: Setup CF CLI
           command: |
             curl -L -o cf.deb --retry 3 'https://packages.cloudfoundry.org/stable?release=debian64&version=7.2.0&source=github-rel'
             file cf.deb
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -137,6 +137,36 @@ commands:
             # Enable routing from this service to the backend applications which are private
             cf add-network-policy "$CF_APP-$REVIEW_URL_ID" "$CF_BACKEND_APP_XI-dev" --protocol tcp --port 8080
             cf add-network-policy "$CF_APP-$REVIEW_URL_ID" "$CF_BACKEND_APP_UK-dev" --protocol tcp --port 8080
+      - run:
+          name: Find JIRA ticket number
+          command: |
+            REPO_API_URL=$(echo "${CIRCLE_REPOSITORY_URL}/" | sed 's|.git/$||' | sed 's|git@github.com:|https://api.github.com/repos/|')
+            JIRA_TICKET=$(curl --silent --show-error "${REPO_API_URL}/pulls/${PR_NUMBER}" | jq -r '.body' | grep -oP 'HOTT-\d{4,5}' | head -1 || true)
```
willfish, line 144: Is this HOTT- regex a millenium bug?

willfish (approved):

.circleci/config.yml
jebw: Yes in theory - though if it covers up to just 1 less than a 100 thousand tickets so I can't see it being a problem

PR #492, opened 2021-12-08
HOTT-1186: Copy changes for 2022 goods classification

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1186

### What?

I have added/removed/altered:

- [x] Added helper function to enable pulling out 10 digit chapter downloads file info for the relevant files listed under cn2021-2022 in public
- [x] Added helper function to enable pulling out 8 digit correlation file info
- [x] Added support for pulling out the chapter collection to enable speedy retrieval of chapter descriptions
- [x] Added specs for all of the above

### Why?

I am doing this because:

- This approach means we can just drop additional files and have them listed alphabetically in on the classification page
- It also avoids having to statically store chapter descriptions and means we get changes out more quickly for traders


app/helpers/pages_helper.rb
```diff
@@ -0,0 +1,30 @@
+module PagesHelper
+  def cn2021_cn2022_10_digit_files(chapters)
+    Dir.glob('public/cn2021_2022/2022 10 digit codes*').map do |file|
```
jebw, line 3: globbing on spaces feels potentially problematic - can we put the '10 digit' files in their own folder (eg, `2022-10-digit`) then just glob for all files in that folder? Spaces in file names could cause problems for anyone who tries to scrape this without appropriate url escaping - I know this is a bug on their side but names without spaces/weird characters would make this simpler for them?

jebw (approved):
Looks good - my only question would be around using file names with spaces in?

app/helpers/pages_helper.rb
willfish: What I want is for our process to be as simple as possible. I think Matt produces these with spaces, if we can get him to change how they're produced then I'm happy to change this

PR #494, opened 2021-12-08
HOTT-1153: Removes measure type help partial

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1153

### What?

I have added/removed/altered:

- [x] Removed measure type help partial

### Why?

I am doing this because:

- This is no longer required


alexdesi (approved):
it looks good ✅

PR #495, opened 2021-12-08
HOTT-1166: Updates shared stw link copy

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1166

### What?

**Picked a country and see the stw link**

Screenshot from 2021-12-08 16-45-06

**No country picked see the trading partners link**

Screenshot from 2021-12-08 16-51-03

I have added/removed/altered:

- [x] Updated copy of the current stw links and moved it to a partial for easier reading
- [x] Added tariff-inset.scss to reflect common colors/styles for govuk-inset-texts
- [x] Altered meursing form to use the new differentiating inset-text style
- [x] Added a new inset text style for informational insets

### Why?

I am doing this because:

- This is a feature request
- The way we manage insets as part of the measures.scss was a bit confused since insets are a global thing

### DIFF
--- a/app/webpacker/src/stylesheets/_tariff-inset.scss
+++ b/app/webpacker/src/stylesheets/_tariff-inset.scss
@@ -1,8 +1,3 @@
-$tariff-inset-background-color-meursing:  #e9f4f9;
-$tariff-inset-border-left-color-meursing: #098dc1;
-
-$tariff-inset-background-color-information:  #f0f4f5;
-
 .tariff-inset-meursing {
   background-color: $tariff-inset-background-color-meursing;
   border-left: 10px solid $tariff-inset-border-left-color-meursing;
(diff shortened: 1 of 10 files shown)


app/webpacker/src/stylesheets/_tariff-inset.scss
```diff
@@ -0,0 +1,14 @@
+$tariff-inset-background-color-meursing:  #e9f4f9;
+$tariff-inset-border-left-color-meursing: #098dc1;
+
+$tariff-inset-background-color-information:  #f0f4f5;
```
alexdesi: should we move these colour definitions to `stylesheets/_colors.scss`?

willfish: I think you're right. I'll put those in _colors.scss

jebw (approved):
Looks good - only query would be adding more colours but I know there's separate work going on to standardise these.

PR #497, opened 2021-12-09
Move to zeitwerk

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1193

### What?

I have added/removed/altered:

- [x] Migrate us to zeitwerk

### Why?

I am doing this because:

- This enables an upgrade to rails 7
- And makes loading constants simpler


jebw (approved):
LGTM - well that was a quicker to review then I expected :wink:

PR #498, opened 2021-12-09
HOTT-1194: Handle broken date type implementation 

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1194

### What?

I have added/removed/altered:

- [x] Handle invalid dates during Date type casting more gracefully

### Why?

I am doing this because:

- This was meant to resolve our need for the sentry-rails monkey patch by
  removing `include ActiveRecord::AttributeAssignment` (having ActiveRecord
  loaded in the AST triggers preloading connections before sending telemetry
  as part of an optimisation for serializing ActiveRecord model instances) but
  unfortunately was not able to achieve this
- It does achieve a significant tidy up in the api of the ImportExportDate model
  however

### DIFF
--- a/app/controllers/import_export_dates_controller.rb
+++ b/app/controllers/import_export_dates_controller.rb
@@ -12,14 +12,14 @@ class ImportExportDatesController < ApplicationController
   def update
     @import_export_date = ImportExportDate.new(update_import_export_date_params)
 
-    if @import_export_date.errors.messages.any?
-      render 'show'
-    else
+    if @import_export_date.valid?
       redirect_to goods_nomenclature_path(
         day: @import_export_date.day,
         month: @import_export_date.month,
         year: @import_export_date.year,
       )
+    else
+      render 'show'
     end
   end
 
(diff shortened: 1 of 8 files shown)


jebw (changes requested):
Code is a nice clean up but I'm getting an odd behaviour

Previously - changing to 30 Feb 2021 would result in a validation failure
Now - it just swaps back to todays date?

willfish:
Weird. I tested this with Madhu. Will have a play tomorrow

jebw (changes requested):
This isn't letting me change either month or year? Day does work though?

Just silently fails and goes back to todays date?

willfish:
I can't reproduce that myself.

willfish:
Debugging this specific commodity https://dev.trade-tariff.service.gov.uk/commodities/2915704095 it doesn't exist on months earlier than December so changing the date causes a redirect to today when landing on the commodity show action

jebw (changes requested):
Code looks great and works well. Can we get a simple spec for the type - eg just checking behaviour with valid and invalid dates

jebw (approved):
Looks great!

PR #502, opened 2021-12-13
HOTT-1155: Update cache policy for cookies

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1155

### What?

I have added/removed/altered:

- [x] Added explicit cache policy in application controller
- [x] Added specs to cover the new policy

### Why?

I am doing this because:

- This is part of working through the issues we've been seeing in production cookie sharing between multiple responses from different users

```
The no-store response directive indicates that any caches of any kind (private or shared) should not store this response.
```

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -111,7 +111,9 @@ class ApplicationController < ActionController::Base
   end
 
   def set_cache
-    expires_now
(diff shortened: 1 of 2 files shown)


alexdesi (approved):
it's hard to say if this is gonna work, probably it's better to merge and try it on staging.

PR #512, opened 2021-12-14
HOTT-1195 coloured breadcrumb

jebw:
### Jira link

[HOTT-1195](https://transformuk.atlassian.net/browse/HOTT-1195)

### What?

I have added/removed/altered:

- [x] Added a 'colour' option to the segmented comm codes
- [x] Added a helper to shorten comm codes
- [x] Added the Azeret Mono font into webpacker, together with a common `_fonts.scss` definition file, much like the existing `_colours.scss`
- [x] Added a tree like view of the declarable's ancestry, using `aria-owns` attribute to define the relationships
- [x] Removed the background shaded panel on country picker
- [x] Removed the ancestors from the classification row above the ancestor list now that we have the ancestor list

### Why?

I am doing this because:

- our users value the hierarchical view of the commodity ancestry and want it to return in a more accessible form

### Have you? (optional)

- low technical risk but high user visibility

### Screenshots

Screenshot from 2021-12-14 14-25-28
Screenshot from 2021-12-14 14-24-34
Screenshot from 2021-12-14 14-24-15


app/helpers/commodities_helper.rb
```diff
@@ -110,13 +110,28 @@ def declarable_heading_full(commodity)
...
+    code = code.to_s
+    case code.gsub(/0*\z/, '').length
+    when 7..8
+      code.slice(0, 8)
+    when 5..6
+      code.slice(0, 6)
+    else
+      code
+    end
+  end
```
alexdesi, line 136: not sure if it can help,
I wrote the `commodity_code_for_check_duties_service` : 
https://github.com/trade-tariff/trade-tariff-frontend/blob/b660f45024ed63fd7692d1550fa44344c03e9518/app/models/concerns/declarable.rb#L77

but maybe the logic is slightly different.

app/views/commodities/_ancestors.html.erb
```diff
@@ -0,0 +1,102 @@
...
+
+          <span class="commodity-ancestors__descriptor">
+            <%= declarable.heading.description %>
+          </span>
+        <% end %>
+      </li>
+
+      <%# list each of commodities ancestors %>
+      <% declarable.ancestors.each.with_index do |ancestor, index| %>
+      <li id="commodity-ancestors__ancestor-<%= index + 1 %>"
```
alexdesi: it would be nice to have `commodity-ancestors__ancestor-?`  in a helper,
that is used in many places, 
also, should we check a max index?

spec/views/commodities/_ancestors.html.erb_spec.rb
```diff
@@ -0,0 +1,54 @@
+require 'spec_helper'
+
+RSpec.describe 'rules_of_origin/_ancestors.html.erb', type: :view do
+  subject(:rendered_page) { render_page && rendered }
+
+  let(:render_page) { render 'commodities/ancestors', declarable: declarable }
+
+  context 'with commodity' do
+    let :declarable do
+      VCR.use_cassette 'commodities_2208909110' do
```
alexdesi: often we specify the cassette in the test or context 
`context 'with commodity', vcr: { cassette_name: ' ... '} do`

it seems a bit cleaner, but I not very opinionated on that anyway

alexdesi (approved):
the PR with lot's of html/css are difficult to review, 
this looks good to me anyway

app/helpers/commodities_helper.rb
jebw: I think I'll standardise this in a separate PR

app/views/commodities/_ancestors.html.erb
jebw: done

spec/views/commodities/_ancestors.html.erb_spec.rb
jebw: done

PR #515, opened 2021-12-14
HOTT-1183 Fix section table col width containing the section ID

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-1183

### What?
Fix spacing between chapter number and description (see screenshot)

image

Fixed version:
Now the col with is fixed, non-dependant on the length of the description.

Screenshot 2021-12-14 at 17 42 26


willfish (approved):
LGTM but we're holding off merging until after Xmas

PR #517, opened 2021-12-15
Exposes backend healthcheck api

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Expose backend healthcheck api

### Why?

I am doing this because:

- I want to be able to validate the current sha that is deployed on the backend without looking in the circle ci dashboard


jebw (approved):
LGTM

PR #526, opened 2021-12-16
HOTT-1208: Adds coverage for measures.html template rendering

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1208

### What?

I have added/removed/altered:

- [x] Added coverage for template rendering (especially measure/_references)

### Why?

I am doing this because:

- We missed the removal of this in a recent PR and want to avoid regressions


jebw (approved):
LGTM

Just cherry-picked onto `37efb2d9` (ie pre fix) and it fails there :+1:

PR #535, opened 2021-12-20
Remove $, @, and other special symbols from description.

alexdesi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1211

### What?
In many entries like $, @ and other special chars are used in place of HTML tag `<sup>` or `<sub>` in the good nomenclatures description entries.

### Why?
To make the description clearly readable.

Issue:
image

Fix 1:
Screenshot 2021-12-20 at 11 22 04

Fix 2:
Screenshot 2021-12-20 at 11 21 54

### DIFF
--- a/app/views/commodities/_ancestors.html.erb
+++ b/app/views/commodities/_ancestors.html.erb
@@ -92,7 +92,7 @@
           </span>
 
           <span class="commodity-ancestors__descriptor">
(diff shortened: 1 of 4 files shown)


alexdesi:
Opt, wrong HOTT number, this should be HOTT-1211

willfish (approved):

PR #536, opened 2021-12-20
HOTT-1145 Fix bad encoding error

jebw:
### Jira link

[HOTT-1145](https://transformuk.atlassian.net/browse/HOTT-1145)

### What?

I have added/removed/altered:

- [x] Changed the existing bad encoding middleware to not try to process the badly encoded query string into JSON

### Why?

I am doing this because:

- If the string is mis-encoded, and processing it causes an error - trying to process it again in the error handler is a bad idea.

### Notes

No test case because I haven't been able to replicate.

Review carefully because this is a rack middleware and this has no specs - though to my eyes this should be a safe fix.

### DIFF
--- a/lib/trade_tariff_frontend.rb
+++ b/lib/trade_tariff_frontend.rb
@@ -128,7 +128,7 @@ module TradeTariffFrontend
             {
               status: @status.to_s,
               title: 'There was a problem with your query',
-              source: { parameter: @query_string },
+              source: { parameter: 'Invalid query string' },
             },
           ],
         }.to_json,


alexdesi (approved):
✅ What about adding some logging message that could be handy for debugging too?

jebw:
I think the problem is a client (not our frontend) thats passing UTF8 in the query string without URI encoding the UTF8 characters - our problem is _any_ attempt to process that string will trigger the error. I think the best solution is to just drop the bad request.

willfish (approved):

PR #540, opened 2021-12-21
HOTT-1174 Fixes in suggestions

alexdesi:
### What?
This PR fixes:
1) suggestion called with beginning spaces.
2) 500 Internal Server Error when suggestions are not found.

### Jira link
https://transformuk.atlassian.net/browse/HOTT-1174

### Why?
We don't like bugs!

Screenshot 2021-12-21 at 12 54 10

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -37,7 +37,7 @@ class SearchController < ApplicationController
   end
 
   def suggestions
-    search_term = Regexp.escape(params[:term].to_s)
+    search_term = Regexp.escape(params[:term].to_s.strip)
     start_with = SearchSuggestion.start_with(search_term).sort_by(&:value)
     results = start_with.map { |s| { id: s.value, text: s.value } }
 
--- a/app/models/search_suggestion.rb
+++ b/app/models/search_suggestion.rb
@@ -17,12 +17,8 @@ class SearchSuggestion
     end
 
     def start_with(term)
-      cached_suggestions = cached_suggestions_for(term)
-
-      return [] unless cached_suggestions
-
-      cached_suggestions.select { |s| s.value =~ /^#{term}/i }
-                        .first(TradeTariffFrontend.suggestions_count)
+      cached_suggestions_for(term).select { |s| s.value =~ /^#{term}/i }
+                                  .first(TradeTariffFrontend.suggestions_count)
     end
 
   private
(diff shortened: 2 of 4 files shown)


app/controllers/search_controller.rb
```diff
@@ -37,7 +37,7 @@ def search
   end
 
   def suggestions
-    search_term = Regexp.escape(params[:term].to_s)
+    search_term = Regexp.escape(params[:term].to_s.strip)
```
willfish, line 40: Should we add a test for this to avoid regressions?

app/models/search_suggestion.rb
```diff
@@ -17,8 +17,12 @@ def cached_suggestions_for(term)
     end
 
     def start_with(term)
-      cached_suggestions_for(term).select { |s| s.value =~ /^#{term}/i }
-                                  .first(TradeTariffFrontend.suggestions_count)
+      cached_suggestions = cached_suggestions_for(term)
+
+      return [] unless cached_suggestions
```
willfish: This isn't needed AFAIK

If the cached_suggestions_for method returns an empty Array when there are none (it definitely should do), then the select will iterate over nothing and return an empty Array and the first method will return an empty array, too.

spec/models/search_suggestion_spec.rb
```diff
@@ -28,5 +28,13 @@
     it 'returns less then 10 values' do
       expect(described_class.start_with('80').length).to be <= 10
     end
+
+    context 'when cached suggestions is nil' do
+      it 'returns empty array' do
+        allow(described_class).to receive(:cached_suggestions_for).and_return(nil)
```
willfish: We shouldn't mock the subject under test

app/models/search_suggestion.rb
alexdesi: The issue is that  cached_suggestions can return nil,
in that case we get NoMethodError,
for example when you try to search "+" (the example in the screenshot above)

app/controllers/search_controller.rb
alexdesi: yep, I agree

willfish (approved):

app/models/search_suggestion.rb
```diff
@@ -32,12 +32,13 @@ def cached_suggestions_for_letter(first_letter)
...
 
       grouped_results = all.group_by { |r| r.value.to_s.downcase.slice(0, 1) }
+
       grouped_results.transform_keys! { |letter| "#{cache_prefix}#{letter}" }
       grouped_results["#{cache_prefix}loaded"] = true
 
       Rails.cache.write_multi grouped_results, expires_in: 24.hours
 
-      grouped_results["#{cache_prefix}#{first_letter}"]
+      grouped_results["#{cache_prefix}#{first_letter}"].presence || []
```
willfish, line 41: :tada:

PR #541, opened 2021-12-22
HOTT-1205: Replace EU members with EU in measure exclusions

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1205

**Before**

Screenshot from 2021-12-22 15-53-57

**After**

Screenshot from 2021-12-22 15-54-46

### What?

I have added/removed/altered:

- [x] Altered excluded_countries_list method to return the EU when all EU members are present
- [x] Added specs to cover changes
- [x] Added memoization of EU members for better performance
- [x] Moved all GeographicalArea class methods together for easier reading 

### Why?

I am doing this because:

- This is required for a better user experience

### DIFF
--- a/app/models/measure.rb
+++ b/app/models/measure.rb
@@ -41,7 +41,18 @@ class Measure
   end
 
   def excluded_country_list
-    excluded_countries.map(&:description).join(', ').html_safe
+    countries = if exclusions_include_european_union?
+                  # Replace EU members with the EU geographical_area
+                  [GeographicalArea.european_union] + excluded_countries.delete_if(&:eu_member?)
+                else
+                  excluded_countries
+                end
+
+    countries.map(&:description).join(', ').html_safe
+  end
+
+  def exclusions_include_european_union?
+    GeographicalArea.eu_members_ids.all? { |eu_member| eu_member.in?(excluded_country_ids) }
   end
 
   def national?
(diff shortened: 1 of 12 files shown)


app/models/measure.rb
```diff
@@ -41,7 +41,18 @@ def relevant_for_country?(country_code)
   end
 
   def excluded_country_list
-    excluded_countries.map(&:description).join(', ').html_safe
+    countries = if exclusions_include_european_union?
+                  # Replace EU members with the EU geographical_area
+                  [GeographicalArea.european_union] + excluded_countries.delete_if(&:eu_member?)
```
alexdesi, line 46: On a second thought, 
we do not need to actually remove the element from the excluded_countries (`delete_if` or `reject!` ), just filtering it out is enough (`reject`).

spec/spec_helper.rb
```diff
@@ -26,6 +26,10 @@
   )
 end
 
+VCR.use_cassette('geographical_areas#1013') do
+  GeographicalArea.european_union
+end
+
```
alexdesi, line 32: I am not sure if we need this.

alexdesi (approved):
Nice and tidy! 👍 

spec/spec_helper.rb
willfish: I think I can remove it now that I've removed the mocks on `GeographicalArea.find('1013')`

PR #543, opened 2021-12-24
HOTT-1225: Expose feed

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Exposes feed to the world

### Why?

I am doing this because:

- This is needed to get access to feeds in the backend


alexdesi (approved):
✅

PR #558, opened 2022-01-05
HOTT-1225: Optional news item filters for service and target

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Make the parameters to the news items api optional as query params
- [x] Updates stubbing to reflect migration to query params

### Why?

I am doing this because:

- We need to extend these to enable us to not filter based on service


alexdesi (approved):
LGTM

PR #559, opened 2022-01-06
Adds news item feed for users

willfish:
### Jira link

### What?

Surface news items for all services for the update target under /feed/news_items

I have added/removed/altered:

- [x] Added atom feed for news items

### Why?

I am doing this because:

- This is a feature request

### DIFF
--- a/app/models/news_item.rb
+++ b/app/models/news_item.rb
@@ -38,6 +38,13 @@ class NewsItem
       )
     end
 
+    def for_feed
+      collection(
+        collection_path,
+        target: 'updates',
+      )
+    end
+
     def any_updates?
       return false unless TradeTariffFrontend.news_items_enabled?
 
--- a/app/views/feed/news_items/index.atom.builder
+++ b/app/views/feed/news_items/index.atom.builder
@@ -1,0 +1,19 @@
+atom_feed do |feed|
+  feed.title t('feed.news_item_title')
+
+  @news_items.each do |news_item|
+    feed.entry(
+      news_item,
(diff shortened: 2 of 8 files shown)


app/views/feed/news_items/index.atom.builder
```diff
@@ -0,0 +1,19 @@
...
+      entry.content      news_item.content
+      entry.start_date   news_item.start_date.try(:to_s, :rfc822) || ''
+      entry.end_date     news_item.end_date.try(:to_s, :rfc822) || ''
+      entry.uk           news_item.show_on_uk
+      entry.xi           news_item.show_on_xi
+      entry.home_page    home_url
+      entry.updates_page news_items_url
+    end
+  end
+end
```
alexdesi, line 19: is it ok this file not to have .RB extension?

app/models/news_item.rb
```diff
@@ -38,6 +38,13 @@ def updates_page
       )
     end
 
+    def for_feed
+      collection(
+        collection_path,
+        target: 'updates',
+      )
+    end
```
alexdesi, line 46: it might be worth adding a test for `for_feed`

alexdesi (approved):
✅ LGTM

app/models/news_item.rb
willfish: Good idea

app/views/feed/news_items/index.atom.builder
willfish: Yeah, this is right :)

PR #572, opened 2022-01-11
HOTT-1242 Fix duty rendering

jebw:
### Jira link

[HOTT-1242](https://transformuk.atlassian.net/browse/HOTT-1242)

### What?

I have added/removed/altered:

- [x] Drive by fix for misspelt noreferrer
- [x] Added a helper to reformat the duty expression

### Why?

I am doing this because:

- Currently the duty expressions are being inappropriately wrapped - improving the html allow using css to control the where the duty expression will be wrapped

### Before

Screenshot from 2022-01-11 11-07-18
Screenshot from 2022-01-11 11-07-05

### After

Screenshot from 2022-01-11 11-08-31
Screenshot from 2022-01-11 11-08-13

### DIFF
--- a/app/helpers/measures_helper.rb
+++ b/app/helpers/measures_helper.rb
@@ -33,4 +33,26 @@ module MeasuresHelper
       link_to('Check how to export commodity goods (link opens in new tab)', TradeTariffFrontend.check_duties_service_url, target: '_blank', rel: 'noopener')
     end
   end
+
+  def reformat_duty_expression(expression)
+    sanitized = sanitize(expression, tags: %w[abbr], attributes: %w[title])
+    components = []
+
+    while sanitized.length.positive?
+      matched = sanitized.match(%r{ (\+|-|MIN|MAX) })
+
+      if matched
+        components << tag.span(sanitized.slice(0, matched.begin(0)).html_safe)
+        components << matched[1]
+        sanitized = sanitized.slice(matched.end(0), sanitized.length)
+      else
+        components << tag.span(sanitized.html_safe)
+        sanitized = ''
+      end
+    end
+
+    tag.span(class: 'duty-expression') do
+      safe_join components.flatten.compact, ' '
+    end
+  end
 end
--- a/app/helpers/measures_helper.rb
+++ b/app/helpers/measures_helper.rb
@@ -2,7 +2,7 @@ module MeasuresHelper
   def filter_duty_expression(measure)
     return '' if measure.duty_expression.to_s == 'NIHIL'
 
-    reformat_duty_expression(measure.duty_expression.to_s)
+    control_line_wrapping_in_duty_expression(measure.duty_expression.to_s)
   end
 
   def legal_act_regulation_url_link_for(measure)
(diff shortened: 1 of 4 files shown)


app/helpers/measures_helper.rb
```diff
@@ -33,4 +33,28 @@ def check_how_to_export_goods_link(declarable:, country_code:, country_name:, eu
       link_to('Check how to export commodity goods (link opens in new tab)', TradeTariffFrontend.check_duties_service_url, target: '_blank', rel: 'noopener')
     end
   end
+
+  def reformat_duty_expression(expression)
```
willfish: Not sure this conveys the special snowflake purpose of this method. I'd probably add a comment saying it adjusts spans on component expressions that are conjoined to enable more sensible table wrapping

app/helpers/measures_helper.rb
```diff
@@ -33,4 +33,28 @@ def check_how_to_export_goods_link(declarable:, country_code:, country_name:, eu
...
+    # meursing measures are wrapped with strong tags it seems
+    wrapping_tag = expression.start_with?('<strong>') ? :strong : :span
+    sanitized = sanitize(expression, tags: %w[abbr], attributes: %w[title])
+    components = []
+
+    while sanitized.length.positive?
```
willfish: This feels unrubyish to me. Feels like we could handle splitting and iterating and accumulating with each_wiith_object, etc

spec/helpers/measures_helper_spec.rb
```diff
@@ -68,4 +68,48 @@
...
+    subject { reformat_duty_expression expression }
+
+    context 'with simple percentage' do
+      let(:expression) { '<span>20.00</span> % ' }
+
+      it { is_expected.to have_css 'span.duty-expression > span', text: '20.00 %' }
+    end
+
+    context 'with abbreviation' do
+      let :expression do
```
willfish, line 82: We do a mixture of DSL let and non-DSL lets without parentheses, here. Not blocking but I reckon we should just pick one

willfish (changes requested):
Seems to do what we want but I do think we should add a comment explaining to future developers what this is for

spec/helpers/measures_helper_spec.rb
jebw: Not sure I follow - pattern I'm using is brackets for methods with inline blocks, and without brackets if using multiline `do`/`end` - same as for the `context` blocks?

willfish: I've not seen foregoing parentheses on let regardless of multi/single line style in the wild/other projects. Better specs do not mention that style, etiher. Is fine just jarring if you're not used to it. I always just use parentheses so I guess we have two different styles there

app/helpers/measures_helper.rb
willfish: Spoken about this offline. The split implementation doesn't yield what was split on so we'd struggle to recreate this behaviour

willfish (approved):
Amaze :soccer: :soccer:

PR #574, opened 2022-01-11
Dropped rails-erb-loader and sentry/browser

jebw:
### Jira link

HOTT-???

### What?

I have added/removed/altered:

- [x] Removed `rails-erb-loader` which allows usage of erb in webpacker'd assets
- [x] Removed sentry/browser that was relying upon it - integration was already broken

### Why?

I am doing this because:

- its an added complexity that does not receive updates and its only use is to configure sentry, and that doesn't work correctly now we are using docker
- if we require the `sentry/browser` integration it can be brought back with a configuration that actually works


package.json
```diff
@@ -26,7 +25,6 @@
     "jquery-roving-tabindex": "^1.0.1",
     "jquery-tabs": "^0.5.3",
     "mark.js": "^8.11.1",
-    "rails-erb-loader": "^5.5.2",
```
willfish, line 29: Do we want to also remove https://github.com/trade-tariff/trade-tariff-frontend/blob/main/config/webpack/loaders/erb.js

willfish (approved):

package.json
jebw: good point - I'll amend that

PR #591, opened 2022-01-24
HOTT-1256 Only show CHEG link to UK users

jebw:
### Jira link

[HOTT-1256](https://transformuk.atlassian.net/browse/HOTT-1256)

### What?

I have added/removed/altered:

- [x] Conditionalised display of CHEG link - if the Cloudfront header is either missing or its present and set to the UK then we'll display the link box, otherwise we exclude it

### Why?

I am doing this because:

- We only want to present that link to UK users


willfish:
We're going to pick this up when we have the infrastructure ready

PR #592, opened 2022-01-24
HOTT-1255 Show certificate description instead of requirement

jebw:
### Jira link

[HOTT-1255](https://transformuk.atlassian.net/browse/HOTT-1255)

### What?

I have added/removed/altered:

- [x] Changed to show the `certiicate_description` field if set in place of the requirement field - for Default, Document and Quantity Measure condition codes

### Why?

I am doing this because:

- It avoids showing the duplication in the requirements field

### DIFF
--- a/app/views/measure_conditions/_measure_condition_code_default.html.erb
+++ b/app/views/measure_conditions/_measure_condition_code_default.html.erb
@@ -13,7 +13,7 @@
     <tr class="govuk-table__row">
       <td class="govuk-table__cell numerical"><%= condition.document_code %></td>
       <td class="govuk-table__cell">
-        <%= condition.requirement.to_s.html_safe %>
+        <%= condition.certificate_description.presence || condition.requirement.to_s.html_safe %>
       </td>
       <td class="govuk-table__cell"><%= condition.action %></td>
       <td class="govuk-table__cell numerical">
--- a/app/views/measure_conditions/_measure_condition_code_document.html.erb
+++ b/app/views/measure_conditions/_measure_condition_code_document.html.erb
@@ -13,7 +13,7 @@
       <% if condition.document_code.present? %>
         <td class="govuk-table__cell numerical"><%= condition.document_code %></td>
         <td class="govuk-table__cell">
-          <%= condition.requirement %>
+          <%= condition.certificate_description.presence || condition.requirement %>
         </td>
       <% else %>
         <td class="govuk-table__cell">&nbsp;</td>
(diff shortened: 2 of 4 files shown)


app/views/measure_conditions/_measure_condition_code_document.html.erb
```diff
@@ -13,7 +13,7 @@
       <% if condition.document_code.present? %>
         <td class="govuk-table__cell numerical"><%= condition.document_code %></td>
         <td class="govuk-table__cell">
-          <%= condition.requirement %>
+          <%= condition.certificate_description.presence || condition.requirement %>
```
alexdesi, line 16: I guess it need `.to_s.html_safe` as in `_measure_condition_code_default.html.erb `

app/views/measure_conditions/_measure_condition_code_default.html.erb
```diff
@@ -13,7 +13,7 @@
     <tr class="govuk-table__row">
       <td class="govuk-table__cell numerical"><%= condition.document_code %></td>
       <td class="govuk-table__cell">
-        <%= condition.requirement.to_s.html_safe %>
+        <%= condition.certificate_description.presence || condition.requirement.to_s.html_safe %>
```
alexdesi, line 16: does the description of the ticket
"Show certificate description instead of requirement"
imply that `condition.requirement` has priority on the `certificate_description`?

app/views/measure_conditions/_measure_condition_code_quantity.html.erb
```diff
@@ -21,7 +21,7 @@
       <% else %>
         <td class="govuk-table__cell numerical"><%= condition.document_code %></td>
         <td class="govuk-table__cell">
-          <%= condition.requirement.to_s.html_safe %>
+          <%= condition.certificate_description.presence || condition.requirement.to_s.html_safe %>
```
alexdesi, line 24: a method in `condition` might be handy, but I am not sure about a potential name for it.

willfish (approved):

app/views/measure_conditions/_measure_condition_code_document.html.erb
jebw: the html content in `requirement` appears to be for non-document measure conditions - and in that scenario certificate_description will be nil and fall back to `requirement` - `requirement` is already marked as html_safe in the data model.

app/views/measure_conditions/_measure_condition_code_default.html.erb
jebw: no,  quite the opposite ?

app/views/measure_conditions/_measure_condition_code_quantity.html.erb
jebw: This feels quite like a 'view' level concern but in general terms this will all likely get replaced as part of the rework Matt raised last week so I'm keen to stick with the existing pattern for now which at least makes it easy to tailor to different types.

alexdesi (approved):

PR #597, opened 2022-01-26
HOTT-1250: Expose subheadings

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1250

### What?

I have added/removed/altered:

- [x] Added subheadings to list of allowed backend apis

### Why?

I am doing this because:

- This is required so we can expose subheadings (see https://github.com/trade-tariff/trade-tariff-backend/pull/377)


alexdesi (approved):
✅

PR #600, opened 2022-01-28
HOTT-1226: Removes obfuscating fetching of the page heading

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1226

### What?

I have added/removed/altered:

- [x] Removed obfuscating code around fetching a page header for goods nomenclature

### Why?

I am doing this because:

- We know from where we're called (e.g. show me a section/chapter/heading/subheading/commodity) what the page header should be
- All of these entities implement a page_heading method (rightly or wrongly)
- This is obfuscating and defensive when we know what the resource is

### DIFF
--- a/app/helpers/service_helper.rb
+++ b/app/helpers/service_helper.rb
@@ -3,12 +3,6 @@ module ServiceHelper
     t('title.default', service_name: service_name, service_description: service_description)
   end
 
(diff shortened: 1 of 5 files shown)


alexdesi (approved):
it make sense to me!

PR #601, opened 2022-01-28
HOTT-1226: Add ability to view subheading

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1226

image

### What?

I have added/removed/altered:

- [x] Added controllers, models, routes and specs for viewing a subheading

### Why?

I am doing this because:

- This is required so that we can populate search references and have them redirect to a subheading


app/controllers/subheadings_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def xi_subheading
+    @xi_subheading ||= TradeTariffFrontend::ServiceChooser.with_source(:xi) do
+      Subheading.find(params[:id], query_params)
+    end
+  end
+
+  def subheading
+    @subheading ||= TradeTariffFrontend::ServiceChooser.uk? ? uk_subheading : xi_subheading
+  end
```
alexdesi: Should we move this method before `uk_subheading` and `xi_subheading` ?
Just for readability, nothing wrong with it.

app/models/subheading.rb
```diff
@@ -0,0 +1,54 @@
...
+                :declarable
+
+  def page_heading
+    "Subheading #{code} - #{self}"
+  end
+
+  def code
+    case goods_nomenclature_item_id
+    when /\d+0000/ then harmonised_system_subheading_code
+    when /\d+00/ then combined_nomenclature_subheading_code
```
alexdesi: are the 0s on the last position?
maybe we should add the `$` char in the Regexps, at the end
so if the `when` are switched:

```
 when /\d+00/ then combined_nomenclature_subheading_code
when /\d+0000/ then harmonised_system_subheading_code
```
it still works.

app/views/shared/_footnote.html.erb
```diff
@@ -0,0 +1,21 @@
...
+      <tr class="govuk-table__row">
+        <th scope="col" class="govuk-table__header">Code</th>
+        <th scope="col" class="govuk-table__header">Description</th>
+      </tr>
+    </thead>
+    <tbody class="govuk-table__body">
+      <% footnotes.each do |footnote| %>
+        <% next unless footnote&.code.present? %>
+        <%= render footnote %>
+      <% end %>
```
alexdesi: small suggestion:
somewhere:
```
def footnotes_with_valid_code
  footnotes.reject{|f| code.blank?}
end
```

here in the view:

```
<% footnotes_with_valid_code.each do |footnote| %>
   <%= render footnote %>
<% end %>
```

app/views/shared/_top_breadcrumbs.html.erb
```diff
@@ -20,7 +20,14 @@
 
         <% if @heading && !@heading.declarable?%>
           <li class="govuk-breadcrumbs__list-item">
-            <%= breadcrumb_link_or_text(@heading, @commodity, "Heading #{@heading.short_code}") %>
+            <% child = @commodity || @subheading %>
```
alexdesi: to remove the assignment, what about just using the value (I think it is not worth adding a helper for that)

```ruby
breadcrumb_link_or_text(@heading,
  @commodity || @subheading,
  "Heading #{@heading.short_code}
)
```

alexdesi (commented):
few minor comments

app/controllers/subheadings_controller.rb
willfish: Yeah, I agree and it follows the newspaper metaphor. Good shout

app/models/subheading.rb
willfish: We can anchor from the start/end of the string. I didn't add since we always know what this is going to be and it felt defensive.

willfish: I will add it

app/views/shared/_footnote.html.erb
willfish: This is me moving code about, really

willfish: I can change that in a separate pr

app/views/shared/_top_breadcrumbs.html.erb
willfish: What's wrong with the assignment?

app/views/shared/_footnote.html.erb
willfish: I think it's historical/unused I can probably safely remove the check

spec/requests/subheadings_controller_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  before do
+    VCR.use_cassette('subheadings#show') do
+      visit subheading_path('0101210000-10')
+    end
+  end
+
+  it { expect(page).to have_link 'Search', href: '/find_commodity' }
+  it { expect(page).to have_link 'Browse', href: '/browse' }
+
+  it { expect(page).to have_content "There are 3 commodities in this category. Choose the commodity code that best matches your goods to see more information. If your item is not listed by name, it may be shown under what it's used for, what it's made from or 'Other'." }
```
alexdesi, line 13: this is loooooooooong :)

alexdesi (approved):
lgtm ✅

spec/requests/subheadings_controller_spec.rb
willfish: I could probably trim it :dagger:

PR #630, opened 2022-02-10
HOTT-1295: Bump rails to 7.0

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1295

### What?

I have added/removed/altered:

- [x] Bumps rails and surrounding dependencies

### Why?

I am doing this because:

- We're globally upgrading rails 7 as it becomes possible via our dependencies
- This enables us to more easily upgrade to ruby 3.1.0

### Not included

- I'll tweak the config defaults separately


Gemfile
```diff
@@ -25,7 +25,7 @@ gem 'responders'
 gem 'webpacker'
 
 # gov UK
-gem 'govspeak'
+gem 'govspeak', github: 'willfish/govspeak', branch: 'HOTT-1295-bump-govspeak-rails-compatibility'
```
willfish: I have an open PR that should get merged soon hopefully

willfish: I've validated that we're still able to render notes

jebw (approved):
LGTM

PR #634, opened 2022-02-11
HOTT-1280: Support overrides on measure type descriptions

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1280

### What?

I have added/removed/altered:

- [x] Added mechanism for translating measure type descriptions
- [x] Added translations for Jersey and Guernsey MFN measure types
- [x] Added specs to validate this new mechanism

### Why?

I am doing this because:

- This is required so we can tweak anticipated measures for Jersey and Guernsey


jebw (approved):
LGTM

PR #638, opened 2022-02-14
HOTT-1295: Bumps ruby to 3.1.0

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1295

### What?

I have added/removed/altered:

- [x] Bump ruby to 3.1.0

### Why?

I am doing this because:

- We get performance, debugging and error feedback improvements in this release


jebw (commented):
imo this should be held back and deployed separately from the rails 7 changes

PR #639, opened 2022-02-14
HOTT-1295: Move to Rails 7 configuration

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Move to rails 7 configuration defaults

### Why?

I am doing this because:

- This is part of a 3 step upgrade to rails

### DIFF
--- a/app/models/tariff_date.rb
+++ b/app/models/tariff_date.rb
@@ -31,4 +31,6 @@ class TariffDate < SimpleDelegator
   def to_s(format = :date)
     to_formatted_s(format)
   end
+
+  alias_method :to_fs, :to_s
 end
--- a/app/views/pages/tools.html.erb
+++ b/app/views/pages/tools.html.erb
@@ -38,7 +38,7 @@
       </li>
       <li>
         <p>
-          <%= link_to t('breadcrumb.footnotes'), footnote_search_path %>
+          <%= link_to t('breadcrumb.footnote_search'), footnote_search_path %>
           <br>Search the tariff for footnotes
         </p>
       </li>
--- a/config/application.rb
+++ b/config/application.rb
@@ -1,11 +1,11 @@
 require_relative 'boot'
 
 require 'rails'
-require 'active_support'
 # Pick the frameworks you want:
 require 'active_model/railtie'
 # require "active_job/railtie"
 require 'active_record/attribute_assignment'
+# require "active_record/railtie"
 # require "active_storage/engine"
 require 'action_controller/railtie'
 require 'action_mailer/railtie'
@@ -13,7 +13,6 @@ require 'action_mailer/railtie'
 # require "action_text/engine"
 require 'action_view/railtie'
 # require "action_cable/engine"
-# require "sprockets/railtie"
 
 # Require the gems listed in Gemfile, including any gems
 # you've limited to :test, :development, or :production.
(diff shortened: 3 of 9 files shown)


config/application.rb
```diff
@@ -24,59 +23,22 @@
...
-    # Set Time.zone default to the specified zone and make Active Record auto-convert to this zone.
-    # Run "rake -D time" for a list of tasks for finding time zone names. Default is UTC.
-    # config.time_zone = 'Central Time (US & Canada)'
-
-    # The default locale is :en and all translations from config/locales/*.rb,yml are auto loaded.
-    # config.i18n.load_path += Dir[Rails.root.join('my', 'locales', '*.{rb,yml}').to_s]
-    # config.i18n.default_locale = :de
-
-    # Configure the default encoding used in templates for Ruby 1.9.
-    config.encoding = 'utf-8'
```
willfish, line 60: https://edgeguides.rubyonrails.org/configuring.html#config-encoding

config/environments/production.rb
```diff
@@ -16,49 +16,55 @@
...
-  config.read_encrypted_secrets = true
+  # Ensures that a master key has been made available in either ENV["RAILS_MASTER_KEY"]
+  # or in config/master.key. This key is used to decrypt credentials (and other encrypted files).
+  # config.require_master_key = true
 
-  # Disable Rails's static asset server (Apache or nginx will already do this)
-  config.public_file_server.enabled = true
+  # Disable serving static files from the `/public` folder by default since
+  # Apache or NGINX already handles this.
+  config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?
```
willfish, line 25: This is already set in the environment of the frontend to true

app/views/pages/tools.html.erb
```diff
@@ -38,7 +38,7 @@
       </li>
       <li>
         <p>
-          <%= link_to t('breadcrumb.footnotes'), footnote_search_path %>
+          <%= link_to t('breadcrumb.footnote_search'), footnote_search_path %>
```
willfish, line 41: This localisation was broken and only noticed because I changed the test localisation config to raise on misses

config/application.rb
```diff
@@ -1,19 +1,18 @@
 require_relative 'boot'
 
 require 'rails'
-require 'active_support'
 # Pick the frameworks you want:
 require 'active_model/railtie'
 # require "active_job/railtie"
 require 'active_record/attribute_assignment'
+# require "active_record/railtie"
```
alexdesi, line 8: I guess we don't need this comment

app/models/tariff_date.rb
```diff
@@ -31,4 +31,6 @@ def to_param
   def to_s(format = :date)
     to_formatted_s(format)
   end
+
+  alias_method :to_fs, :to_s
```
willfish, line 35: This makes us consistent with https://github.com/rails/rails/pull/44354/files

config/environments/production.rb
```diff
@@ -16,45 +16,56 @@
...
+  config.log_tags = [:request_id]
 
-  config.webpacker.check_yarn_integrity = false
+  # Use a different cache store in production.
+  config.cache_store = :redis_store,
+                       PaasConfig.redis.merge({
+                         expires_in: 1.day,
+                         namespace: ENV['GOVUK_APP_DOMAIN'],
+                         pool_size: Integer(ENV['MAX_THREADS'] || 5),
+                       })
```
alexdesi, line 50: yes, this sounds  correct (what we got now)

app/models/tariff_date.rb
alexdesi, line 35: We discussed this. It's ok to keep it for now.

alexdesi (approved):
I looks good to me ✅

config/application.rb
willfish: I left it in because it comes from running rails app:update and its just stuff to alter everytime we do an upgrade

PR #643, opened 2022-02-15
HOTT-1333: Tidy up TariffDate

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Refactored tariff date to have a simpler api/consolidate responsibilities
- [x] Refactored specs to up coverage
- [x] Removed dead methods

### Why?

I am doing this because:

- This code is confusing and is critical to the time machine and search functionality

### DIFF
--- a/app/models/search.rb
+++ b/app/models/search.rb
@@ -51,8 +51,7 @@ class Search
   end
 
   def date
-    # TODO: We should remove support for as_of as a query param when making html calls
-    @date ||= as_of.present? ? TariffDate.build_legacy(as_of) : TariffDate.build(attributes)
+    @date ||= TariffDate.build(attributes)
   end
 
   def filtered_by_date?
--- a/app/models/search.rb
+++ b/app/models/search.rb
@@ -51,24 +51,11 @@ class Search
   end
 
   def date
-    @date ||= if as_of.present?
-                TariffDate.new(as_of)
-              else
-                TariffDate.parse(attributes.slice(*TariffDate::DATE_KEYS))
-              end
-  end
-
-  def currency_name
-    case currency
-    when 'GBP'
-      'Pound sterling'
-    else
-      'Euro'
-    end
+    @date ||= TariffDate.build(as_of, attributes)
   end
 
   def filtered_by_date?
-    date.date != TariffUpdate.latest_applied_import_date
+    date != TariffUpdate.latest_applied_import_date
   end
 
   def filtered_by_country?
--- a/app/models/search.rb
+++ b/app/models/search.rb
@@ -51,7 +51,8 @@ class Search
   end
 
   def date
-    @date ||= TariffDate.build(as_of, attributes)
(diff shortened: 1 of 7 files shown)


app/models/search.rb
```diff
@@ -51,24 +51,11 @@ def geographical_area
...
 
   def date
-    @date ||= if as_of.present?
-                TariffDate.new(as_of)
-              else
-                TariffDate.parse(attributes.slice(*TariffDate::DATE_KEYS))
-              end
-  end
-
-  def currency_name
```
willfish, line 61: This is completely unused code

app/models/tariff_date.rb
```diff
@@ -1,59 +1,50 @@
...
   def to_param
-    date.to_formatted_s(:full)
+    to_formatted_s(:full)
   end
 
   def to_s(format = :date)
-    date.to_formatted_s(format)
-  end
-
-  def to_taric_date
```
willfish, line 51: Completely unused

app/models/tariff_date.rb
```diff
@@ -1,59 +1,50 @@
...
   DATE_KEYS = %w[year month day].freeze
 
-  delegate :day, :month, :year, :to_formatted_s, :strftime, :today?, :>=, :<=, :==, to: :date
+  class << self
+    def build(as_of, attributes)
+      date = if as_of.present?
+               handle_as_of(as_of)
+             else
+               handle_date_params(attributes)
+             end
```
alexdesi: It looks like these are two different builders mixed into a unique `build`.

What is unclear to me is the logic, why if we pass in `as_of` them we must rescue with TariffUpdate.latest_applied_import_date,
and 
with attributes we rescue with`Time.zone.today`.

`as_of` and `attributes` are both dates, it sounds to me we'd need two builders:

`build_by_<what_as_of_represent>`
and 
`build_by_<what_the_attributes_represent>`

willfish: So my motivation was to move the functionality for differently constructed TariffDates behind a single build method. This simplified the api significantly. I can separate these quite easily with two separate build methods but I'd need to validate how the search model drives the different builders depending on the availability of the as_of. This implementation is actually simpler to test

willfish: I don't know why BZ used `TariffUpdate.latest_applied_import_date` I'm just preserving existing behaviour

spec/models/search_spec.rb
```diff
@@ -1,8 +1,29 @@
...
+    subject(:search) do
+      described_class.new(
+        'q' => 'foo',
+        'day' => '01',
+        'month' => '02',
+        'year' => '2021',
+      )
+    end
+
+    it 'calls the legacy TariffDate builder' do
```
willfish:
```suggestion
    it 'calls the TariffDate builder' do
```

alexdesi (approved):
🧹 🧹 🧹

PR #648, opened 2022-02-15
HOTT-1333: Tidy up tariff update model

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Removed unused crud from the tariff update model, adds missing specs, removes unused specs and generally tidy up tariff update behaviour

### DIFF
--- a/app/models/tariff_update.rb
+++ b/app/models/tariff_update.rb
@@ -1,56 +1,15 @@
-require 'api_entity'
-
 class TariffUpdate
   include ApiEntity
 
   collection_path '/updates/latest'
 
-  attr_accessor :update_type, :state, :created_at, :updated_at, :filename
-
-  def update_type
-    case attributes['update_type']
-    when /Taric/ then 'TARIC'
-    when /Chief/ then 'CHIEF'
-    end
-  end
-
-  def updated_at
-    date_attribute('updated_at')
-  end
-
   def applied_at
-    date_attribute('applied_at')
+    Date.parse(attributes['applied_at'])
   end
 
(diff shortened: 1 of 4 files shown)


app/models/tariff_update.rb
```diff
@@ -5,52 +5,13 @@ class TariffUpdate
...
-    end
-  end
-
-  def updated_at
-    date_attribute('updated_at')
-  end
-
   def applied_at
-    date_attribute('applied_at')
+    Date.parse(attributes['applied_at'].to_s)
```
alexdesi: maybe  `to_s` is redundant.

app/models/tariff_update.rb
```diff
@@ -5,52 +5,13 @@ class TariffUpdate
...
-        # Cache for 1 hour
-        TradeTariffFrontend::ServiceChooser.cache_with_service_choice(
-          'tariff_last_updated', expires_in: 1.hour
-        ) do
-          func.call
-        end
-      end
+  def self.latest_applied_import_date
+    TradeTariffFrontend::ServiceChooser.cache_with_service_choice('tariff_last_updated', expires_in: 23.hours) do
+      all.first.try(:applied_at) || Time.zone.today
```
alexdesi, line 12: `first` should work as `all.first`, and it should be more efficient

alexdesi (approved):
👍 + two minor comments

app/models/tariff_update.rb
willfish: Not activerecord, sadly

PR #650, opened 2022-02-15
More as_of search cleanup

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Updated specs to use a search factory
- [x] Removed unused as_of attribute from the Search model
- [x] Removed old references to as_of

### Why?

I am doing this because:

- This is part of the TariffDate cleanup

### DIFF
--- a/app/controllers/goods_nomenclatures_controller.rb
+++ b/app/controllers/goods_nomenclatures_controller.rb
@@ -10,9 +10,9 @@ class GoodsNomenclaturesController < ApplicationController
   def find_relevant_goods_code_or_fallback
     @search = Search.new(
       q: goods_code_id,
-      'day' => @tariff_last_updated.try(:day),
-      'month' => @tariff_last_updated.try(:month),
-      'year' => @tariff_last_updated.try(:year),
+      day: @tariff_last_updated.try(:day),
+      month: @tariff_last_updated.try(:month),
+      year: @tariff_last_updated.try(:year),
     )
     results = @search.perform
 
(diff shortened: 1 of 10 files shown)


spec/views/find_commodities/show.html.erb_spec.rb
```diff
@@ -5,9 +5,10 @@
 
   before { assign :search, search }
 
-  let(:now) { Time.zone.today }
+  let(:search_date) { Time.zone.today }
```
alexdesi, line 8: 👍

app/controllers/goods_nomenclatures_controller.rb
```diff
@@ -8,7 +8,12 @@ def goods_code_id
   end
 
   def find_relevant_goods_code_or_fallback
-    @search = Search.new(q: goods_code_id, as_of: @tariff_last_updated)
+    @search = Search.new(
+      q: goods_code_id,
+      'day' => @tariff_last_updated.try(:day),
+      'month' => @tariff_last_updated.try(:month),
+      'year' => @tariff_last_updated.try(:year),
```
alexdesi: doesn't work with symbols?

```
day: @tariff_last_updated.try(:day),
month: ...,
year: ...
```

alexdesi (approved):
it looks good ✅

app/controllers/goods_nomenclatures_controller.rb
willfish: We could move the api entity to HashWithIndifferentAccess. The answer is no, though

alexdesi (approved):

PR #661, opened 2022-02-21
HOTT-1198: Adds news feed section

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1198

### What?

I have added/removed/altered:

- [x] Added information to news to explain the news feed

### Why?

I am doing this because:

- This is required so users can subscribe to news updates


jebw (approved):
LGTM

PR #662, opened 2022-02-21
HOTT-1298: Expose quota order numbers api

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1298

### What?

See related: https://github.com/trade-tariff/trade-tariff-backend/pull/425

I have added/removed/altered:

- [x] Added quota_order_numbers to api whitelist

### Why?

I am doing this because:

- This is required to enable the quotas api


jebw (approved):
LGTM

PR #668, opened 2022-02-23
This PR reduces the noise (10k of lines) in the Ruby console

alexdesi:
### What?
This PR reduces the noise (10k of lines) in the Ruby console, that occurred when method missing for any type of GoodsNomenclature.

### Why?
To make the world less digitally noisy, and improve Alessandro's mental health.

### Jira link
HOTT-1362

Screenshot:
Screenshot 2022-02-23 at 11 43 52

### DIFF
--- a/app/models/goods_nomenclature.rb
+++ b/app/models/goods_nomenclature.rb
@@ -69,4 +69,15 @@ class GoodsNomenclature
   def self.is_heading_id?(goods_nomenclature_item_id)
     goods_nomenclature_item_id.ends_with?('000000') && goods_nomenclature_item_id.slice(2, 2) != '00'
   end
+
+  def inspect
+    {
+      object_id: object_id,
+      id:,
+      type:,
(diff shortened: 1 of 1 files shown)


app/models/goods_nomenclature.rb
```diff
@@ -69,4 +69,15 @@ def rules_of_origin_rules(...)
...
     goods_nomenclature_item_id.ends_with?('000000') && goods_nomenclature_item_id.slice(2, 2) != '00'
   end
+
+  def inspect
+    {
+      object_id: object_id,
+      id:,
+      type:,
+      goods_nomenclature_item_id:,
+      formatted_description:,
```
alexdesi, line 79: we don't need this, I'll remove it!

willfish: My view is we should override ApiEntity and use accumulated attr_accessors

alexdesi:
This has been implemented in a different PR.

PR #669, opened 2022-02-23
Handle large console output

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Alter the api entity module so that it accumulates relationships and casted by and excludes them from inspect output
- [x] Add specs for the relationships method

### Why?

I am doing this because:

- This reduces console heavyness


alexdesi (approved):
✅

PR #675, opened 2022-02-28
HOTT-1371 Allow rendering XI measures without UK commodity

jebw:
### Jira link

[HOTT-1371](https://transformuk.atlassian.net/browse/HOTT-1371)

### What?

I have added/removed/altered:

- [x] Updated measures views to still render when no UK commodity present
- [x] Updated commodities controller to handle commodities present in the XI service but not the UK service

### Why?

I am doing this because:

- Currently we cannot render XI commodities if there is not a UK equivalent

### Have you? (optional)

### DIFF
--- a/spec/views/measures/_measures.html.erb_spec.rb
+++ b/spec/views/measures/_measures.html.erb_spec.rb
@@ -104,4 +104,28 @@ RSpec.describe 'measures/_measures', type: :view, vcr: {
       it { is_expected.to have_css '#rules-of-origin h2', text: 'rules of origin for trading' }
     end
   end
+
+  context 'with xi service with no equivalent uk commodity' do
+    include_context 'with XI service'
+
+    let :render_page do
+      render 'measures/measures',
+             declarable: presented_commodity,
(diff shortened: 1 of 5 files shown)


spec/views/measures/_measures.html.erb_spec.rb
```diff
@@ -104,4 +104,28 @@
       it { is_expected.to have_css '#rules-of-origin h2', text: 'rules of origin for trading' }
     end
   end
+
+  context 'with xi service with no equivalent uk commodity' do
```
alexdesi, line 108: did you mean:
'with XI service having no equivalent UK commodity'
?

alexdesi (approved):
It looks a bit defensive programming,
with many lots's of `&.`, but it sounds good overall.
✅

spec/views/measures/_measures.html.erb_spec.rb
jebw: Not sure it really means much difference in my reading of it tbh?

PR #684, opened 2022-03-03
Remove the measures that are Supplementary unit from the Ex…

alexdesi:
### What?
Remove the measures that are Supplementary unit from the Exporting from the UK table.

### Jira link
[hott-1353](https://transformuk.atlassian.net/browse/HOTT-1353)

### DIFF
--- a/app/models/measure_collection.rb
+++ b/app/models/measure_collection.rb
@@ -10,6 +10,10 @@ class MeasureCollection
     @measures = measures.clone
   end
 
+  def exclude_supplementary_unit
+    self.class.new(measures.reject { |m| m.measure_type.description == 'Supplementary unit' })
+  end
+
   def for_country(country)
     self.class.new(measures.select { |m| m.relevant_for_country?(country) })
   end
--- a/app/models/measure_collection.rb
+++ b/app/models/measure_collection.rb
@@ -10,8 +10,8 @@ class MeasureCollection
     @measures = measures.clone
   end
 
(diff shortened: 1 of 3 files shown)


app/models/measure_collection.rb
```diff
@@ -10,6 +10,10 @@ def initialize(measures)
     @measures = measures.clone
   end
 
+  def exclude_supplementary_unit
+    self.class.new(measures.reject { |m| m.measure_type.description == 'Supplementary unit' })
```
willfish: We 100% shouldn't use a changing description for this

app/models/measure_collection.rb
```diff
@@ -10,6 +10,10 @@ def initialize(measures)
     @measures = measures.clone
   end
 
+  def exclude_supplementary_unit
```
willfish:
```suggestion
  def without_supplemenatry_unit
    self.class.new(measures.reject(&:supplementary_unit?))
  end
```

willfish (changes requested):
We 100% shouldn't ship this using a non-uniquely identifying description when we have the types available

app/models/measure_collection.rb
alexdesi: Yeh, I was not sure about it, but I could not find any info in the table `measure_types` and `measures` where the relevant info is stored (and not the description).

willfish (approved):

PR #689, opened 2022-03-07
HOTT-1387: Handle excluded and ungrouped measure types

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1387

### What?

image

I have added/removed/altered:

- [x] Added filter mechanism to ApiEntity to handle frontend-excluded measure types - this is separate/in addition to the backend exclusion mechanism.
- [x] Added business methods for determining whether an unclassified measure should be included in the customs duties or import controls groups of the import measures tab
- [x] Renamed the declarable types file to reflect what it actually contains (grouped measure types)
- [x] Added specs for all of the above

### Why?

I am doing this because:

- This is required so that we can exclude certain measure types from the frontend and so that we can handle the situation where new measure types are created by either the EU/UK that are currently unknowable


jebw (approved):
LGTM

PR #692, opened 2022-03-08
Tidy up spec factories

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Removes stringify keys from the project

### Why?

I am doing this because:

- Api attributes use HashWithIndifferentAccess so no longer require factories to use strings

### DIFF
--- a/spec/controllers/chapters/changes_controller_spec.rb
+++ b/spec/controllers/chapters/changes_controller_spec.rb
@@ -1,7 +1,7 @@
 require 'spec_helper'
 
 RSpec.describe Chapters::ChangesController, 'GET to #index', type: :controller do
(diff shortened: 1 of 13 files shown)


spec/controllers/chapters/changes_controller_spec.rb
```diff
@@ -1,7 +1,7 @@
 require 'spec_helper'
 
 RSpec.describe Chapters::ChangesController, 'GET to #index', type: :controller do
-  let!(:chapter) { Chapter.new(attributes_for(:chapter, goods_nomenclature_item_id: '0100000000').stringify_keys) }
```
willfish, line 4: I mean, this should just use the factory - another day, another commit, though

jebw (approved):
LGTM

PR #693, opened 2022-03-08
Removes timecop

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Removed timecop

### Why?

I am doing this because:

- Its not actually used and should never be necessary in the frontend (or anywhere?)


jebw (approved):
LGTM

PR #694, opened 2022-03-09
HOTT-1375: Order customs duties section by sub category

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1375

### What?

I have added/removed/altered:

- [x] Added ordering of customs duties
- [x] Removed now-redundant groups of measure types in the measure type file

### Why?

I am doing this because:

- Third country duties were less obvious to users so we wanted to hoist these to
  the top of the customs duties section


app/views/commodities/_commodity.html.erb
```diff
@@ -23,7 +23,7 @@
           <% end %>
         </div>
         <div class="duty" aria-describedby="commodity-duty-title">
-          <% tcd = commodity.overview_measures.third_country_duty %>
+          <% tcd = commodity.overview_measures.third_country_duties %>
```
jebw, line 26: I think this shifts from using a MeasureCollection to using a Array - not sure that makes any difference though?

jebw (approved):
Looks good

app/views/commodities/_commodity.html.erb
willfish: I don't think it actually matters for the overview measures. I believe they're also not behind the MeasurePresenter :shrug:

willfish: Probably better for consistency, though

PR #697, opened 2022-03-10
HOTT-1361 Allow jsonapi objects without attributes

jebw:
### Jira link

[HOTT-1361](https://transformuk.atlassian.net/browse/HOTT-1361)

### What?

I have added/removed/altered:

- [x] Changed the JSON-API parser to not require an attributes key on a resource

### Why?

I am doing this because:

- `attributes` is not a required key according to the spec (see https://jsonapi.org/format/#document-resource-objects)
- a resource which only exists to define relationships (ie the new Permutations resource) will not have any attributes

- Makes changes to our api parser, which is used by pretty much every screen

### DIFF
--- a/lib/tariff_jsonapi_parser.rb
+++ b/lib/tariff_jsonapi_parser.rb
@@ -31,7 +31,7 @@ class TariffJsonapiParser
   def parse_resource(resource)
     result = {}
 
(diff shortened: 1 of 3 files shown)


alexdesi (approved):
lgtm 🟢

PR #702, opened 2022-03-14
HOTT-1401: Handle commodity mismatch on XI commodities

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1401

### What?

I have added/removed/altered:

- [x] Added handling for commodities on XI that have no UK commodity

### Why?

I am doing this because:

- We updated the behaviour around 404s so that we show only the EU measures when there is no corresponding UK commodity and missed this on a recent refactor
- See: https://github.com/trade-tariff/trade-tariff-frontend/pull/675

### DIFF
--- a/app/controllers/commodities_controller.rb
+++ b/app/controllers/commodities_controller.rb
@@ -29,7 +29,13 @@ class CommoditiesController < GoodsNomenclaturesController
   end
 
   def uk_declarable
-    @uk_declarable ||= heading? ? uk_heading : uk_commodity
+    @uk_declarable ||= begin
+      heading? ? uk_heading : uk_commodity
+    rescue Faraday::ResourceNotFound
+      # When loading EU commodities we pull some measures from the UK commodity. If the UK commodity is not present
+      # in the UK database, then we should show only the applicable EU measures and not render the 404 page.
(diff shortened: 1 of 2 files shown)


app/controllers/commodities_controller.rb
```diff
@@ -29,7 +29,17 @@ def declarable
...
+    @uk_declarable ||= begin
+      heading? ? uk_heading : uk_commodity
+    rescue Faraday::ResourceNotFound
+      if TradeTariffFrontend::ServiceChooser.xi?
+        # When loading EU commodities we pull some measures from the UK commodity. If the UK commodity is not present
+        # in the UK database, then we should show only the applicable EU measures.
+        nil
+      else
+        raise
+      end
```
alexdesi: this looks the same as:
`raise if TradeTariffFrontend::ServiceChooser.uk?`

alexdesi (approved):
it looks good ✅,
maybe it can be simplified a bit (see comment)

app/controllers/commodities_controller.rb
willfish: Derp, yeah

PR #713, opened 2022-03-15
Refactor measure collection

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Remove unused Enumerable implementation from MeasureCollection wrapper
- [x] Use SimpleDelegator to simplify MeasureCollection wrapping Measures
- [x] Move away from class_eval in ApiEntity
- [x] Add support for presenter option in has_many method of ApiEntity to clean up measure collection implementation 
- [x] Removes testing of Array functionality

### Why?

I am doing this because:

- The ApiEntity module is a bit hard to understand and there's a mixture of responsibilities between the MeasureCollection wrapper and the ApiEntity module
- class_eval is confusing/less explicit and less obvious to debug at runtime
- Enumerable doesn't fit the use case of the MeasureCollection implementation which uses the measures array to drive the behaviour it needs. It also wasn't implemented properly.
- Adding a presenter option means we have a global way of setting collection presenters rather than not really knowing what you're getting

### DIFF
--- a/app/models/concerns/declarable.rb
+++ b/app/models/concerns/declarable.rb
@@ -9,10 +9,12 @@ module Declarable
     has_many :footnotes
     has_many :import_measures, class_name: 'Measure',
                                wrapper: MeasureCollection,
-                               filter: :without_excluded
+                               filter: :without_excluded,
+                               presenter: MeasurePresenter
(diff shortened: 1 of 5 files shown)


alexdesi (approved):
Nice one! It's easier to follow now.

PR #719, opened 2022-03-17
HOTT-1411: Support polymorphic quota search goods nomenclature links

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1411

### What?

Quota definitions belong to a measure which is associated with either a Heading, a Subheading or a Commodity.

Currently the quota search results only points at commodities for the goods nomenclature that comes back for the definition measure.

I have added/removed/altered:

- [x] Update quota results to point at the correct goods nomenclature

### Why?

I am doing this because:

- This is required so the user gets directed to the correct goods nomenclature


lib/tariff_jsonapi_parser.rb
```diff
@@ -44,12 +45,17 @@ def parse_collection(collection)
...
+    parent.merge!('type' => resource['type'])
+  end
+
+  def parse_attributes!(resource, parent)
     parent.merge!(resource['attributes'])
   end
 
   def parse_relationships!(relationships, parent)
     relationships.each do |name, values|
+
```
jebw: Minor but do seeing a reason for the additional whitespace?

lib/tariff_jsonapi_parser.rb
```diff
@@ -44,12 +45,17 @@ def parse_collection(collection)
     end
   end
 
-  def parse_top_level_attributes!(resource, parent)
+  def parse_type!(resource, parent)
+    parent.merge!('type' => resource['type'])
```
jebw: This feels like quite a generic attribute name with potential for clashes with the attributes in the api itself - the original attribute is obviously called `type` but this is internal only and and in JSON-API they hive it off to its own location to avoid clashes, just unfortunate ApiEntity forces us to munge it all together - maybe `resource_type`, `resource_class_name` or `relationship_class` instead - AFAICT its only for our own use internally (currently on the `has_one` in `ApiEntity`) ?

app/views/search/common/_goods_nomenclature.html.erb
```diff
@@ -1,10 +1,16 @@
 <tr class="govuk-table__row">
   <td class="govuk-table__cell">
-    <a href="<%= commodity_path(goods_nomenclature) %>" aria-describedby="code-for-commodity-<%= goods_nomenclature.id %>">
-      <%= goods_nomenclature.description.html_safe %>
-      <span class="govuk-visually-hidden">
-        (code <%= goods_nomenclature.code %>)
-      </span>    
+    <% if goods_nomenclature.instance_of?(GoodsNomenclature) %>
+        <a href="<%= commodity_path(goods_nomenclature) %>" aria-describedby="code-for-commodity-<%= goods_nomenclature.id %>">
```
jebw: minor - double indentation?

jebw (changes requested):
Looks really nice addition - only item I'd query is the actual use of `type` as an attribute key - a bit too easy to run into a clash at some future point?

lib/tariff_jsonapi_parser.rb
willfish: I'll use resource_type

willfish:
Sounds good re the type attribute naming. I've renamed to resource_type

jebw (approved):
LGTM

app/views/search/common/_goods_nomenclature.html.erb
```diff
@@ -1,10 +1,16 @@
...
-    <a href="<%= commodity_path(goods_nomenclature) %>" aria-describedby="code-for-commodity-<%= goods_nomenclature.id %>">
-      <%= goods_nomenclature.description.html_safe %>
-      <span class="govuk-visually-hidden">
-        (code <%= goods_nomenclature.code %>)
-      </span>    
+    <% if goods_nomenclature.instance_of?(GoodsNomenclature) %>
+      <a href="<%= commodity_path(goods_nomenclature) %>" aria-describedby="code-for-commodity-<%= goods_nomenclature.id %>">
+    <% else %>
+      <a href="<%= polymorphic_path(goods_nomenclature) %>" aria-describedby="code-for-commodity-<%= goods_nomenclature.id %>">
+    <% end  %>
```
alexdesi, line 7: To clean this we could:
`<a href="<%= goods_nomenclature_url(goods_nomenclature) %>" aria-describedby="code-for-commodity-<%= goods_nomenclature.id %>">`

and add a helper on `goods_nomenclature_path`

actually, there is already a `goods_nomenclature_path`, we could generalize that.

alexdesi (approved):

app/views/search/common/_goods_nomenclature.html.erb
willfish: I think we can actually get rid of our use of commodity path in a separate PR and then we can remove the conditional

PR #730, opened 2022-03-23
HOTT-1414: Tidy up footnote model and search result view

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Removed unused footnote methods
- [x] Refactored all_goods_nomenclature to return a list of uniq and sorted goods nomenclatures for the foonote search result view

### Why?

I am doing this because:

- I've been reviewing this code to determine whether we need to make changes for the polymorphic goods nomenclature

### DIFF
--- a/app/models/footnote.rb
+++ b/app/models/footnote.rb
@@ -3,8 +3,6 @@ require 'api_entity'
 class Footnote
   include ApiEntity
 
-  ECO_CODE = '05002'.freeze
-
   collection_path '/footnotes'
 
   attr_accessor :code, :footnote_type_id, :footnote_id, :description, :formatted_description, :extra_large_measures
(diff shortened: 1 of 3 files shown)


app/models/footnote.rb
```diff
@@ -3,20 +3,18 @@
...
   has_many :goods_nomenclatures
 
-  def id
-    @id ||= "#{casted_by.destination}-#{casted_by.id}-footnote-#{code}"
-  end
-
-  def eco?
-    code == ECO_CODE
+  def all_goods_nomenclatures
+    @all_goods_nomenclatures ||= measures
```
alexdesi, line 14: `measures.map(&:goods_nomenclature)`
will trigger `measures.length` queries, it might be slow, should we use a Sequel query here?

alexdesi:
test failure 🙀

app/models/footnote.rb
willfish: I think you might be confusing the frontend/backend, here

willfish:
see https://github.com/trade-tariff/trade-tariff-frontend/pull/731 @alexdesi

alexdesi (approved):
✅

PR #731, opened 2022-03-23
Enable setting resource type in factories

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Enable setting the resource type in factories
- [x] Set the resource type in the goods nomenclature model

### Why?

I am doing this because:

- We need this to simulate polymorphic measure goods nomenclature relationships in factories

### DIFF
--- a/spec/lib/api_entity_spec.rb
+++ b/spec/lib/api_entity_spec.rb
@@ -17,6 +17,22 @@ RSpec.describe ApiEntity do
     end
   end
 
+  describe '#resource_type' do
+    subject(:resource_type) { mock_entity.new(attributes).resource_type }
+
+    context 'when initialized with the resource type' do
+      let(:attributes) { { resource_type: 'foo' } }
+
+      it { is_expected.to eq('foo') }
+    end
+
+    context 'when initialized without the resource type' do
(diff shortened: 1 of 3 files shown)


spec/lib/api_entity_spec.rb
```diff
@@ -17,6 +17,22 @@ def self.name
...
+    context 'when initialized with the resource type' do
+      let(:attributes) { { resource_type: 'foo' } }
+
+      it { is_expected.to eq('foo') }
+    end
+
+    context 'when initialized without the resource type' do
+      let(:attributes) { {} }
+
+      it { is_expected.to eq('mock_entity') }
```
alexdesi, line 32: this is weird 🤔

willfish: How comes? Have you looked at the mock_entity implementation? It might help beyond just this diff

willfish: https://github.com/trade-tariff/trade-tariff-frontend/blob/main/spec/lib/api_entity_spec.rb#L14-L16

alexdesi (approved):
✅

PR #733, opened 2022-03-24
HOTT-1414: Tidy additional code search

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Added method for enumerating goods nomenclature in additional code model
- [x] Added missing coverage for additional code model
- [x] Removed unused methods in additional code model

### Why?

I am doing this because:

- Just keeping things tidy

### DIFF
--- a/app/presenters/measure_presenter.rb
+++ b/app/presenters/measure_presenter.rb
@@ -8,7 +8,7 @@ class MeasurePresenter < SimpleDelegator
   end
 
   def has_additional_code?
-    additional_code.present?
+    additional_code.code.present?
   end
 
   def has_measure_footnotes?
--- a/app/models/additional_code.rb
+++ b/app/models/additional_code.rb
@@ -9,15 +9,10 @@ class AdditionalCode
 
   has_many :measures
 
-  def id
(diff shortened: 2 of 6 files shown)


app/presenters/measure_presenter.rb
```diff
@@ -8,7 +8,7 @@ def has_measure_conditions?
   end
 
   def has_additional_code?
-    additional_code.present?
```
jebw, line 11: Previously:

Additional code model present with code: True
Additional code model present without code: False
Additional code model not present: False

Now:

Additional code model present with code: True
Additional code model present without code: False
Additional code model not present: Exception

Will there always be an AdditionalCode model present?

jebw (commented):
Looks good but there's a definite change in behaviour - just not sure if its a scenario that ever occurs?

app/presenters/measure_presenter.rb
willfish: This is true. AFAIK we never iterate over nothing/nil record values so this should be fine

willfish: I think it was checking the model because a delegator existed for present? from the model to its code attribute value

willfish: And this was just a byproduct of that

PR #737, opened 2022-03-28
HOTT-1414: Refactor how we handle goods nomenclature and api entity nil relationships

willfish:
### Jira link

### What?

I have added/removed/altered:

- [x] Added support for duplicate goods nomenclature in all search results (additional code, certificates, footnotes and quotas)
- [x] Added coverage for all_goods_nomenclatures method
- [x] Added support for nil relationships in has_one (has_many already works) - updating the api entity and jsonapi parser modules
- [x] Compact nil goods_nomenclature

### Why?

I am doing this because:

- We are seeing duplicates in production

### DIFF
--- a/app/models/order_number/definition.rb
+++ b/app/models/order_number/definition.rb
@@ -41,7 +41,7 @@ class OrderNumber
     end
 
     def all_goods_nomenclatures
-      measures.map(&:goods_nomenclature).uniq(&:goods_nomenclature_item_id).sort_by(&:goods_nomenclature_item_id)
+      measures.map(&:goods_nomenclature).compact.uniq(&:goods_nomenclature_item_id).sort_by(&:goods_nomenclature_item_id)
     end
 
     def geographical_areas
--- a/app/models/order_number/definition.rb
+++ b/app/models/order_number/definition.rb
@@ -41,7 +41,11 @@ class OrderNumber
     end
 
     def all_goods_nomenclatures
-      measures.map(&:goods_nomenclature).compact.uniq(&:goods_nomenclature_item_id).sort_by(&:goods_nomenclature_item_id)
+      measures
+        .map(&:goods_nomenclature)
+        .compact
+        .uniq(&:goods_nomenclature_item_id)
+        .sort_by(&:goods_nomenclature_item_id)
     end
 
     def geographical_areas
--- a/app/models/order_number/definition.rb
+++ b/app/models/order_number/definition.rb
@@ -40,6 +40,10 @@ class OrderNumber
       @id ||= quota_definition_sid
     end
 
+    def all_goods_nomenclatures
+      measures.map(&:goods_nomenclature).uniq(&:goods_nomenclature_item_id).sort_by(&:goods_nomenclature_item_id)
+    end
+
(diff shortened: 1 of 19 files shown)


app/models/order_number/definition.rb
```diff
@@ -40,6 +40,10 @@ def id
       @id ||= quota_definition_sid
     end
 
+    def all_goods_nomenclatures
+      measures.map(&:goods_nomenclature).uniq(&:goods_nomenclature_item_id).sort_by(&:goods_nomenclature_item_id)
```
jebw: this needs to handle measures without goods nomenclatures (ie `.compact` after `.map`)

jebw (dismissed):

app/models/order_number/definition.rb
```diff
@@ -40,6 +40,14 @@ def id
       @id ||= quota_definition_sid
     end
 
+    def all_goods_nomenclatures
+      measures
+        .map(&:goods_nomenclature)
+        .compact
+        .uniq(&:goods_nomenclature_item_id)
+        .sort_by(&:goods_nomenclature_item_id)
```
alexdesi: this piece of code is duplicated 4 times now,
it'd be great to dry this out, not sure which class should host that code for now,
maybe `MeasureCollection`? 🤔

willfish: I'm not sure. There wasn't a very good place for it and the implementation is slightly different for footnotes so I ended up just leaving as is. I can move it to measure collection but it would break the common return signature of that class

alexdesi (approved):

spec/factories.rb
```diff
@@ -464,9 +464,14 @@
...
   factory :additional_code do
-    code { Forgery(:basic).text }
+    additional_code_type_id { Forgery(:basic).text(exactly: 1).upcase }
+    additional_code { Forgery(:basic).text(exactly: 3).upcase }
+    code { Forgery(:basic).text(exactly: 4).upcase }
     description { Forgery(:basic).text }
+    formatted_description { Forgery(:basic).text }
+    measures { [attributes_for(:measure)] }
   end
+  j
```
willfish: Remove this

lib/api_entity.rb
```diff
@@ -140,13 +140,16 @@ def has_one(association, opts = {})
...
 
       define_method("#{association}=") do |attributes|
-        class_name = if options[:polymorphic]
-                       attributes['resource_type'].classify
-                     else
-                       options[:class_name]
-                     end
-
-        entity = class_name.constantize.new((attributes.presence || {}).merge(casted_by: self))
+        entity = if attributes.nil?
```
jebw, line 143: This looks like a change of behaviour - previously a relationship without attributes (which is valid I think) - though initialize `options[:class_name]` with an empty hash (in the non polymorphic scenario)

Now it sets the relationship to nil instead? What if the relationship entity is just a join model? We have that scenario in permutations?

willfish: It is definitely a change in behaviour. We're going from not supporting (exploding) on null entities, to supporting null entities

jebw (approved):
LGTM

PR #741, opened 2022-03-29
HOTT-1381 Show banner news items

jebw:
### Jira link

[HOTT-1381](https://transformuk.atlassian.net/browse/HOTT-1381)

### What?

I have added/removed/altered:

- [x] Added querying the latest banner from NewsItem
- [x] Added caching of that query
- [x] Show the banner news item in the header banner
- [x] Changed the govspeak sanitizer to allow `<a target="_blank"` links

### Why?

I am doing this because:

- We want to drive the header banner from the news items, controlled via the admin area.

- Affects every page, should not be merged before the backend changes are released

### DIFF
--- a/app/models/news_item.rb
+++ b/app/models/news_item.rb
@@ -64,7 +64,7 @@ class NewsItem
       end
     end
 
-    def cached_banner
+    def cached_latest_banner
       Rails.cache.fetch(banner_cache_key, expires_in: BANNER_CACHE_LIFETIME) do
         latest_banner
       end
--- a/app/models/news_item.rb
+++ b/app/models/news_item.rb
@@ -14,7 +14,7 @@ class NewsItem
   collection_path '/news_items'
 
   attr_accessor :id, :title, :content, :display_style, :show_on_xi, :show_on_uk,
-                :show_on_updates_page, :show_on_home_page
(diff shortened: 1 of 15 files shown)


app/models/news_item.rb
```diff
@@ -53,6 +64,12 @@ def any_updates?
       end
     end
 
+    def cached_banner
```
alexdesi: Looking where this is called, 
it looks it does not matter if it is cached or not,
I guess we could call this just `last_banner` (or maybe, even just `banner`)

config/initializers/govspeak_sanitizer.rb
```diff
@@ -0,0 +1,16 @@
...
+    Sanitize::Config.merge(
+      orig_config,
+      attributes: {
+        'a' => orig_config[:attributes]['a'] + %w[target],
+      },
+    )
+  end
+end
+
+Govspeak::HtmlSanitizer.prepend TariffGovspeakSanitizer
```
alexdesi, line 16: monkey patching is a last resort, wondering if we cannot just replace that piece of content (not very elegant, but maybe less bad than monkey patching (and easier to understand).

alexdesi (approved):
It looks good to me✅
I've discussed some potential improvements (not blocking).

PR #751, opened 2022-04-01
Fixed specs when last month only has 5 letters

jebw:
### Jira link

HOTT-????

### What?

I have added/removed/altered:

- [x] Fixed spec to allow for when last month only has 5 letters

### Why?

I am doing this because:

- Specs on the frontend are currently failing


alexdesi (approved):
it looks good ✅

PR #756, opened 2022-04-05
HOTT-1448: Adds from to date expression to default modal

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1448

### What?

Screenshot from 2022-04-05 14-30-40
Screenshot from 2022-04-05 14-30-47

I have added/removed/altered:

- [x] Added calendar.png asset
- [x] Added from to expression and tests

### Why?

I am doing this because:

- This is required as part of a new modal dialgo ui


jebw (approved):
Code looks good, couple of subjective comments

1. The image could go in webpacker handled assets, that way it'll end up with long lived cache lifetimes and if we sort the CDN out the user will only ever need to download it once?
2. The use of icons is discouraged by GDS but I'm guessing this is there in the original design?
3. Feels like showing a from/to is useful elsewhere - could this be a 'common' helper method?

PR #757, opened 2022-04-05
HOTT-1448: Only show guidance on UK data

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1468

### What?

I have added/removed/altered:

- [x] Only show guidance on UK service data

### Why?

I am doing this because:

- CDS and CHIEF guidance is never relevant on European data

### DIFF
--- a/app/views/search/certificate_search.html.erb
+++ b/app/views/search/certificate_search.html.erb
@@ -35,34 +35,36 @@
         </div>
       </dl>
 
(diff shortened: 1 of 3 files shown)


jebw (approved):
Is it worth adding a view spec for the two scenarios (uk vs xi)?

willfish:
I honestly wasn't sure. I will why not

PR #758, opened 2022-04-05
Add guidance table.

alexdesi:
### Jira link

[hott-1464](https://transformuk.atlassian.net/browse/HOTT-1464)

### What?
Add guidance table:

Screenshot 2022-04-05 at 17 26 39

### DIFF
--- a/app/models/measure.rb
+++ b/app/models/measure.rb
@@ -173,10 +173,8 @@ class Measure
     end
   end
 
-  def measure_conditions_with_valid_guidance
-    measure_conditions.select do |mc|
-      mc.document_code.present? && (mc.guidance_cds.present? || mc.guidance_chief.present?)
-    end
+  def measure_conditions_with_guidance
+    measure_conditions.select(&:has_guidance?)
   end
 
   private
--- a/app/models/measure.rb
+++ b/app/models/measure.rb
@@ -173,6 +173,12 @@ class Measure
     end
   end
 
+  def measure_conditions_with_valid_guidance
+    measure_conditions.select do |mc|
+      mc.document_code.present? && (mc.guidance_cds.present? || mc.guidance_chief.present?)
+    end
+  end
+
   private
 
   def excluded_country_ids
--- a/app/views/measures/_guidance_table.html.erb
+++ b/app/views/measures/_guidance_table.html.erb
@@ -5,26 +5,27 @@
   <div class="govuk-details__text  govuk-!-font-size-16">
     <table class="govuk-table govuk-!-margin-bottom-2">
       <thead class="govuk-table__head">
-      <tr class="govuk-table__row">
-        <th scope="col" class="govuk-table__header">Document code</th>
-        <th scope="col" class="govuk-table__header">CDS guidance</th>
-        <th scope="col" class="govuk-table__header">CHIEF guidance</th>
-      </tr>
+        <tr class="govuk-table__row">
+          <th scope="col" class="govuk-table__header">Document code</th>
+          <th scope="col" class="govuk-table__header">CDS guidance</th>
+          <th scope="col" class="govuk-table__header">CHIEF guidance</th>
+        </tr>
       </thead>
 
       <tbody class="govuk-table__body">
-      <% measure_conditions_with_valid_guidance.each do |mc| %>
-        <tr class="govuk-table__row">
+        <% measure_conditions_with_guidance.each do |mc| %>
+          <tr class="govuk-table__row">
             <td class="govuk-table__cell">
-              <%= mc.document_code %></td>
+              <%= mc.document_code %>
+            </td>
             <td class="govuk-table__cell">
               <%= govspeak mc.guidance_cds %>
             </td>
             <td class="govuk-table__cell">
               <%= govspeak mc.guidance_chief %>
             </td>
-        </tr>
-      <% end %>
+          </tr>
+        <% end %>
       </tbody>
     </table>
   </div>
--- a/app/views/measures/_guidance_table.html.erb
+++ b/app/views/measures/_guidance_table.html.erb
@@ -1,0 +1,31 @@
+<details class="govuk-details" data-module="govuk-details">
+  <summary class="govuk-details__summary">
+    <span class="govuk-details__summary-text">Guidance for completing Box 44 or Data Element 2/3</span>
+  </summary>
+  <div class="govuk-details__text  govuk-!-font-size-16">
+    <table class="govuk-table govuk-!-margin-bottom-2">
+      <thead class="govuk-table__head">
+      <tr class="govuk-table__row">
+        <th scope="col" class="govuk-table__header">Document code</th>
+        <th scope="col" class="govuk-table__header">CDS guidance</th>
+        <th scope="col" class="govuk-table__header">CHIEF guidance</th>
+      </tr>
+      </thead>
+
+      <tbody class="govuk-table__body">
+      <% measure_conditions_with_valid_guidance.each do |mc| %>
+        <tr class="govuk-table__row">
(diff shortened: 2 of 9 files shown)


app/models/measure.rb
```diff
@@ -173,6 +173,12 @@ def additional_code_sort
     end
   end
 
+  def measure_conditions_with_valid_guidance
+    measure_conditions.select do |mc|
+      mc.document_code.present? && (mc.guidance_cds.present? || mc.guidance_chief.present?)
+    end
+  end
+
```
alexdesi, line 179: ops, I'll add test for this!

app/models/measure.rb
```diff
@@ -173,6 +173,12 @@ def additional_code_sort
     end
   end
 
+  def measure_conditions_with_valid_guidance
+    measure_conditions.select do |mc|
+      mc.document_code.present? && (mc.guidance_cds.present? || mc.guidance_chief.present?)
```
willfish: You could make the measure condition model responsible for knowing about what it means to have guidance:

```ruby
measure_conditions.select(&:has_guidance?)
```

app/views/measures/_guidance_table.html.erb
```diff
@@ -0,0 +1,31 @@
+<details class="govuk-details" data-module="govuk-details">
+  <summary class="govuk-details__summary">
+    <span class="govuk-details__summary-text">Guidance for completing Box 44 or Data Element 2/3</span>
+  </summary>
+  <div class="govuk-details__text  govuk-!-font-size-16">
+    <table class="govuk-table govuk-!-margin-bottom-2">
+      <thead class="govuk-table__head">
+      <tr class="govuk-table__row">
```
willfish: Is this indentation askew?

app/views/measures/_guidance_table.html.erb
```diff
@@ -0,0 +1,31 @@
+<details class="govuk-details" data-module="govuk-details">
```
willfish, line 1: We'll have to rename this file if we change the component type from a table to a summary list, say. How do you feel about just calling this partial `_guidance.html.erb?`

app/views/measures/_guidance_table.html.erb
```diff
@@ -0,0 +1,31 @@
...
+      <thead class="govuk-table__head">
+      <tr class="govuk-table__row">
+        <th scope="col" class="govuk-table__header">Document code</th>
+        <th scope="col" class="govuk-table__header">CDS guidance</th>
+        <th scope="col" class="govuk-table__header">CHIEF guidance</th>
+      </tr>
+      </thead>
+
+      <tbody class="govuk-table__body">
+      <% measure_conditions_with_valid_guidance.each do |mc| %>
```
willfish: Think this indentation is off, too?

app/models/measure.rb
```diff
@@ -173,6 +173,12 @@ def additional_code_sort
     end
   end
 
+  def measure_conditions_with_valid_guidance
```
willfish:
```suggestion
  def measure_conditions_with_guidance
```

willfish: We probably don't know whether the guidance is valid

spec/models/measure_condition_spec.rb
```diff
@@ -6,4 +6,16 @@
   describe '#requirement' do
     it { expect(condition.requirement).to be_html_safe }
   end
+
+  describe 'attributes' do
+    it {
+      expect(condition).to respond_to(:condition,
```
willfish: We could use a shared example to achieve this/make this a validation of all api entity models:

```ruby
RSpec.shared_examples 'a resource that has attributes' do |attributes|
  subject(:resource) { described_class.new(attributes) }

  describe 'implemented attributes' do
    attributes.each do |attribute_key, attribute_value|
      it "defines an accessor for #{attribute_key}" do
        expect(resource.public_send(attribute_key)).to eq(attribute_value)
      end
    end
  end
end
  
it_behaves_like 'a resource that has attributes',
                id: 'flibble',
                condition_code: 'B',
                condition: 'B: Presentation of a certificate/licence/document',
                requirement: nil,
                action: 'The entry into free circulation is not allowed',
                duty_expression: '',
                condition_duty_amount: nil,
                condition_monetary_unit_code: nil,
                monetary_unit_abbreviation: nil,
                condition_measurement_unit_code: nil,
                condition_measurement_unit_qualifier_code: nil,
                measure_condition_components: []
```

lib/api_entity.rb
```diff
@@ -45,7 +45,7 @@ def url
     delegate :relationships, to: :class
 
     def inspect
-      keys_to_exclude = relationships + [:casted_by]
+      keys_to_exclude = relationships.to_a + [:casted_by]
```
willfish, line 48: I'm curious about the need for this

spec/factories.rb
```diff
@@ -470,6 +479,15 @@
     trait :universal_waiver do
       document_code { '999L' }
     end
+
+    trait :with_guidance do
+      guidace_cds { 'Guidance CDS' }
+      guidance_chief { 'Guidance CHIEF' }
+    end
+
+    initialize_with do
```
willfish: So I _think_ you don't need this

spec/models/measure_spec.rb
```diff
@@ -251,4 +251,22 @@
       it { expect(measure.excluded_country_list).to eq(expected_list) }
     end
   end
+
+  describe '#measure_conditions_with_valid_guidance' do
+    context 'with a valid condition guidance' do
+      let(:measure) { build :measure, :with_conditions_with_guidance }
+
+      it 'returns measure_conditions with valid guidance' do
```
willfish:
```suggestion
      it 'returns measure_conditions with guidance' do
```

I think we know we have/don't have guidance. I'm not sure this is valid/invalid test, etc

spec/models/measure_spec.rb
```diff
@@ -251,4 +251,22 @@
       it { expect(measure.excluded_country_list).to eq(expected_list) }
     end
   end
+
+  describe '#measure_conditions_with_valid_guidance' do
+    context 'with a valid condition guidance' do
+      let(:measure) { build :measure, :with_conditions_with_guidance }
```
willfish:
```suggestion
      subject(:measure) { build :measure, :with_conditions_with_guidance }
```

spec/models/measure_spec.rb
```diff
@@ -251,4 +251,22 @@
...
+
+      it 'returns measure_conditions with valid guidance' do
+        expect(measure.measure_conditions_with_valid_guidance.size).to eq(1)
+      end
+    end
+
+    context 'without condition guidances' do
+      let(:measure) { build :measure, :with_conditions }
```
willfish:
```suggestion
      subject(:measure) { build :measure, :with_conditions }
```

spec/models/measure_spec.rb
```diff
@@ -251,4 +251,22 @@
...
+
+      it 'returns no measure_conditions with valid guidance' do
```
willfish:
```suggestion
      it 'returns no measure_conditions with guidance' do
```

lib/api_entity.rb
alexdesi: to prevent this error:
`this is to fix NoMethodError: undefined method `+' for nil:NilClass`

spec/factories.rb
alexdesi: yep, we tried and we don't need it anymore.

app/views/measures/_guidance_table.html.erb
willfish: _measure_condition_guidances.html.erb

willfish (approved):

PR #760, opened 2022-04-07
HOTT-1448: Revert guidance hiding on XI

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1468

### What?

I have added/removed/altered:

- [x] Revert hiding guidance on XI

### Why?

I am doing this because:

- This is incorrect behaviour


alexdesi (approved):
✅

PR #762, opened 2022-04-07
HOTT-1475: Factory fixes

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1475

### What?

I have added/removed/altered:

- [x] Handle attributes assigned with default factorbot constructor
- [x] Adds missing setters/getters for consumed attributes
- [x] Removes unnecessary custom factory constructors

### Why?

I am doing this because:

- The factories are a bit of a mess and are inconsistently constructed
- We should be able to rely on the default constructor to build api entities

### DIFF
--- a/lib/api_entity.rb
+++ b/lib/api_entity.rb
@@ -83,7 +83,7 @@ module ApiEntity
 
     attributes.each do |name, value|
       if respond_to?(:"#{name}=")
-        send(:"#{name}=", value)
+        public_send(:"#{name}=", value)
       end
     end
   end
--- a/lib/api_entity.rb
+++ b/lib/api_entity.rb
@@ -45,9 +45,18 @@ module ApiEntity
     delegate :relationships, to: :class
 
     def inspect
-      keys_to_exclude = relationships.to_a + [:casted_by]
+      if @attributes.blank?
+        candidate_variables = instance_variables - %i[@attributes @casted_by]
 
-      @attributes.except(*keys_to_exclude)
+        candidate_variables.each_with_object({}) do |variable, acc|
+          key = variable.to_s.sub('@', '')
+          acc[key] = public_send(key)
+        end
+      else
+        keys_to_exclude = relationships.to_a + [:casted_by]
+
+        @attributes.except(*keys_to_exclude)
+      end
     end
 
     def resource_path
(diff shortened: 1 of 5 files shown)


lib/api_entity.rb
```diff
@@ -72,11 +81,9 @@ def initialize(attributes = {})
...
     @attributes = attributes
 
-    if attributes.present?
-      attributes.each do |name, value|
-        if respond_to?(:"#{name}=")
-          send(:"#{name}=", value.is_a?(String) && value == 'null' ? nil : value)
-        end
+    attributes.each do |name, value|
+      if respond_to?(:"#{name}=")
+        send(:"#{name}=", value)
```
jebw: `public_send` ? In theory `respond_to?` should only return true for public methods anyway but if its getting updated might be worth correcting

lib/api_entity.rb
```diff
@@ -72,11 +81,9 @@ def initialize(attributes = {})
   def attributes=(attributes = {})
     @attributes = attributes
 
-    if attributes.present?
-      attributes.each do |name, value|
-        if respond_to?(:"#{name}=")
-          send(:"#{name}=", value.is_a?(String) && value == 'null' ? nil : value)
```
jebw, line 78: Are we sure this type cast hack isn't needed anywhere? - does look wrong but presumably was added at some point for good reason

jebw (commented):
LGTM - couple of minor comments re the attribute assignment in ApiEntity

lib/api_entity.rb
willfish: Yeah, good shout :)

lib/api_entity.rb
willfish: I've never seen a value as the literal string 'null' in the wild looking at api responses. The ruby json libs would serialize a ruby `nil` as a json `null`  and I couldn't see anything in the serializers/models/presenters that might be doing this

willfish: It was added around the display of synonyms :eyes: https://github.com/trade-tariff/trade-tariff-frontend/commit/b48c050c376ad92b0b6e6fab443b9e9f0ad0950b

willfish: Yeah, this synonym code (really search references) isn't implemented in the frontend anymore and we're never going to be serializing the literal string 'null'. This could be a weird bug in the JSONAPI gem from 2012, maybe but this isn't possible AFAIK

jebw: Yeh, I'd say at the time they were hitting a bug in their serializer - that commits 10 years old, should be fine :+1:

PR #763, opened 2022-04-07
HOTT-1475: Separate factories

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1475

### What?

I have added/removed/altered:

- [x] Separate all factories into their own files (following pattern in the backend)

### Why?

I am doing this because:

- Its easier to navigate specific factories/only update a single responsibility file


jebw (approved):
LGTM

PR #766, opened 2022-04-11
HOTT-1449 Permutations modal

jebw:
### Jira link

[HOTT-1449](https://transformuk.atlassian.net/browse/HOTT-1449)

### What?

I have added/removed/altered:

- [x] Added ApiEntity models for the Permutation and Permutation Groups
- [x] Feature flagged showing the permutations dialog
- [x] Added a new 'permutations' dialog replacing the existing 'non overridden' measure conditions modals
- [x] Show the appropriate guidance for each permutation group

### Why?

I am doing this because:

- We wish to more clearly show which combinations of conditions are available to our users

### Screenshots

Screenshot from 2022-05-13 12-31-12
Screenshot from 2022-05-13 12-30-41
Screenshot from 2022-05-13 12-29-47
Screenshot from 2022-05-13 12-28-42
Screenshot from 2022-05-13 12-27-54

### DIFF
--- a/app/helpers/measures_helper.rb
+++ b/app/helpers/measures_helper.rb
@@ -82,8 +82,7 @@ module MeasuresHelper
   end
 
   def format_measure_condition_document_code(condition)
-    case condition.measure_condition_class
-    when 'threshold'
+    if condition.measure_condition_class.threshold?
       'Threshold condition'
     else
       condition.document_code
--- a/app/helpers/measures_helper.rb
+++ b/app/helpers/measures_helper.rb
@@ -73,7 +73,9 @@ module MeasuresHelper
         condition.requirement.presence || 'Condition not fulfilled'
       end
     else
-      (condition.certificate_description.presence || condition.requirement)
+      (condition.certificate_description.presence ||
+        condition.requirement.presence ||
+        'No document provided')
         .to_s
         .html_safe
     end
--- a/app/helpers/measures_helper.rb
+++ b/app/helpers/measures_helper.rb
@@ -61,4 +61,24 @@ module MeasuresHelper
       safe_join components.flatten.compact, ' '
     end
   end
+
+  def format_measure_condition_requirement(condition)
+    case condition.measure_condition_class
+    when 'threshold'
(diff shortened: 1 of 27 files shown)


app/models/measure_condition.rb
```diff
@@ -3,7 +3,11 @@
 class MeasureCondition
   include ApiEntity
 
+  WEIGHT_UNITS = %w[DTN DAP DHS GFI GRM GRT KGM KMA RET TNE].freeze
```
willfish, line 6: This feels like something we should add to the measure condition response

app/helpers/measures_helper.rb
```diff
@@ -61,4 +61,48 @@ def control_line_wrapping_in_duty_expression(expression)
       safe_join components.flatten.compact, ' '
     end
   end
+
+  def format_measure_condition_requirement(condition)
+    case condition.measure_condition_class
```
willfish, line 66: We're asking a lot of the condition in this method. Should we move some of the logic around presentation to the model?

willfish: You could also ask the measure condition class whether it's a threshold

```suggestion
    if condition.measure_condition_class.threshold?
```

app/views/measure_conditions/_permutation_group.html.erb
```diff
@@ -0,0 +1,33 @@
+<div class="permutation-group">
+  <p>
+    <% if permutation_group.permutations.many? %>
+    Meet one of the following conditions
+    <% else %>
+    Meet the following condition
```
willfish: Indentation issues :)

willfish:
Would be good to get some screenshots of the various permutations (including thresholds with weight/volume units and maybe some of the fallbacks, too). I'm guessing negative conditions just get filtered out, too

willfish (approved):
LGTM!

app/models/measure_condition.rb
jebw: Probably yes - I'll create a separate ticket for that, once we have it in the backend we can switch to using it from the frontend

PR #771, opened 2022-04-12
HOTT-1496 Fix usage of sentry-cli

jebw:
### Jira link

[HOTT-1496](https://transformuk.atlassian.net/browse/HOTT-1496)

### What?

I have added/removed/altered:

- [x] Downgraded the version of Sentry CLI we use in CI

### Why?

I am doing this because:

- our CLI usage seems to be incompatible with the newly released v2


alexdesi (approved):
✅  🚀

PR #772, opened 2022-04-12
HOTT-1493 Show full 10 digit commodity code on heading page.

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-1493

### What?
The end line comm codes should be still showing all of the 10 digits in all cases, not the cut down 8 digits etc

image


jebw (changes requested):
I don't think this is right - the abbreviate comm code helper was originally added because I non-declarables should be shortened. Checking the Commodites page, the subheadings are now showing as full 10 digit codes even though they cannot be declared?

willfish (approved):

jebw (approved):
LGTM

PR #775, opened 2022-04-14
HOTT-1507 Fix view specs not being run in CI

jebw:
### Jira link

[HOTT-1507](https://transformuk.atlassian.net/browse/HOTT-1507)

### What?

I have added/removed/altered:

- [x] Switch to using the latest v1 ruby Orb
- [x] Fixed failing specs in _commodity.html.erb_spec.rb 
- [x] Globally set `initialize_with` on all factories.

### Why?

I am doing this because:

We have failing view specs on `main` - these were missed because CI isn't running all specs. The number of specs being run jumps from **295** to **390**

This appears to be a bug in the CircleCI Ruby orb. It's already been fixed upstream but we're not benefiting because we're locked to a specific (old) version.

The Orb uses semver so switching `@1` allows us to benefit from bugfixes whilst avoid any switches to incompatible versions.

Spec fixes:
* `_commodity.html.erb_spec.rb` - caused by css selector not being updated in specs to match code change
* factory fixes - `initialize_with` lines had been dropped from factories but some models (notably Search) require these for correct instantiation. The `initialize_with { new attributes }` actually better matches our ApiEntity usage, and expectations of any PORO classes so setting that as the default makes sense.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 
 orbs:
   aws-cli: circleci/aws-cli@2.0.3
-  ruby: circleci/ruby@1.4.0
+  ruby: circleci/ruby@1
   node: circleci/node@4.7.0
   browser-tools: circleci/browser-tools@1.1
   cloudfoundry: circleci/cloudfoundry@1.0
--- a/spec/support/factoy_bot.rb
+++ b/spec/support/factoy_bot.rb
@@ -1,0 +1,3 @@
+FactoryBot.define do
+  initialize_with { new attributes }
+end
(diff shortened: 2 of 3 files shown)


.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2.1
 
 orbs:
   aws-cli: circleci/aws-cli@2.0.3
-  ruby: circleci/ruby@1.4.0
+  ruby: circleci/ruby@1
```
alexdesi, line 5: 👍

spec/support/factoy_bot.rb
```diff
@@ -0,0 +1,3 @@
+FactoryBot.define do
+  initialize_with { new attributes }
```
alexdesi, line 2: I guess it is still used in some factories.

alexdesi (approved):
it looks good! ✅

spec/support/factoy_bot.rb
jebw: Anything which has attributes setting via the constructor but not via a method - `Search` being the key culprit.

PR #777, opened 2022-04-14
HOTT-1482 Link subheadings on the commodities page tree

jebw:
### Jira link

[HOTT-1482](https://transformuk.atlassian.net/browse/HOTT-1482)

### What?

I have added/removed/altered:

- [x] Link the subheadings items shown in the commodities tree

### Why?

I am doing this because:

- We now have a subheadings page and it will help our users to find related commodity codes - particularly if the code they are looking at is similar to what they require but not quite right

### Have you? (optional)

- None


willfish (approved):
Yay!

PR #778, opened 2022-04-19
HOTT-1485 Fix for rendering bullets in guidance tables

jebw:
### Jira link

[HOTT-1485](https://transformuk.atlassian.net/browse/HOTT-1485)

### What?

I have added/removed/altered:

- [x] Added missing `tariff-markdown` classes to rendered markdown

### Why?

I am doing this because:

- The alignment of bullets is off on the guidance table
- The govuk design system did have fixes for these elements but these were not being applied because govspeak renders 'unclassed' HTML nodes
- We have a css class to automatically apply appropriate `govuk-*` css classes to unclassed HTML elements

### Before

Screenshot from 2022-04-19 14-56-18

### After

Screenshot from 2022-04-19 14-51-12

### DIFF
--- a/app/views/measures/_guidance_table.html.erb
+++ b/app/views/measures/_guidance_table.html.erb
@@ -18,10 +18,10 @@
             <td class="govuk-table__cell">
               <%= mc.document_code %>
             </td>
-            <td class="govuk-table__cell">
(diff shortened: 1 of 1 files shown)


alexdesi (approved):
cool ✅

PR #779, opened 2022-04-20
HOTT-1379 commodities history behaviour

jebw:
### Jira link

[HOTT-1379](https://transformuk.atlassian.net/browse/HOTT-1379)

### What?

I have added/removed/altered:

- [x] Dropped the automatic redirects for the search since they are too simplistic
- [x] Redirect subheadings from the commodities endpoint

### Why?

I am doing this because:

- We want to match the search and navigation patterns described in the 'scenarios' document attached to the Jira ticket linked above


app/controllers/commodities_controller.rb
```diff
@@ -1,7 +1,13 @@
 class CommoditiesController < GoodsNomenclaturesController
+  rescue_from Faraday::ResourceNotFound, with: :show_validity_periods
+
   helper_method :uk_declarable, :xi_declarable, :declarable
 
   def show
+    unless declarable.declarable?
+      return redirect_to polymorphic_path(declarable)
```
willfish, line 8:
```ruby
      def find_commodity
        @commodity = Commodity.actual
                              .declarable
                              .by_code(params[:id])
                              .eager(:goods_nomenclature_indents, :goods_nomenclature_descriptions, :footnotes)
                              .take

        raise Sequel::RecordNotFound if commodity_has_children?
        raise Sequel::RecordNotFound if @commodity.goods_nomenclature_item_id.in? HiddenGoodsNomenclature.codes
      end
```

app/controllers/commodities_controller.rb
```diff
@@ -1,7 +1,13 @@
 class CommoditiesController < GoodsNomenclaturesController
+  rescue_from Faraday::ResourceNotFound, with: :show_validity_periods
+
   helper_method :uk_declarable, :xi_declarable, :declarable
 
   def show
+    unless declarable.declarable?
```
willfish, line 7: So this is for headings that might not be declarable, right?

willfish: AFAIK the commodities controller never returns non-declarable subheadings

app/models/subheading.rb
```diff
@@ -38,6 +38,10 @@ def to_s
     formatted_description || description
   end
 
+  def declarable?
```
willfish: :tada:

willfish: I believe this comes back in the api but why not

willfish: Actually, should we just let the api tell us this? Every goods nomenclature does

willfish: https://github.com/trade-tariff/trade-tariff-backend/blob/main/app/serializers/api/v2/subheadings/subheading_serializer.rb#L18-L20

spec/models/search_spec.rb
```diff
@@ -90,72 +90,4 @@
       it { is_expected.to be true }
     end
   end
-
-  describe '#search_term_is_commodity_code?' do
```
willfish, line 94: Wow, actually tested :laughing:

willfish:
Looks awesome. Small comment about duplicating the declarable concept

willfish (approved):

app/models/subheading.rb
jebw: Ah, ok - will drop this

spec/models/search_spec.rb
jebw: :laughing: I tend to aim for 100%, sincerely doubt I get it - more then anything it avoids me typoing something then scratching my head why tests further up the stack fail in odd places

jebw: tbh if that whole describe block took 5 minutes I'd be surprised, wrote one bit then cut n pasted lots, should probably have used a shared behaviour in hindsight

willfish: :tada:

PR #785, opened 2022-04-26
HOTT-1504 Fix for redirecting to url instead of path

jebw:
### Jira link

[HOTT-1504](https://transformuk.atlassian.net/browse/HOTT-1504)

### What?

I have added/removed/altered:

- [x] Replace generation of URLs for redirects, with absolute paths

### Why?

I am doing this because:

- If the site is being accessed on a different url to the one configured as the canonical url, then the redirect will trigger `ActionController::Redirecting::UnsafeRedirectError`
- This can be observed when accessing the sites via the GovUK PaaS URLs and doing a search

### DIFF
--- a/app/controllers/goods_nomenclatures_controller.rb
+++ b/app/controllers/goods_nomenclatures_controller.rb
@@ -17,7 +17,7 @@ class GoodsNomenclaturesController < ApplicationController
     results = @search.perform
 
     if results.exact_match?
-      redirect_to url_for(results.to_param.merge(url_options))
+      redirect_to url_for(results.to_param.merge(url_options).merge(only_path: true))
     else
       redirect_to sections_url
     end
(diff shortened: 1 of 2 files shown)


app/controllers/goods_nomenclatures_controller.rb
```diff
@@ -17,7 +17,7 @@ def find_relevant_goods_code_or_fallback
     results = @search.perform
 
     if results.exact_match?
-      redirect_to url_for(results.to_param.merge(url_options))
+      redirect_to url_for(results.to_param.merge(url_options).merge(only_path: true))
```
alexdesi, line 20: An alternative approach would be to use the `:host` option.
But I am not sure which approach is better.

alexdesi (approved):
It looks good to me ✅ 
It would be nice to have a test, in case someone removes the `only_path` in the future (by thinking it is redundant)

app/controllers/goods_nomenclatures_controller.rb
jebw: Whats the advantage? We'd need to use `host: request.host` otherwise we'd potentially still run into the same issue (ie `request.host != default_url_options[:host]`)

If we just use internal paths only the problem goes away?

PR #798, opened 2022-05-03
HOTT-1550 Added search serializer for subheadings

jebw:
### Jira link

[HOTT-1550](https://transformuk.atlassian.net/browse/HOTT-1550)

### What?

I have added/removed/altered:

- [x] Added a search serializer for subheadings

### Why?

I am doing this because:

- Any search results for `/search.json` which included subheadings would result in an error


alexdesi (approved):
nice one! ✅

PR #800, opened 2022-05-04
HOTT-1549: Fix links for Subheadings in a-z-index page

alexdesi:
### Jira link
[HOTT-1549](https://transformuk.atlassian.net/browse/HOTT-1549)

### What?

Fix links for Subheadings in a-z-index page.

### Why?
The links are broken, they point to a 404 Not found page.

### DIFF
--- a/app/presenters/search_reference_presenter.rb
+++ b/app/presenters/search_reference_presenter.rb
@@ -10,9 +10,16 @@ class SearchReferencePresenter
   end
 
   def link
-    reference_link = "/#{APP_SLUG}/#{referenced_class.tableize}/#{referenced_id}"
+    "/#{APP_SLUG}/#{referenced_class.tableize}/#{composite_id}"
+  end
 
-    append_product_line_suffix(reference_link)
+  def composite_id
+    case referenced_class
+    when 'Subheading'
+      "#{referenced_id}-#{productline_suffix}"
+    else
+      referenced_id
+    end
   end
 
   private
--- a/app/presenters/search_reference_presenter.rb
+++ b/app/presenters/search_reference_presenter.rb
@@ -10,12 +10,18 @@ class SearchReferencePresenter
   end
 
   def link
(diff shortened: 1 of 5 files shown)


app/presenters/search_reference_presenter.rb
```diff
@@ -10,12 +10,18 @@ def to_s
   end
 
   def link
-    "/#{APP_SLUG}/#{referenced_class.tableize}/#{referenced_id}"
+    reference_link = "/#{APP_SLUG}/#{referenced_class.tableize}/#{referenced_id}"
```
jebw: This will clearly work fine but feels a bit complicated because the logic is around the link rather then around the id generation? An alternative would be another method to return the appropriate id? eg

```ruby
def link
  "/#{APP_SLUG}/#{referenced_class.tableize}/#{composite_id}"
end

def composite_id
  case referenced_class
  when 'Subheading'
    "#{referenced_id}-#{productline_suffix}"
  else
    referenced_id
  end
end
```

FYI - why `case` and no `if` / `else` ? Its because its makes it clear it is special casing according to `referenced_class` and sets up a pattern for extending with other types in the future if necessary.

if thats too long then a ternary would have the same effect but without the future extensibility

```ruby
def composite_id
  referenced_class == 'Subheading' ? "#{referenced_id}-#{productline_suffix}" : referenced_id
end
```

jebw (approved):
LGTM - looks great but I've included a suggestion for clarifying the code?

jebw:
This doesn't actually hard depend on the additional key from the backend, I think if the key is not included we'll continue to generate invalid urls (different `/subheadings/1111111111-` instead of `/commodities/1111111111` but still invalid)

Thats not really any different from the users viewpoint and it starts working properly as soon as the backend deploys is fix for the additional key

app/presenters/search_reference_presenter.rb
alexdesi: I agree, it's more readable.

PR #802, opened 2022-05-04
HOTT-1564 Fix for changing date on subheadings page

jebw:
### Jira link

[HOTT-1564](https://transformuk.atlassian.net/browse/HOTT-1564)

### What?

I have added/removed/altered:

- [x] Added support for generating subheading paths from goods_nomenclature_path

### Why?

I am doing this because:

- we currently generate a 500 when the user clicks the 'change date' link on a subheadings page

- Changes a generic path helper which is used in multiple places

### DIFF
--- a/app/models/subheading.rb
+++ b/app/models/subheading.rb
@@ -1,4 +1,7 @@
 class Subheading < GoodsNomenclature
+  # Commodity code: 10 chars  + separator: 1 char +  Product Line Suffix: 2 chars
+  FULL_CODE_LENGTH = 13
+
   include ApiEntity
 
   has_one :section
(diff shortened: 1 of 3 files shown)


app/models/subheading.rb
```diff
@@ -1,4 +1,6 @@
 class Subheading < GoodsNomenclature
+  SHORT_CODE_LENGTH = 13
```
alexdesi: Is this still considered "Short" code?

Also:
It'd be great to add a comment on this like, to clarify the number 13, something like:
```
# Commodity code: 10 chars  + separator: 1 char +  Product Line Suffix: 2 chars 
```

alexdesi (approved):
it looks good to me ✅

app/models/subheading.rb
jebw: Have renamed it to FULL_CODE and add comment

alexdesi (approved):

PR #813, opened 2022-05-16
HOTT-1483 Show ancestors tree on Subheading page

jebw:
### Jira link

[HOTT-1483](https://transformuk.atlassian.net/browse/HOTT-1483)

### What?

I have added/removed/altered:

- [x] Moved ancestors partial into common goods_nomenclatures folder to show it applies to more then declarables/commodities
- [x] Show the ancestors tree on the Subheading page
- [x] Changed the Subheading tree to only show commodities below the Subheading being viewed
- [x] Default Commodity Tree on Heading and Subheading pages to show open instead of closed

### Why?

I am doing this because:

- It will aid users with navigating the site

### Have you? (optional)

### DIFF
--- a/app/models/subheading.rb
+++ b/app/models/subheading.rb
@@ -42,7 +42,19 @@ class Subheading < GoodsNomenclature
   end
 
   def ancestors
-    []
+    @ancestors ||= [].tap { |ancestors|
+      commodity = find_self_in_commodities_list
+
+      while (commodity = commodity&.parent)
+        ancestors << commodity
+      end
+    }.reverse
+  end
+
+  def find_self_in_commodities_list
+    @find_self_in_commodities_list ||= commodities.find do |c|
+      c.goods_nomenclature_sid == goods_nomenclature_sid
+    end
   end
 
   private
--- a/app/models/subheading.rb
+++ b/app/models/subheading.rb
@@ -10,6 +10,7 @@ class Subheading < GoodsNomenclature
 
   has_many :footnotes
   has_many :commodities
+  has_many :ancestors, class_name: 'Commodity'
 
   collection_path '/subheadings'
 
@@ -39,16 +40,6 @@ class Subheading < GoodsNomenclature
 
   def to_s
     formatted_description || description
-  end
-
-  def ancestors
-    @ancestors ||= [].tap { |ancestors|
-      commodity = find_self_in_commodities_list
-
-      while (commodity = commodity&.parent)
-        ancestors << commodity
-      end
-    }.reverse
   end
 
   def find_self_in_commodities_list
--- a/app/models/subheading.rb
+++ b/app/models/subheading.rb
@@ -41,6 +41,10 @@ class Subheading < GoodsNomenclature
     formatted_description || description
   end
 
+  def ancestors
+    []
+  end
+
   private
 
   def harmonised_system_subheading_code
--- a/spec/factories/heading_factory.rb
+++ b/spec/factories/heading_factory.rb
@@ -15,8 +15,12 @@ FactoryBot.define do
 
     trait :with_commodity_tree do
       commodities do
-        attributes_for_list(:commodity, commodity_count || 2) do |c, i|
-          c[:parent_sid] = (i > 0 ? c[:goods_nomenclature_sid] - 1 : nil)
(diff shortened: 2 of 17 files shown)


spec/factories/subheading_factory.rb
```diff
@@ -1,13 +1,28 @@
...
+
+    chapter do
+      attributes_for :chapter,
+                     goods_nomenclature_item_id: "#{goods_nomenclature_item_id.first(2)}00000000"
+    end
+
+    heading do
+      attributes_for :heading,
+                     goods_nomenclature_item_id: "#{goods_nomenclature_item_id.first(4)}000000"
+    end
```
alexdesi, line 17: Should these be in traits to not slow down the other tests that don't need those?

alexdesi (dismissed):
it looks good to me ✅ 
only a comment

spec/factories/subheading_factory.rb
jebw: This is used to generated models simulating what we get back from the API, the API would include the heading data so the absence of heading makes the generated subheading not particularly functional or reflective of a 'real' API response. 

I don't think it should impact the performance in any measurable way - its just generating a hash of test data, not persisting anything.

app/models/subheading.rb
```diff
@@ -41,6 +41,22 @@ def to_s
     formatted_description || description
   end
 
+  def ancestors
```
willfish: We should just return this from the backend

spec/factories/heading_factory.rb
```diff
@@ -1,12 +1,24 @@
 FactoryBot.define do
   factory :heading do
+    transient do
+      commodity_count { nil }
```
willfish, line 4:
```suggestion
      commodity_count { 0 }
```

spec/factories/heading_factory.rb
```diff
@@ -1,12 +1,24 @@
...
+    end
+
     chapter { attributes_for(:chapter) }
     description { Forgery(:basic).text }
     formatted_description { Forgery(:basic).text }
     goods_nomenclature_item_id { '0101000000' }
 
-    commodities { [] }
+    commodities { attributes_for_list :commodity, commodity_count || 0 }
```
willfish, line 12:
```suggestion
    commodities { attributes_for_list :commodity, commodity_count }
```

spec/factories/heading_factory.rb
```diff
@@ -1,12 +1,24 @@
...
     import_measures { [] }
     export_measures { [] }
+
+    trait :with_commodity_tree do
+      commodities do
+        attributes_for_list(:commodity, commodity_count || 2) do |c, i|
```
willfish: I'm assuming c is a commodity and i is the index of the list of the commodities starting out at 0?

spec/models/subheading_spec.rb
```diff
@@ -86,9 +86,26 @@
...
+    subject { subheading.ancestors.map(&:goods_nomenclature_sid) }
+
+    let :subheading do
+      build :subheading, goods_nomenclature_sid: commodities[3][:goods_nomenclature_sid],
+                         commodities:
+    end
+
+    let :commodities do
+      attributes_for_list :commodity, 5 do |c, i|
+        c[:parent_sid] = (i > 0 ? c[:goods_nomenclature_sid] - 1 : nil)
```
willfish: There's a lot of baked in programmatic understanding in the fixtures, here. Is there any chance we could just be explicit with some sids to simulate real data/also make the fixtures a bit more dumb?

spec/factories/heading_factory.rb
jebw: I'll change them to commodity and index

spec/models/subheading_spec.rb
jebw: Problem with hard coding sid's (or any ids) is that you can only use the factory once per spec (you couldn't create 2 subheadings with tree's of commodities under them for example).

jebw: I'm updating this to use manually specified ids but since I'm having to specify more ids I'm pretty sceptical that it _reduces_ the knowledge subheading has of commodity tree's - it seems to result in an increase

willfish (approved):

PR #818, opened 2022-05-17
Updated ApiEntity#inspect to include class name

jebw:
### Jira link

HOTT-????

### What?

I have added/removed/altered:

- [x] Changed the output of ApiEntity#inspect to return a string
- [x] Include the Class name in the inspect output to make it clear the object being viewed

### Why?

I am doing this because:

- When navigating relationships in the console during development its to understand the object being looked at
- Previously there was no different in the output from `FactoryBot.build` and `FactoryBot.attributes_for` despite one giving you an entity instance and one giving you a Hash
- This aligns with how ActiveRecord formats its output and is well understood

- Makes changes to our api client

### DIFF
--- a/lib/api_entity.rb
+++ b/lib/api_entity.rb
@@ -45,18 +45,15 @@ module ApiEntity
     delegate :relationships, to: :class
 
     def inspect
-      if @attributes.blank?
-        candidate_variables = instance_variables - %i[@attributes @casted_by]
+      attrs = if defined?(@attributes) && @attributes.any?
+                attributes_for_inspect
+              else
+                instance_variables_for_inspect
+              end
 
-        candidate_variables.each_with_object({}) do |variable, acc|
-          key = variable.to_s.sub('@', '')
-          acc[key] = public_send(key)
-        end
-      else
-        keys_to_exclude = relationships.to_a + [:casted_by]
+      inspection = attrs.any? ? attrs.join(', ') : 'not initialised'
 
-        @attributes.except(*keys_to_exclude)
-      end
+      "#<#{self.class.name.presence || self.class} #{inspection}>"
     end
 
     def resource_path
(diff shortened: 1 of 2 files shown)


lib/api_entity.rb
```diff
@@ -45,18 +45,21 @@ def url
...
-        candidate_variables.each_with_object({}) do |variable, acc|
-          key = variable.to_s.sub('@', '')
-          acc[key] = public_send(key)
-        end
-      else
-        keys_to_exclude = relationships.to_a + [:casted_by]
-
-        @attributes.except(*keys_to_exclude)
-      end
+      inspection = if defined?(@attributes) && @attributes.any?
```
willfish: Reads quite a lot more complex than before (lots happening in a condensed space)

lib/api_entity.rb
```diff
@@ -45,18 +45,21 @@ def url
...
+                     instance_variables.without(%i[@attributes @casted_by])
+                                       .map { |k| k.to_s.sub('@', '') }
+                                       .select(&method(:respond_to?))
+                                       .map { |k| "#{k}: #{public_send(k).inspect}" }
+                                       .join(', ')
+                   else
+                     'not initialised'
+                   end
+
+      "#<#{self.class.name.presence || self.class} #{inspection}>"
```
willfish, line 56: Do we also want the object id?

willfish (approved):

lib/api_entity.rb
jebw: Can do - personally that wouldn't mean much to me - I'm more likely to care about the 'id' attribute, eg `goods_nomenclature_item_id` - I was trying to follow the output of ActiveRecord as closely as possible on the theory that theirs must work well or someone would have changed it by now

lib/api_entity.rb
jebw: Yeh, I was trying to avoid adding a nest of methods just for inspect but I'll split it out into private methods

lib/api_entity.rb
willfish: I had a review of the active record inspect implementation. I don't think we'll want it on second thoughts. I've only ever needed it when learning about how ruby works, tbh, and that was a few :moon: :moon: ago

PR #819, opened 2022-05-17
Disable RSpec dsl in main

willfish:
### Jira link

HOTT-<TODO>

### What?

I have added/removed/altered:

- [x] Avoid monkey patched rspec dsl

### Why?

I am doing this because:

- This is consistent with the rails generators and it is generally more desirable to do things in one way, only: https://github.com/rspec/rspec-rails/commit/ca0d249858903949052e06884e8e7f9d596cdc79


jebw (approved):
LGTM

PR #826, opened 2022-05-20
HOTT-1633 suggestions dropdown fixes

jebw:
### Jira link

[HOTT-1633](https://transformuk.atlassian.net/browse/HOTT-1633)

### What?

I have added/removed/altered:

- [x] Highlight when the app is waiting for a search result
- [x] Submit the users typed search result when pressing enter instead of the first most recent search result
- [x] Improve debounce behaviour

### Why?

I am doing this because:

- The search field is submitting partially completed search results if the rate the user types exceeds the response rate from the search suggestions api

### DIFF
--- a/spec/features/search_banner_spec.rb
+++ b/spec/features/search_banner_spec.rb
@@ -15,18 +15,18 @@ RSpec.describe 'Search banner', js: true do
             page.find('.tariff-search-banner .autocomplete__input').click
 
             page.find('.tariff-search-banner .autocomplete__input').set('gold')
-            sleep 1
+            sleep 1.5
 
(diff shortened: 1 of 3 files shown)


spec/features/search_banner_spec.rb
```diff
@@ -15,18 +15,18 @@
             page.find('.tariff-search-banner .autocomplete__input').click
 
             page.find('.tariff-search-banner .autocomplete__input').set('gold')
-            sleep 1
+            sleep 1.5
```
willfish, line 18: Ooph

willfish (approved):

PR #827, opened 2022-05-20
HOTT-1567: Make use of redis_cache_store instead of outdated redis_st…

alexdesi:
### What?
Make use of redis_cache_store instead of outdated redis_store.

Why?

The reason of this PR is because Rails 5.2.0 (and following versions) includes a Redis cache store out of the box, so you don't need this gem anymore if you just need to store the cache in Redis.
For more: https://github.com/redis-store/redis-activesupport

### Jira link
https://transformuk.atlassian.net/browse/HOTT-1567

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -40,7 +40,7 @@ gem 'puma'
 gem 'rack-cors'
 
 # Redis
-gem 'redis-rails'
+gem 'redis'
 
 # AWS
 gem 'aws-sdk-rails'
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -40,7 +40,7 @@ gem 'puma'
 gem 'rack-cors'
 
 # Redis
-gem 'redis-rails'
+gem 'redis'
```
alexdesi, line 43: I had to include the Redis gem explicitly.
I guess that was not necessary in the BE because it is using the `redis-elasticache` which requires Redis.

jebw (approved):
LGTM :raised_hands:

willfish:
Can we add a why/explanation that doesn't require access to jira?

willfish (approved):

PR #836, opened 2022-05-26
HOTT-1654 Switch to Welsh using url path

jebw:
### Jira link

[HOTT-1654](https://transformuk.atlassian.net/browse/HOTT-1654)

### What?

I have added/removed/altered:

- [x] Added path based switching of locale
- [x] Added fallback handling for when a translation is not available in the selected locale
- [x] Added a stub welsh locale, currently feature flagged
- [x] Added LOCALE_PATH and PREFIX_PATH service tags for including the locale prefix in URLs in News Item content

### Why?

I am doing this because:

- We wish to enable translations of the site into Welsh in the future


willfish (approved):
LGTM

PR #838, opened 2022-05-26
Drop the now redundant news items feature flag

jebw:
### Jira link

HOTT-???

### What?

I have added/removed/altered:

- [x] Dropped the news items feature flag

### Why?

I am doing this because:

- It is no long needed, we always want news items enabled now


alexdesi (approved):
it looks good to me 🧹 🧹 🧹 ✅

PR #839, opened 2022-05-26
HOTT-1635: Add preferential treatment on Rule of Origin tab

alexdesi:
### What?
Add an optional section to show preferential treatment on the Rule of Origin tab.
The new sections are stored on a YAML file, I think that is appropriate for now since there is only one entry. More will come soon (but not so many).

### Jira link
https://transformuk.atlassian.net/browse/HOTT-1635

### Why?
To show custom information about preferential treatment for certain countries.


alexdesi:
Screenshot 2022-05-26 at 17 23 13

alexdesi:
Screenshot 2022-05-26 at 17 23 13

config/rules_of_origin_tab.yml
```diff
@@ -0,0 +1,10 @@
+---
+rules_of_origin_tab:
```
willfish: Just my 2c but I'm wondering whether we should put this in `config/locales/*` since it feels like some text that we might want to translate in future?

willfish: I think you can then do:

```ruby
I18n.t("rules_of_origin_tab.preferential_treatment.#{country_id}", default: '').html_safe
```

willfish: I actually think you might be able to do:

```ruby
I18n.t("rules_of_origin_tab.preferential_treatment.#{country_id}_html", default: '') # spot the _html
```

willfish: https://guides.rubyonrails.org/i18n.html#using-safe-html-translations

alexdesi: good point,
I used a different locale file instead of putting all inside `en.yml` because the preferential treatments are pretty long chunk of HTML.

spec/helpers/commodities_helper_spec.rb
```diff
@@ -157,4 +157,21 @@
...
+      it { is_expected.to be_html_safe }
+      it { is_expected.to be_present }
+    end
+
+    context 'when the country does NOT have preferential treatment in locals' do
+      let(:geographical_area_id) { 'IT' }
+
+      it { is_expected.to be_empty }
+    end
+  end
```
alexdesi: Now that we are using locales, test tests are less valuable.
I am thinking to remove them.

config/locales/rules_of_origin.yml
```diff
@@ -0,0 +1,16 @@
+---
+en:
+  rules_of_origin:
+    preferential_treatment:
+      UA:
```
alexdesi: this could be confusing since it is not the language used, but the county code :)

willfish: So I think we should just put this in en.yml so we can easily handover localised strings for translation?

willfish: I also think only the first key is the actual locale :tada: so it shouldn't be too confusing

config/rules_of_origin_tab.yml
willfish: So I actually think that's okay for reasons I mention, here: https://github.com/trade-tariff/trade-tariff-frontend/pull/839/files#r883507520

spec/helpers/commodities_helper_spec.rb
willfish: I'm not actually sure a helper method makes sense for this anymore

willfish (approved):

PR #841, opened 2022-05-30
HOTT-1638 initial wizard screens

jebw:
### Jira link

[HOTT-1638](https://transformuk.atlassian.net/browse/HOTT-1638)

### What?

I have added/removed/altered:

- [x] Added a Wizard for the Rules of Origin
- [x] Added a first import/export step

### Why?

I am doing this because:

- We want to guide the user through the Rules of Origin process


app/views/rules_of_origin/steps/show.html.erb
```diff
@@ -0,0 +1,21 @@
+<% content_for :title, t('.meta_title',  step: current_step.model_name.human) %>
+
+<section id="rules-of-origin-<%= current_step.key %>">
+  <%= link_to 'Back', step_path(wizard.previous_key), class: 'govuk-back-link' %>
```
willfish: We should we translate this string?

app/views/rules_of_origin/steps/show.html.erb
```diff
@@ -0,0 +1,21 @@
...
+        <%= form.govuk_error_summary %>
+
+        <%= render current_step.key, form: form, wizard: wizard %>
+
+        <%= form.govuk_submit %>
+      <% end %>
+    </div>
+
+    <div class="govuk-grid-column-one-third">
+      Step listing goes here
```
willfish, line 18: We should we translate this string?

willfish (approved):
Simple and awesome

app/views/rules_of_origin/steps/show.html.erb
jebw: not necesssary - its just stub text

PR #845, opened 2022-05-31
Fix broken tests for _proof.html view

alexdesi:
### What?
Fix broken tests for _proof.html view

### DIFF
--- a/spec/views/rules_of_origin/_proofs.html.erb_spec.rb
+++ b/spec/views/rules_of_origin/_proofs.html.erb_spec.rb
@@ -1,10 +1,13 @@
 require 'spec_helper'
 
-RSpec.describe 'rules_of_origin/_proofs', type: :view do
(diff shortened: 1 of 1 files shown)


willfish:
Not sure I understand this. What's going on here @alexdesi?

jebw (approved):
@willfish This should resolve the failing specs on `main`

The partial had been changed to rely on global state (a `@search`) variable - presumably this change got merged with failing specs? This mocks out that state resolving the failing specs

In general, since this is a partial and not a page template I think it might be better if the partial was changed to rely upon locals being explicitly passed into it, rather then assuming the presence of global state? 

@alexdesi would be good if you could fix the rubocop failure prior to merge - this doesn't relate to your change but should be a one liner to correct

alexdesi:
Guys, since we are already using VCR on the views, and since the relevant cassette does exist,
I am using VCR.

Rubocop fixed.

jebw (approved):
LGTM

PR #870, opened 2022-06-20
HOTT-1689 drop permutations feature flag

jebw:
### Jira link

[HOTT-1689](https://transformuk.atlassian.net/browse/HOTT-1689)

### What?

I have added/removed/altered:

- [x] Dropped the permuations feature flag

### Why?

I am doing this because:

- We have now deployed permutations dialogs to production

### DIFF
--- a/app/presenters/measure_presenter.rb
+++ b/app/presenters/measure_presenter.rb
@@ -48,6 +48,6 @@ class MeasurePresenter < SimpleDelegator
   end
 
   def permutations_enabled?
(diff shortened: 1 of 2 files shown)


alexdesi (approved):
It looks good ✅ 
Things to check: is this affecting the tests? (actually, they are green, maybe they test only the new PERMUTATIONS = true behaviour)

jebw:
@alexdesi I don't think there were any specs for the old views

PR #898, opened 2022-07-05
HOTT-1619: Update measure type description for Channel Islands

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-1659

### What?
Update measure type description for the Channel Islands.

### Why?

I am doing this because:

> We have been asked to alter the visible display of measure type “103” for the geographical area “1080”, which is the geo. area for “Channel Islands”

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -92,6 +92,9 @@ en:
     1400:
       103: 'Channel Islands duty'
       105: 'Channel Islands duty'
+    1080:
+      103: 'UK-CD Customs Union'
+      105: 'UK-CD Customs Union'
   service_banner:
     service_name:
       uk: "UK Integrated Online Tariff"


config/locales/en.yml
```diff
@@ -92,6 +92,8 @@ en:
     1400:
       103: 'Channel Islands duty'
       105: 'Channel Islands duty'
```
alexdesi, line 94: I am wondering if we should also update these to "UK-CD Customs Union"

config/locales/en.yml
```diff
@@ -92,6 +92,8 @@ en:
...
+    1080:
+      103: 'UK-CD Customs Union'
```
alexdesi, line 96: ... and is we should we also include the code `105` (which stands for “Non preferential duty under authorised use”)

alexdesi:
Screenshot 2022-07-04 at 17 28 24

config/locales/en.yml
willfish: The answer to this question is in the database:

```ruby
Measure.where(geographical_area_id: '1080').distinct(:measure_type_id).pluck(:measure_type_id)
=> ["103"]
```

willfish: At least from a current state of play perspective

willfish: I think it would be harmless to add this for 105 and would avoid us falling into another change later down the line

config/locales/en.yml
willfish: We can safely remove these AFAIK:

```ruby
# on uk
Measure.where(geographical_area_id: '1400').distinct(:measure_type_id).pluck(:measure_type_id)
=> []
# on xi
Measure.where(geographical_area_id: '1400').distinct(:measure_type_id).pluck(:measure_type_id)
=> []
```

willfish:
Couple of small things to change:

1. Future proof with coverage of 105 measure type
2. ~~Remove dead overrides for 1400~~

config/locales/en.yml
willfish: Okay, ignore this - we're just wayyyy ahead of CDS people and these measures are coming soon

willfish (approved):

PR #903, opened 2022-07-07
HOTT-1724: Add support for n/a on duty rate.

alexdesi:
### What?
Add support for n/a for specific combinations of geographical areas and measure types.

### Jira link
https://transformuk.atlassian.net/browse/HOTT-1724

### DIFF
--- a/app/helpers/measures_helper.rb
+++ b/app/helpers/measures_helper.rb
@@ -1,8 +1,11 @@
 module MeasuresHelper
   def filter_duty_expression(measure)
+    # NIHIL is an abbreviation (see backend DutyExpressionDescription model),
+    # our best guess is that it is used to indicate that a previously applied measure
+    # no longer carries any duties.
     return '' if measure.duty_expression.to_s == 'NIHIL'
 
-    control_line_wrapping_in_duty_expression(measure.duty_expression.to_s)
+    measure.measure_type_duty_label || control_line_wrapping_in_duty_expression(measure.duty_expression.to_s)
   end
 
   def legal_act_regulation_url_link_for(measure)
--- a/app/models/measure.rb
+++ b/app/models/measure.rb
@@ -179,6 +179,10 @@ class Measure
     measure_conditions.select(&:has_guidance?)
   end
 
+  def measure_type_duty_label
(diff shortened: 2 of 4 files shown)


spec/factories/duty_expression_factory.rb
```diff
@@ -12,5 +12,13 @@
...
       description { 'VAT' }
     end
+
+    trait :with_geographical_area do
+      attributes_for { :geographical_area }
+    end
+
+    trait :with_measure_type do
+      attributes_for { :measure_type }
+    end
```
alexdesi: these can be removed

app/helpers/measures_helper.rb
```diff
@@ -1,8 +1,11 @@
 module MeasuresHelper
   def filter_duty_expression(measure)
+    # NIHIL is an abbreviation (see backend DutyExpressionDescription model),
+    # our best guess is that the it is used to indicate that a previously applied measure
```
willfish:
```suggestion
    # our best guess is that it is used to indicate that a previously applied measure
```

app/models/measure.rb
```diff
@@ -179,6 +179,12 @@ def measure_conditions_with_guidance
     measure_conditions.select(&:has_guidance?)
   end
 
+  def measure_type_duty_label
+    return unless geographical_area.id && measure_type.id
+
+    I18n.t("measure_type_duties.#{geographical_area.id}.#{measure_type.id}", default: nil)
```
willfish, line 183:
```suggestion
    I18n.t("measure_type_duties.#{geographical_area&.id}.#{measure_type&.id}", default: nil)
```

You can drop the guard above with this syntax. I honestly don't think it's necessary in real life, though. Possibly you added it for the tests

app/models/measure.rb
```diff
@@ -179,6 +179,12 @@ def measure_conditions_with_guidance
     measure_conditions.select(&:has_guidance?)
   end
 
+  def measure_type_duty_label
+    return unless geographical_area.id && measure_type.id
```
willfish: I think we should remove this

alexdesi: yes, I've checked, and measures always have geo ID and measure type ID

alexdesi:
Screenshot 2022-07-07 at 14 55 23

willfish (approved):

PR #912, opened 2022-07-13
HOTT-1749 Display new message for EPS measure conditions with condition code V

maverickvi:
### Jira link

HOTT-1749

### What?

I have added/removed/altered:

- [ ] Added new message

### Why?

I am doing this because:

- We want to display the new message for EPS measure conditions with condition code V on XI

### DIFF
--- a/app/helpers/measures_helper.rb
+++ b/app/helpers/measures_helper.rb
@@ -68,7 +68,9 @@ module MeasuresHelper
   def format_measure_condition_requirement(condition)
     case condition.measure_condition_class
     when 'threshold'
-      if condition.is_price_condition?
+      if condition.is_eps_condition_code_v?
+        safe_join ['The price of your goods is greater than or equal to', condition.requirement], ' '
+      elsif condition.is_price_condition?
         safe_join ['The price of your goods does not exceed', condition.requirement], ' '
       elsif condition.is_weight_condition?
         safe_join ['The weight of your goods does not exceed', condition.requirement], ' '
--- a/app/helpers/measures_helper.rb
+++ b/app/helpers/measures_helper.rb
@@ -68,7 +68,7 @@ module MeasuresHelper
   def format_measure_condition_requirement(condition)
     case condition.measure_condition_class
     when 'threshold'
-      if condition.is_eps_condition_code_v?
+      if condition.is_eps_condition?
         safe_join ['The price of your goods is greater than or equal to', condition.requirement], ' '
       elsif condition.is_price_condition?
         safe_join ['The price of your goods does not exceed', condition.requirement], ' '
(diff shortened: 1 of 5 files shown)


app/helpers/measures_helper.rb
```diff
@@ -68,7 +68,9 @@ def control_line_wrapping_in_duty_expression(expression)
   def format_measure_condition_requirement(condition)
     case condition.measure_condition_class
     when 'threshold'
-      if condition.is_price_condition?
+      if condition.is_eps_condition_code_v?
```
willfish: I think we could probably do with localising these strings? @jebw not sure how you feel

app/models/measure_condition.rb
```diff
@@ -44,4 +45,8 @@ def is_weight_condition?
   def is_volume_condition?
     VOLUME_UNITS.include? condition_measurement_unit_code
   end
+
+  def is_eps_condition_code_v?
```
willfish:
```suggestion
  def is_eps_condition?
```

Technically v and eps are synonymous

willfish: https://github.com/trade-tariff/trade-tariff-backend/blob/main/spec/factories/measure_factory.rb#L281-L285

spec/helpers/measures_helper_spec.rb
```diff
@@ -153,6 +153,12 @@
         it { is_expected.to match 'The price of your goods does not exceed' }
       end
 
+      context 'with eps condition code V' do
+        let(:condition) { build :measure_condition, :eps, condition_code: 'V' }
```
willfish:
```suggestion
        let(:condition) { build :measure_condition, :eps }
```

We should move the condition code V to the factory trait since :eps means V

app/helpers/measures_helper.rb
jebw: I'd say not, we're not localising the rest of the site only RoO at the moment - if the decision is made to do that we need to do it en-masse, until then though it will just add another layer of indirection.

app/models/measure_condition.rb
```diff
@@ -5,6 +5,7 @@ class MeasureCondition
 
   WEIGHT_UNITS = %w[DTN DAP DHS GFI GRM GRT KGM KMA RET TNE].freeze
   VOLUME_UNITS = %w[HLT KLT LPA LTR MIL MTQ].freeze
+  CONDITION_CODE_V = 'V'.freeze
```
jebw: Can this constant be renamed to `EPS_CONDITION` or similar to clarify what it is representing

jebw (changes requested):
:+1: Looks spot, one point is just the constant still referencing V and we now know that V represents EPS

app/models/measure_condition.rb
```diff
@@ -44,4 +45,8 @@ def is_weight_condition?
   def is_volume_condition?
     VOLUME_UNITS.include? condition_measurement_unit_code
   end
+
+  def is_eps_condition?
+    CONDITION_CODE_V == condition_code && is_price_condition? && is_weight_condition?
```
willfish:
```suggestion
    CONDITION_CODE_V == condition_code
```

jebw (approved):
LGTM :raised_hands:

willfish (approved):

PR #919, opened 2022-07-18
Display scheme title on components definition page

maverickvi:
### Jira link

HOTT-1730

### What?

I have added/removed/altered:

- [x] Displayed scheme title on components definition page

### Why?

I am doing this because:

- The scheme title was missing


jebw (approved):
LGTM :+1:

PR #927, opened 2022-07-19
Remove switching links on error pages

maverickvi:
### Jira link

HOTT-1762

### What?

I have added/removed/altered:

- [x] Removed switching links on 404, 500, 503 pages

### Why?

I am doing this because:

- We don't want switching links on error pages


jebw (approved):
LGTM :+1:

PR #929, opened 2022-07-20
Added back to top link on ROO proof requirements

maverickvi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1716

### What?

I have added/removed/altered:

- [ ] Added back to top link on ROO proof requirements

### Why?

I am doing this because:

- having a Back to top link makes the user experience better

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -5,7 +5,6 @@ en:
   generic:
     warning: Warning
     contents: Contents
-    back_to_top: Back to top
 
   meta_description: "Search for import and export commodity codes and for tax, duty and licences that apply to your goods"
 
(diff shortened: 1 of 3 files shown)


config/locales/en.yml
```diff
@@ -5,6 +5,7 @@ en:
   generic:
     warning: Warning
     contents: Contents
+    back_to_top: Back to top
```
alexdesi: I guess it's better to put this under  `navigation`. There is already a key called `back`  

```
  navigation:
    back: Back
```

maverickvi: Done, thanks.

alexdesi (approved):
✅

PR #934, opened 2022-07-25
HOTT-1712 subdivisions

jebw:
### Jira link

[HOTT-1712](https://transformuk.atlassian.net/browse/HOTT-1712)

### What?

I have added/removed/altered:

- [x] Added subdivision selector step

### Why?

I am doing this because:

- We need to ask the user which subdivision they are checking when there are multiple subdivisions

### Have you? (optional)

- Low, feature flagged off

### Screenshots

Screenshot from 2022-07-25 12-29-50

### DIFF
--- a/app/models/rules_of_origin/steps/subdivision.rb
+++ b/app/models/rules_of_origin/steps/subdivision.rb
@@ -1,0 +1,30 @@
+module RulesOfOrigin
+  module Steps
+    class Subdivision < Base
+      self.section = 'originating'
+
+      delegate :description, to: :commodity, prefix: true
+
+      attribute :subdivision
+      validates :subdivision, inclusion: { in: :available_subdivisions }
+
+      def skipped?
+        @store['sufficient_processing'] == 'no'
+      end
+
+      def options
+        chosen_scheme.rule_sets
+      end
+
+    private
+
+      def commodity
+        @commodity ||= Commodity.find(commodity_code)
+      end
+
+      def available_subdivisions
+        options.map(&:resource_id)
+      end
+    end
+  end
+end
--- a/app/models/rules_of_origin/steps/subdivision.rb
+++ b/app/models/rules_of_origin/steps/subdivision.rb
@@ -13,7 +13,7 @@ module RulesOfOrigin
       end
 
       def options
-        chosen_scheme.rule_sets
+        chosen_scheme.rule_sets.select(&:subdivision)
       end
 
     private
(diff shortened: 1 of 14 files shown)


app/models/rules_of_origin/steps/subdivision.rb
```diff
@@ -0,0 +1,30 @@
...
+    class Subdivision < Base
+      self.section = 'originating'
+
+      delegate :description, to: :commodity, prefix: true
+
+      attribute :subdivision
+      validates :subdivision, inclusion: { in: :available_subdivisions }
+
+      def skipped?
+        @store['sufficient_processing'] == 'no' || options.none?
```
alexdesi: should we check for the capitalized `NO` ?
(maybe downcasing it?)

app/views/rules_of_origin/steps/_about_commodity.html.erb
```diff
@@ -0,0 +1,15 @@
+<%= render 'shared/details', summary: t('.summary') do %>
+  <%# full key necessary for i18n because we're inside shared/details %>
```
alexdesi, line 2: maybe remove this comment?

config/locales/en.yml
```diff
@@ -233,6 +238,9 @@ en:
         sufficient_processing:
           Have non-originating parts been subject to sufficient processing to
           qualify for preferential treatment?
+      rules_of_origin_steps_subdivision:
+        subdivision: Provide more information about your product
```
alexdesi: it looks a bit redundant to use `subdivision` under `rules_of_origin_steps_subdivision`,
at least a first sight (maybe there is a good reason)

alexdesi (approved):
This looks great, I love the simple readable small methods.
Just added few naive comments
✅

app/models/rules_of_origin/steps/subdivision.rb
jebw: This is a value from a radio button which we control so I think its fine to rely on our own Ids

app/views/rules_of_origin/steps/_about_commodity.html.erb
jebw: I'd left it there for future developers - if they use abbreviated keys then they'll probably not realise why they don't work - since they need to be under `shared.details` instead of `rules_of_origin.steps.about_commodity` in the `en.yml`?

config/locales/en.yml
jebw: These are the govuk form builders default keys, not ones I'm adding - this is equivalent to

```yaml
model_name_space_plus_model_class_underscored:
  attribute_name: Legend to show for attribute
```

jebw: I've renamed the step to `Subdivisions` and the attribute field to `subdivision_id` to make this clearer :+1:

PR #941, opened 2022-07-26
HOTT-1775: Return empty array if article content is nil

maverickvi:
### Jira link

[HOTT-1775](https://transformuk.atlassian.net/browse/HOTT-1775)

### What?

I have added/removed/altered:

- [x] modified the processes_sections method to return an empty array if article content is nil

### Why?

I am doing this because:

- We were getting errors when trying to processes empty sections


jebw (approved):
LGTM

PR #957, opened 2022-08-04
Hott 1791 VAT and Excise messages

maverickvi:
### Jira link

[HOTT-1791](https://transformuk.atlassian.net/browse/HOTT-1791)

### What?

I have added/removed/altered:

- [x] Added Vat and Excise messages

### Why?

I am doing this because:

- It will make it easier for users to understand VAT and Excise


javier-transform (approved):
Looks good to me

willfish (approved):

PR #959, opened 2022-08-08
HOTT-1866: Rules text now uses html_safe

maverickvi:
### Jira link

[HOTT-1866](https://transformuk.atlassian.net/browse/HOTT-1866)

### What?

I have added/removed/altered:

- [ ] used html_safe to display each rule

### Why?

I am doing this because:

- We want rules to be displayed using HTML


javier-transform (approved):

alexdesi (approved):
it looks good to me ✅

PR #961, opened 2022-08-09
HOTT-1877 Fix infinite loop when query q='' on Beta search.

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-1877

### What?
Fix infinite loop when query q='' on Beta search.

### DIFF
--- a/app/services/beta/search/perform_search_service.rb
+++ b/app/services/beta/search/perform_search_service.rb
@@ -10,7 +10,7 @@ module Beta
       end
 
       def call
-        retries = 0
+        retries ||= 0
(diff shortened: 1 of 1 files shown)


app/services/beta/search/perform_search_service.rb
```diff
@@ -10,7 +10,7 @@ def initialize(query, filters = {})
       end
 
       def call
-        retries = 0
+        retries ||= 0
```
alexdesi, line 13: Important: without this the retries is reset every time (and does not keep track of the retries)

javier-transform (approved):

PR #969, opened 2022-08-17
HOTT-1893: Added help find commodity page

maverickvi:
### Jira link

[HOTT-1893](https://transformuk.atlassian.net/browse/HOTT-1893)

### What?

I have added/removed/altered:

- [ ] Added help find commodity page

### Why?

I am doing this because:

- We need a separate help page for commodities

### Have you? (optional)

### DIFF
--- a/app/webpacker/src/stylesheets/_help_find_commodity.scss
+++ b/app/webpacker/src/stylesheets/_help_find_commodity.scss
@@ -1,0 +1,33 @@
+.call-to-action, .gem-c-govspeak {
+  margin: 2em 0;
+  background-color: #f3f2f1;
+  padding: 2em;
+}
+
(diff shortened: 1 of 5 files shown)


javier-transform (approved):

app/webpacker/src/stylesheets/_help_find_commodity.scss
```diff
@@ -0,0 +1,33 @@
+.call-to-action, .gem-c-govspeak {
```
willfish, line 1: Do we really need all of this extra styling? Is there nothing built into the govuk libs that we can just use?

willfish (approved):
Seems fine to me. Can you paste a screenshot?

PR #971, opened 2022-08-18
HOTT-1867 - Redirect search at the frontend according to meta data

javier-transform:
### Jira link

[HOTT-1867](https://transformuk.atlassian.net/browse/HOTT-1867)

### What?

I have added/removed/altered:

- [x] Redirect search at the frontend according to meta data

### DIFF
--- a/app/services/beta/search/perform_search_service.rb
+++ b/app/services/beta/search/perform_search_service.rb
@@ -14,7 +14,6 @@ module Beta
         path = "/api/beta/search?#{query_params}"
         response = api.get(path)
         parsed = parse_jsonapi(response)
-        merge_meta_to_result!(parsed, response)
 
         Beta::Search::SearchResult.new(parsed)
       rescue Faraday::Error, UnparseableResponseError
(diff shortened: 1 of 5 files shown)


app/services/beta/search/perform_search_service.rb
```diff
@@ -14,6 +14,7 @@ def call
         path = "/api/beta/search?#{query_params}"
         response = api.get(path)
         parsed = parse_jsonapi(response)
+        merge_meta_to_result!(parsed, response)
```
willfish: This shouldn't be necessary AFAIK

javier-transform: When the parse_jsonapi returns the object there is no meta merged into it. There is a statement to check and merge meta inside the tariff parser api but the meta data does not exist in the `data` part of the response. That is why we need to merge the meta part and i added this line.
image

javier-transform: I didn't remove the code inside tariff parser api or alter it since there could be edge case(s) regarding that, and that's why i waited for your review.

willfish: Oh, okay. Probably we should add a parse_top_level_meta! method to api entity and add tests for that. I don't think it should affect any of the existing api entity models negatively/have any crazy edge cases

willfish (approved):

PR #995, opened 2022-08-26
HOTT-1901: Fix sidebar and extra content leaking into no_scheme view

maverickvi:
### Jira link

[HOTT-1901](https://transformuk.atlassian.net/browse/HOTT-1901)

### What?

I have added/removed/altered:

- [ ] Fixed sidebar and extra content leaking into no_scheme view

### Why?

I am doing this because:

- no_scheme view was broken
Screenshot 2022-08-26 at 16 03 56


alexdesi (approved):
✅ it looks good!

PR #1003, opened 2022-08-30
Exclude Name and Email from param logging

jebw:
### Jira link

HOTT-1730

### What?

I have added/removed/altered:

- [x] Added additional fields to the parameter filtering for logging

### Why?

I am doing this because:

- We have data in the feedbacks model we don't want logged

- Makes changes to the parameter filtering


jebw:
Closing because wrong ticket number - see #1004

PR #1005, opened 2022-08-30
HOTT-1897: Use headings API for heading codes on subdivision and prod…

maverickvi:
…uct specific rules pages

### Jira link

[HOTT-<1897>](https://transformuk.atlassian.net/browse/HOTT-1897)

### What?

I have added/removed/altered:

- [ ] Used headings API for heading codes on subdivision and product specific rules pages

### Why?

I am doing this because:

- We were using the Commodities API previously and it was causing 404 errors

Screenshot 2022-08-30 at 17 42 03

### DIFF
--- a/app/models/rules_of_origin/steps/product_specific_rules.rb
+++ b/app/models/rules_of_origin/steps/product_specific_rules.rb
@@ -3,8 +3,6 @@ module RulesOfOrigin
     class ProductSpecificRules < Base
       self.section = 'originating'
 
-      delegate :description, to: :commodity, prefix: true
-
       attribute :rule
       validates :rule, inclusion: { in: :available_rules }
 
@@ -20,11 +18,11 @@ module RulesOfOrigin
         subdivided_rules? ? subdivision_rules : all_rules
       end
 
-    private
-
-      def commodity
-        @commodity ||= Commodity.find(commodity_code)
+      def declareable_description
+        find_declarable.description
       end
+
+    private
 
       def none_option
         Struct.new(:resource_id, :rule).new('none', none_option_text)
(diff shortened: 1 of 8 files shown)


app/models/rules_of_origin/steps/product_specific_rules.rb
```diff
@@ -20,12 +18,12 @@ def rules
         subdivided_rules? ? subdivision_rules : all_rules
       end
 
-    private
-
-      def commodity
-        @commodity ||= Commodity.find(commodity_code)
```
jebw, line 26: This means we're dropping the 'cache in model instance' - not sure if this is a consideration but if there are multiple attributes request from a GoodsNomenclature (or the same multiple times) this will result in multiple API requests.

At somepoint we should look to cache api requests but for now maybe add an `||=` into `Base#find_declarable` ?

app/views/rules_of_origin/steps/_product_specific_rules.html.erb
```diff
@@ -15,4 +15,4 @@
 <% end %>
 
 <%= render 'about_commodity', commodity_code: current_step.commodity_code,
-                              description: current_step.commodity_description %>
+                              description: current_step.declareable_description %>
```
jebw: I'm not actually sure of the correct spelling but we've used `declarable` instead of `declareable` elsewhere in the codebase so its best to be consistent

jebw (changes requested):
Looks really good - couple of minor comments on spelling corrections and some api caching in `find_declarable`

Linting needs some tweaks as well but that looks quite straight forward to resolve

jebw (approved):
LGTM :raised_hands:

PR #1007, opened 2022-08-31
HOTT-1786 Support multiple wizard sessions

jebw:
### Jira link

HOTT-1786

### What?

I have added/removed/altered:

- [x] Updated routing for rules of origin wizard to include country and commodity
- [x] Used the country and commodity in the session data identifier

### Why?

I am doing this because:

- It will allow multiple concurrent sessions of the Rules of Origin wizard provided they are for different commodities and/or countries

- low, feature flagged off

### DIFF
--- a/app/controllers/rules_of_origin/steps_controller.rb
+++ b/app/controllers/rules_of_origin/steps_controller.rb
@@ -13,10 +13,10 @@ module RulesOfOrigin
 
     before_action :check_session_data,
                   except: :index,
-                  if: -> { params[:id] != 'start' }
+                  if: -> { params[:id] != wizard_class.steps.first.key }
 
     def show
-      if params[:id] == 'start'
+      if params[:id] == wizard_class.steps.first.key
         redirect_to return_to_commodity_path
       else
         super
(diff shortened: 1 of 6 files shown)


app/controllers/rules_of_origin/steps_controller.rb
```diff
@@ -13,10 +13,10 @@ class StepsController < ApplicationController
 
     before_action :check_session_data,
                   except: :index,
-                  if: -> { params[:id] != wizard_class.steps.first.key }
+                  if: -> { params[:id] != 'start' }
 
     def show
-      if params[:id] == wizard_class.steps.first.key
+      if params[:id] == 'start'
```
willfish: Do we have constant for these?

jebw: @willfish Not really though if we're going down that route I'm probably best just grabbing the first key as I was before?

app/controllers/rules_of_origin/steps_controller.rb
```diff
@@ -26,26 +26,31 @@ def show
...
-      rules_of_origin_step_path(step_id, ...)
+      rules_of_origin_step_path(params['commodity'],
+                                params['country'],
+                                step_id,
+                                ...)
     end
 
     def check_session_data
-      if service_name != wizard_store['service']
+      if service_name != wizard_store['service'] ||
```
willfish: Could we extrapolate these three questions into a method for readability?

```ruby
def check_session_data
      if incorrect_wizard_session?
      ....
      end
end

def incorrect_wizard_session?
  service_name != wizard_store['service'] || params['commodity'].blank? || params['country'].blank?
end
```

app/controllers/rules_of_origin/steps_controller.rb
```diff
@@ -26,26 +26,31 @@ def show
...
+          params['country'].blank?
 
         redirect_to return_to_commodity_path
         wizard_store.purge!
       end
     end
 
     def return_to_commodity_path
-      return find_commodity_path if wizard_store['commodity_code'].blank?
+      return find_commodity_path if params['commodity'].blank?
```
willfish: Are we passing an object as a param? commodity code/goods nomenclature code is more explicit/less misleading

willfish (approved):
Seems fine. Couple of small comments

app/controllers/rules_of_origin/steps_controller.rb
willfish: Mainly its just for the convenience of having one place to change the name. I doubt we'll ever change the name, though

PR #1008, opened 2022-09-01
HOTT-1931 - Adds changes to news page layout

javier-transform:
### Jira link

[HOTT-1931](https://transformuk.atlassian.net/browse/HOTT-1931)

### What?

I have added/removed/altered:

- [x] Adds changes to news page layout


jebw (changes requested):
Looks great, I particularly like the wrapping of content  - inconsistent line lengths just waste screen space :smile: 

Looks like there's a view spec which will need updating, and as a future FYI - we've got the automatic classes for `<p>` and `<a>` turned on (build in feature of govuk frontend) so the p tag classes aren't needed but they're not a problem either.

jebw (approved):
LGTM :+1:

PR #1009, opened 2022-09-01
HOTT-1918: Consolidate filtering country in the backend

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1918

### What?

I have added/removed/altered:

- [x] Removes country filtering on frontend
- [x] Moves country filtering to backend

### Why?

I am doing this because:

- We want a single place to control filtering on specific countries
- We have changed the backend to do this in a more sophisticated way for exclusions (see https://github.com/trade-tariff/trade-tariff-backend/pull/756)


app/controllers/commodities_controller.rb
```diff
@@ -84,7 +84,11 @@ def xi_commodity
   end
 
   def query_params
-    super.merge(filter: { meursing_additional_code_id: meursing_lookup_result.meursing_additional_code_id })
+    query = { filter: {} }
+    query[:filter][:meursing_additional_code_id] = meursing_lookup_result.meursing_additional_code_id
+    query[:filter][:geographical_area_id] = country if country.present?
+
+    super.merge(query)
```
willfish, line 91: We no longer render declarable headings in the headings controller - everything that can be filtered/has measures now is consolidated here

spec/requests/commodity_spec.rb
```diff
@@ -107,7 +107,7 @@
 
   context 'when commodity with country filter' do
     it 'will not display measures for other countries except for selected one' do
-      VCR.use_cassette('headings#show_0101') do
+      VCR.use_cassette('commodities#6911100090#show_filter_ad') do
```
willfish, line 110: We needed a new backend response since it is now doing the filtering

alexdesi (approved):
lgtm 🍏

PR #1011, opened 2022-09-02
HOTT-1933 - incorrect country name displayed on wholly obtained page

javier-transform:
### Jira link

[HOTT-1933](https://transformuk.atlassian.net/browse/HOTT-1933)

### What?

I have added/removed/altered:

- [x] Added a condition to render correct country name

### DIFF
--- a/app/views/rules_of_origin/steps/_not_wholly_obtained.html.erb
+++ b/app/views/rules_of_origin/steps/_not_wholly_obtained.html.erb
@@ -8,7 +8,7 @@
   </h1>
 
   <p class="govuk-body-l">
-    <%= t '.lead_para', trade_country_name: form.object.exporting? ? 'UK' : form.object.trade_country_name %>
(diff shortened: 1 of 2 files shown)


app/views/rules_of_origin/steps/_not_wholly_obtained.html.erb
```diff
@@ -8,15 +8,15 @@
   </h1>
 
   <p class="govuk-body-l">
-    <%= t '.lead_para', trade_country_name: 'UK' %>
+    <%= t '.lead_para', trade_country_name: form.object.exporting? ? 'UK' : form.object.trade_country_name %>
```
willfish: Could we consolidate this into a method on the object (possible the base step)?

willfish: That'll be easier to validate than a view spec

javier-transform: Sure thing

PR #1013, opened 2022-09-02
HOTT-1902 Add box "import trade summary" on RoR tab.

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-1902

### What?
Add trade summary duty box on the Rule of Origin tab.

### Why?

To improve UserExperience.


app/models/import_trade_summary.rb
```diff
@@ -0,0 +1,12 @@
+require 'api_entity'
+
+class ImportTradeSummary
+  include ApiEntity
+
+  collection_path '/import_trade_summary'
```
willfish: It's actually misdirecting to have this/you don't need it. This is used to construct the path to pull out the resource (e.g `/geographical_areas`)

app/views/rules_of_origin/_import_trade_summary.html.erb
```diff
@@ -0,0 +1,30 @@
...
+
+        <% if summary.preferential_quota_duty.present? %>
+          <li class="preferential-quota-duty">
+            Preferential quota duty =
+            <strong>
+              <%= summary.preferential_quota_duty.html_safe %>
+            </strong>
+          </li>
+        <% end %>
+    </ul>
```
willfish, line 35: Do you need another one here to conditionally present the an li when no quotas/pref tariffs apply?

```ruby
        <% if summary.no_preferential_duties? %>
          <li class="preferential-quota-duty">
             There is no preferential tariff duty or quota available for this commodity
          </li>
        <% end %>
```

app/models/import_trade_summary.rb
```diff
@@ -0,0 +1,12 @@
...
+
+  attr_accessor :basic_third_country_duty,
+                :preferential_tariff_duty,
+                :preferential_quota_duty
+
```
willfish, line 9:
```ruby

def no_preferential_duties?
  !(preferential_tariff_duty && preferential_quota_duty)  
end
```

spec/views/rules_of_origin/_tab.html.erb_spec.rb
```diff
@@ -97,4 +105,30 @@
       expect(rendered_page).not_to have_css '.rules-of-origin-fta'
     end
   end
+
+  describe 'import trade summary duty box' do
```
willfish, line 107: I don't think you need this describe and actually it doesn't follow better specs

https://www.betterspecs.org/#describe

spec/views/rules_of_origin/_tab.html.erb_spec.rb
```diff
@@ -97,4 +105,30 @@
...
+      it { expect(rendered_page).not_to have_css '.preferential-tariff-duty' }
+    end
+
+    context 'when commodity has preferential quota duty' do
+      let(:preferential_quota_duty) { '1.00 %' }
+
+      it { expect(rendered_page).to have_css '.preferential-quota-duty' }
+    end
+
+    context 'when commodity has NOT preferential quota duty' do
```
willfish:
```suggestion
    context 'when commodity does NOT have a preferential quota duty' do
```

app/models/import_trade_summary.rb
alexdesi: 👍 I agree, I added "just in case", but we don't need it.

app/views/rules_of_origin/_import_trade_summary.html.erb
alexdesi: good spot, thanks!

app/models/import_trade_summary.rb
alexdesi: nice one,
I guess the logic should be
`!(preferential_tariff_duty || preferential_quota_duty) `
or (same logic):
`preferential_tariff_duty.nil? && preferential_quota_duty.nil?`

spec/views/rules_of_origin/_tab.html.erb_spec.rb
alexdesi: In general, I agree with using the method in the `describe`, anyway in this case we refer to a part of UI.
We have other examples in our codebase:

```
describe 'intro paragraph' do
...
describe 'goods nomenclature row' do
...
describe 'supplementary unit row' do
...
....
```

willfish: Well, it also doesn't follow the example of the file its changing but fine :)

willfish (approved):
LGTM

PR #1014, opened 2022-09-05
HOTT-1932: Adds business logic methods for import controls

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1932

### What?

I have added/removed/altered:

- [x] Added business logic methods for import control behaviour
- [x] Added view logic to cater changes

### Why?

I am doing this because:

- This is required as part of UI changes to the commodities change

### DIFF
--- a/app/helpers/declarable_helper.rb
+++ b/app/helpers/declarable_helper.rb
@@ -20,7 +20,7 @@ module DeclarableHelper
 
     stw_link = "#{TradeTariffFrontend.single_trade_window_url}?#{CGI.unescape(stw_options.to_query)}"
     link_to(
-      "check how to #{anchor} #{declarable_type} #{declarable.code} from #{geographical_area&.description}.",
+      "Check how to #{anchor} #{declarable_type} #{declarable.code} from #{geographical_area&.description} (opens in a new tab).",
       stw_link,
       target: '_blank',
       class: 'govuk-link',
@@ -42,5 +42,25 @@ module DeclarableHelper
     else
       commodity_url(declarable, format: :json)
     end
+  end
+
+  def one_or_more_prohibitive_measure?
+    return false if declarable.import_measures.blank?
+
+    declarable.import_measures.each do |measures|
+      return true if measures.prohibitive?
+    end
+
+    false
+  end
+
+  def one_or_more_conditionally_prohibitive_measure?
+    return false if declarable.import_measures.blank? || one_or_more_prohibitive_measure?
+
+    declarable.import_measures.each do |measures|
+      return true if measures.conditionally_prohibitive?
+    end
+
+    false
   end
 end
--- a/app/helpers/declarable_helper.rb
+++ b/app/helpers/declarable_helper.rb
@@ -44,8 +44,8 @@ module DeclarableHelper
     end
   end
 
-  def one_or_more_prohibitive_measure?
-    return false if declarable.import_measures.blank?
+  def one_or_more_prohibitive_measure?(declarable)
+    return false if declarable&.import_measures.blank?
 
     declarable.import_measures.each do |measures|
       return true if measures.prohibitive?
@@ -54,8 +54,8 @@ module DeclarableHelper
     false
   end
 
-  def one_or_more_conditionally_prohibitive_measure?
-    return false if declarable.import_measures.blank? || one_or_more_prohibitive_measure?
+  def one_or_more_conditionally_prohibitive_measure?(declarable)
+    return false if one_or_more_prohibitive_measure?(declarable)
 
     declarable.import_measures.each do |measures|
       return true if measures.conditionally_prohibitive?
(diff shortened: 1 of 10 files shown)


app/helpers/declarable_helper.rb
```diff
@@ -43,4 +43,24 @@ def declarable_url_json(declarable)
       commodity_url(declarable, format: :json)
     end
   end
+
+  def one_or_more_prohibitive_measure?(declarable)
+    return false if declarable&.import_measures.blank?
+
+    declarable.import_measures.each do |measures|
+      return true if measures.prohibitive?
+    end
```
alexdesi: A different approach that does not use loops:
```
def one_or_more_prohibitive_measure?(declarable)
  declarable
    .import_measures.select{|m| m.prohibitive?}
    .any?
end
```

app/helpers/declarable_helper.rb
```diff
@@ -43,4 +43,24 @@ def declarable_url_json(declarable)
       commodity_url(declarable, format: :json)
     end
   end
+
+  def one_or_more_prohibitive_measure?(declarable)
+    return false if declarable&.import_measures.blank?
```
alexdesi: maybe you don't need this, I've added a potential different approach

app/helpers/declarable_helper.rb
```diff
@@ -43,4 +43,24 @@ def declarable_url_json(declarable)
...
+
+  def one_or_more_conditionally_prohibitive_measure?(declarable)
+    return false if one_or_more_prohibitive_measure?(declarable)
+
+    declarable.import_measures.each do |measures|
+      return true if measures.conditionally_prohibitive?
+    end
+
+    false
+  end
```
alexdesi: we could simplify this too

app/views/shared/_stw_link.html.erb
```diff
@@ -1,14 +1,21 @@
...
     <% if search.country.present? %>
-      <p class= "govuk-body">
-        <strong>Check how to import or export goods</strong>
-      </p>
-
-      <p class= "govuk-body">
-        Now you have identified your commodity code, you can <%= declarable_stw_link(declarable, search, anchor) %>
-      </p>
+      <% if one_or_more_prohibitive_measure?(declarable) %>
+        <p> The import of commodity <%= declarable.code %> from <%= trading_partner_country_description(search.country) %> is prohibited  </p>
```
alexdesi, line 4: does this paragraph need the `class= "govuk-body"`  too? 🤔

app/helpers/declarable_helper.rb
```diff
@@ -43,4 +43,24 @@ def declarable_url_json(declarable)
       commodity_url(declarable, format: :json)
     end
   end
+
+  def one_or_more_prohibitive_measure?(declarable)
```
willfish: So we're a bit greedy on the declarable as a dependency. I'm wondering if it would be cleaner to move these questions to the declarable concern?

```ruby
module Declarable
...

def one_or_more_prohibitive_measures?
  return false if import_measures.blank?
  return true if import_measures.any?(&:prohibitive?)

  false
end

def one_or_more_conditionally_prohibitive_measures?
  return false if import_measures.blank?
  return false if one_or_more_prohibitive_measures?
  return true if import_measures.any?(&:conditionally_prohibitive?)

  false
end

end
```

willfish: Then you can just ask the declarable

willfish:
I think we should add some unit tests of the logic

app/models/concerns/declarable.rb
```diff
@@ -37,6 +37,14 @@ module Declarable
     delegate :code, :short_code, to: :chapter, prefix: true
   end
 
+  def one_or_more_prohibitive_measures?
+    import_measures.present? && import_measures.any?(&:prohibitive?)
```
willfish:
```suggestion
    import_measures.any?(&:prohibitive?)
```
I _think_ `any` will return false when an array of import_measures is empty?

spec/models/measure_spec.rb
```diff
@@ -238,4 +238,34 @@
       end
     end
   end
+
+  describe '#prohibitive?' do
+    context 'with prohibitive measures' do
+      subject(:measure) { build(:measure, :prohibitive) }
+
+      it 'returns true' do
+        expect(measure.prohibitive?).to eq(true)
```
willfish:
```suggestion
        is_expected.to be_prohibitive
```

spec/models/measure_spec.rb
```diff
@@ -238,4 +238,34 @@
...
+      end
+    end
+
+    context 'with conditionally prohibitive measures' do
+      subject(:measure) { build(:measure, :prohibitive, :with_additional_code) }
+
+      it 'returns true' do
+        expect(measure.conditionally_prohibitive?).to eq(true)
```
willfish:
```suggestion
        is_expected.to be_conditionally_prohibitive
```

willfish:
LGTM (I can't approve as I opened the first commit of the PR)

javier-transform (approved):
Approved by mutual consent 👍

PR #988, opened 2022-08-23
HOTT-1900 Two cols for the Rule of Origin tab + other improvements.

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-1900

### What?
- [x] Make RoR tab two columns
- [x] remove section about "Product-specific rules"
- [x] Various wording improvements
- [ ] add section "How rules of origin could affect the import duty payable" (coming soo on HOTT-1902)

PR #1024, opened 2022-09-07
HOTT-1944 revert to tabular roo page

jebw:
### Jira link

HOTT-1944

### What?

I have added/removed/altered:

- [x] Reverted removal of existing RoO table which occurred in #988 
- [x] Conditionalised display of RoO tab to use the new tab only when the feature flag is enabled

### Why?

I am doing this because:

- We do not currently display rules of origin data when the feature flag is removed

### Have you? (optional)

- Reverts changes to RoO tab

### DIFF
--- a/app/views/rules_of_origin/_introductory_notes.html.erb
+++ b/app/views/rules_of_origin/_introductory_notes.html.erb
@@ -1,0 +1,58 @@
+<details class="govuk-details" data-module="govuk-details">
+  <summary class="govuk-details__summary">
+    <span class="govuk-details__summary-text">How to read rules of origin</span>
+  </summary>
+
+  <div class="govuk-details__text">
+    <p>The rules of origin table contains:</p>
+
+    <ul class="govuk-list--bullet">
+      <li>
+        the product classification heading ('HS heading') or chapter number
+        (column 1)
+      </li>
+
+      <li>
+        a description of the product, or the headings / subheadings covered by
+        the rules (column 2)
+      </li>
+
+      <li>
+        a description of how non-originating materials must be worked or
+        processed to get country of origin status (column 3)
+      </li>
+    </ul>
+
+    <p>For each row:</p>
+
+    <ul class="govuk-list--bullet">
+      <li>
+        check column 1 to find the product classification heading or chapter
+        number for the product being exported
+      </li>
+
+      <li>check column 2 to find the description that applies to the product</li>
+
+      <li>
+        check column 3 to see how the product must
+        be worked or processed to gain country of origin status
+      </li>
+    </ul>
+
+    <p>
+      Unless preceded by ‘ex’, the rules in column 3 apply to all products that
+      come under the headings and chapters in column 1.
+    </p>
+
+    <p>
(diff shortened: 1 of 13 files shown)


app/views/rules_of_origin/_introductory_notes.html.erb
```diff
@@ -0,0 +1,58 @@
...
+
+    <p>
+      If the heading in column 1 starts with ‘ex’, the rules in column 3
+      will only apply to the product as described in column 2.
+    </p>
+
+    <h2 class="govuk-heading-m">Introductory notes</h2>
+
+    <div class="tariff-markdown">
+      <%= govspeak introductory_notes if introductory_notes.present? %>
```
alexdesi, line 55: should the if stay outside the div?

```
<% if introductory_notes.present? %>
    <div class="tariff-markdown">
      <%= govspeak introductory_notes  %>
    </div>
<% end %>
```

app/views/rules_of_origin/_proofs.html.erb
```diff
@@ -0,0 +1,47 @@
...
+    <% if proofs.many? %>
+      <p>
+        A claim can be made if the importer has one of the following proofs of
+        origin:
+      </p>
+    <% else %>
+      <p>
+        A claim can be made if the importer has the following proof of origin:
+      </p>
+    <% end %>
```
alexdesi, line 29: maybe, to make it clear what's the difference between the two strings:

```
 <p>
   A claim can be made if the importer has
   
 <% if proofs.many? %>
         one of the following proofs of origin:
    <% else %>
        has the following proof of origin:
    <% end %>
</p>
```

app/views/rules_of_origin/_rules_table.html.erb
```diff
@@ -0,0 +1,45 @@
...
+  of them.
+</p>
+
+<table class="govuk-table govuk-table--responsive commodity-rules-of-origin">
+  <thead class="govuk-table__head">
+    <tr class="govuk-table__row">
+      <th class="govuk-table__header govuk-!-width-one-quarter" scope="col">
+        Heading
+      </th>
+
```
alexdesi, line 18: I guess we don't need this kind of extra empty line in this case (the separation is already clear)

spec/views/measures/_measures.html.erb_spec.rb
```diff
@@ -101,8 +120,17 @@
...
+        it_behaves_like 'measures with rules of origin tab'
+        it_behaves_like 'roo_wizard tab'
+        it { is_expected.to have_css '#rules-of-origin h2', text: 'rules of origin for trading' }
+      end
+
+      context 'without roo_wizard feature flag' do
+        before { allow(TradeTariffFrontend).to receive(:roo_wizard?).and_return false }
+
+        it_behaves_like 'legacy roo tab'
+      end
```
alexdesi, line 133: testing again this sounds a bit redundant, but I am not sure 🤔

alexdesi (approved):
It looks good!
added couple of minor comments

PR #1034, opened 2022-09-13
HOTT-1617: Add fallback to commodity when the subheading is not found.

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-1617

### What?
Users should be navigated to the commodity page/commodity history page 
instead of being redirected to the find commodity page.

### Why?
To improve User experience.

# Note:
After a couple of different approaches, which imply changes on the BE too,
I think this solution:
- improves navigation, falling back to the commodity in case the sub-commodity is not found
- very low impact on the code base
- no need to change BE


willfish:
Killing this as we ended up supporting validity periods for subheadings

PR #1036, opened 2022-09-14
HOTT-1973 Remove the duplication logic regarding retrying the Api calls ("retries")

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-1973

### What?
There is a lot of duplication relevant to the retries to the API calls,
the logic is repeated in 5+, this is risky, and a mistake could bring on infinite API calls to the backend.

### DIFF
--- a/app/services/beta/search/perform_search_service.rb
+++ b/app/services/beta/search/perform_search_service.rb
@@ -1,15 +1,18 @@
 require 'tariff_jsonapi_parser'
 require 'errors'
+require 'retriable'
 
 module Beta
   module Search
     class PerformSearchService
+      include Retriable
+
       def initialize(query, filters = {})
         @query = query
         @filters = filters
       end
 
-      def call()
+      def call
         retries(Faraday::Error, UnparseableResponseError) do
           path = "/api/beta/search?#{query_params}"
           response = api.get(path)
@@ -44,18 +47,6 @@ module Beta
 
       def api
         TradeTariffFrontend::ServiceChooser.api_client
-      end
-
-      def retries(*rescue_errors)
-        retries ||= 0
-        yield if block_given?
-      rescue *rescue_errors
-        if retries < Rails.configuration.x.http.max_retry
-          retries += 1
-          retry
-        else
-          raise
-        end
       end
     end
   end
--- a/app/services/beta/search/perform_search_service.rb
+++ b/app/services/beta/search/perform_search_service.rb
@@ -9,19 +9,13 @@ module Beta
         @filters = filters
       end
 
-      def call
-        retries ||= 0
-        path = "/api/beta/search?#{query_params}"
-        response = api.get(path)
-        parsed = parse_jsonapi(response)
+      def call()
+        retries(Faraday::Error, UnparseableResponseError) do
+          path = "/api/beta/search?#{query_params}"
+          response = api.get(path)
+          parsed = parse_jsonapi(response)
 
-        Beta::Search::SearchResult.new(parsed)
-      rescue Faraday::Error, UnparseableResponseError
-        if retries < Rails.configuration.x.http.max_retry
-          retries += 1
-          retry
-        else
-          raise
+          Beta::Search::SearchResult.new(parsed)
         end
       end
 
@@ -50,6 +44,18 @@ module Beta
 
       def api
         TradeTariffFrontend::ServiceChooser.api_client
+      end
+
+      def retries(*rescue_errors)
+        retries ||= 0
+        yield if block_given?
+      rescue *rescue_errors
+        if retries < Rails.configuration.x.http.max_retry
+          retries += 1
+          retry
+        else
+          raise
+        end
       end
     end
   end
--- a/app/services/beta/search/perform_search_service.rb
+++ b/app/services/beta/search/perform_search_service.rb
@@ -1,6 +1,5 @@
 require 'tariff_jsonapi_parser'
 require 'errors'
-require 'retriable'
 
 module Beta
   module Search
--- a/app/services/beta/search/perform_search_service.rb
+++ b/app/services/beta/search/perform_search_service.rb
@@ -13,7 +13,7 @@ module Beta
       end
 
       def call
-        retries(Faraday::Error, UnparseableResponseError) do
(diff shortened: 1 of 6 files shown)


spec/lib/retriable_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+        expect(cat).to have_received(:maow).exactly(6).time
+      end
+    end
+  end
+end
+
+require 'retriable'
+class RetryTest
+  include Retriable
+end
```
alexdesi: RetryTest helps us to test `Retriable`

spec/lib/retriable_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+        allow(cat).to receive(:maow)
+
+        expect {
+          RetryTest.new.retries(StandardError) do
+            cat.maow
+            raise('error!') # <- StandardError
+          end
+        }.to raise_error('error!')
+
+        expect(cat).to have_received(:maow).exactly(6).time
```
alexdesi: These 2 expectations are a way to test the right behaviour.
I could not use just one expectation because unfortunately rspec **has not** something like:

`has_raised('error!').exactly(6).time`

alexdesi: Also, I think trying 6 times is too much,
if a call does not succeed after 2 o 3 times, it is unlikely it succeeds the 6th time.

I'd reduce it to 3 if you agree.

spec/lib/retriable_spec.rb
jebw: RSpec recommends anonymous classes for that purpose to avoid polluting the global namespace so instead of 

```
class RetryTest
  include Retriable
end
```

They recommend 

```
let :retry_test do
  Class.new do
    include Retriable
  end
end
```

jebw: And instead of writing `RetryTest.new` you'd write `retry_test.new`

lib/retriable.rb
```diff
@@ -0,0 +1,13 @@
+module Retriable
+  def retries(*rescue_errors)
+    retries ||= 0
```
jebw, line 3: shadowing the method name is a little confusing - maybe call the variable `retry_count` instead?

spec/lib/retriable_spec.rb
```diff
@@ -0,0 +1,37 @@
+require 'spec_helper'
+
+RSpec.describe Retriable do
+  subject(:call) {}
+
+  describe('#retries') do
+    let(:cat) { double }
+
+    it 'makes the call' do
+      allow(cat).to receive(:maow)
```
willfish: TIL there are different spellings for cat vocalisations :tada: https://en.wikipedia.org/wiki/Meow

willfish: How do you feel about making our cat a kitten and having vocalise as `mew` :laughing:

willfish: *not a real suggestion lol

spec/lib/retriable_spec.rb
willfish: Wondering about this being in the main scope rather than a subject. I'm guessing you could do:

```ruby
subject(:retriable) do
  Class.new do
    include Retriable
  end
end
```

lib/retriable.rb
```diff
@@ -0,0 +1,13 @@
...
+    yield if block_given?
+  rescue *rescue_errors, StandardError
```
jebw: This changes several locations which previously only rescued Faraday::Error and UnparseableResponseError to _also_ rescue from `StandardError` which is much broader in scope and generally best avoided

I think this needs to either explicitly require a list of exceptions, and check the list isn't empty or just default the list to `Faraday::Error` when its empty - so either

```
rescue_errors = [Faraday::Error] if rescue_errors.empty?
retries ||= 0
yield if block_given?
rescue *rescue_errors
```

app/services/beta/search/perform_search_service.rb
```diff
@@ -1,27 +1,24 @@
 require 'tariff_jsonapi_parser'
 require 'errors'
+require 'retriable'
```
willfish: We use zeitwerk which should mean that you don't need these requires

willfish (approved):
LGTM. We'll want to validate that we're still able to request changes, perform beta and old searches and request commodities/headings, etc

spec/lib/retriable_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    context 'when unsuccessfull calls' do
+      it 'retries multiple times' do
+        allow(cat).to receive(:maow)
+
+        expect {
+          RetryTest.new.retries do
+            cat.maow
+            raise('error!') # <- StandardError
```
jebw: Maybe raise a FaradayError instead - see

https://github.com/trade-tariff/trade-tariff-frontend/blob/main/spec/controllers/healthcheck_controller_spec.rb#L13

app/services/beta/search/perform_search_service.rb
jebw: lib isn't autoloaded

jebw (changes requested):
Looks good, I'm requesting changes because we shouldn't be rescuing from StandardError in ApiEntity

lib/retriable.rb
alexdesi: 👍 good point, @willfish  thinks the same, lets's go with `with_retries` for the method's name which sounds more rubyist.

spec/lib/retriable_spec.rb
alexdesi: OMG you are right! Maow is actually a Canadian rock band: https://en.wikipedia.org/wiki/Maow

lib/retriable.rb
willfish: AFAIK, its common practice to rescue StandardError. I think you might be mixing it with the principle of not rescuing Exception?

alexdesi: Since the production, code uses `StandardError`,  and this is just a refactor,
and this retries is general purpose,  I'm gonna use the same `StandardError`.

alexdesi:
Thanks @jebw and @willfish for your comments!

PR #1037, opened 2022-09-14
HOTT-1627: Use threshold_unit_type to display measure_condition requi…

maverickvi:
…rements

Important: This PR should be merged after the backend PR because these changes can not work without the backend changes: https://github.com/trade-tariff/trade-tariff-backend/pull/776

### Jira link

[HOTT-<1627>](https://transformuk.atlassian.net/browse/HOTT-1627)

### What?

I have added/removed/altered:

- [ ] Removed logic from frontend

### Why?

I am doing this because:

- Backend now provides threshold_unit_type string, which contains the information we need to display measure condition requirements

Screenshot 2022-09-14 at 16 43 10

### DIFF
--- a/app/helpers/measures_helper.rb
+++ b/app/helpers/measures_helper.rb
@@ -68,13 +68,14 @@ module MeasuresHelper
   def format_measure_condition_requirement(condition)
     case condition.measure_condition_class
     when 'threshold'
-      if condition.is_eps_condition?
+      case condition.threshold_unit_type
+      when 'eps'
         safe_join ['The price of your goods is greater than or equal to', condition.requirement], ' '
-      elsif condition.is_price_condition?
+      when 'price'
         safe_join ['The price of your goods does not exceed', condition.requirement], ' '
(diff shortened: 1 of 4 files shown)


jebw (changes requested):
LGTM - really clean :raised_hands: 

Only change I'd request is can the PR description be updated to mention this has a hard dependency on the backend PR being released (not merged) first

Probably makes sense to merge this after the corresponding backend is released to production anyway, since that will avoid any issues arising from the frontend being released prior to the backend change

jebw (approved):

maverickvi:
> LGTM - really clean 🙌
> 
> Only change I'd request is can the PR description be updated to mention this has a hard dependency on the backend PR being released (not merged) first
> 
> Probably makes sense to merge this after the corresponding backend is released to production anyway, since that will avoid any issues arising from the frontend being released prior to the backend change

I have updated the description so we don't accidentally merge this before the backend changes.

PR #1038, opened 2022-09-15
HOTT-1935: Fix country name in RoO journeys

maverickvi:
### Jira link

[HOTT-<1935>](https://transformuk.atlassian.net/browse/HOTT-1935)

### What?

I have added/removed/altered:

- [x] Fixed the country name in RoO journeys and refactored existing code to use the same translation

### Why?

I am doing this because:

-  We were displaying the wrong country name in RoO journeys

Screenshot 2022-09-15 at 15 16 35
Screenshot 2022-09-15 at 15 16 17
Screenshot 2022-09-15 at 15 16 10


jebw (approved):
LGTM :raised_hands:

PR #1040, opened 2022-09-16
HOTT-1974: Fixes conditionally prohibitive bug

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1974

### What?

I have added/removed/altered:

- [x] Rejigged ordering of logic in the stw link view

### Why?

I am doing this because:

- We're not showing conditionally prohibitive messages currently because we never ask the question

### DIFF
--- a/spec/factories/commodity_factory.rb
+++ b/spec/factories/commodity_factory.rb
@@ -57,5 +57,21 @@ FactoryBot.define do
         attributes_for(:import_trade_summary)
       end
     end
+
+    trait :with_conditionally_prohibitive_measures do
+      import_measures do
+        [
+          attributes_for(:measure, :conditionally_prohibitive),
+        ]
+      end
+    end
+
+    trait :with_prohibitive_measures do
+      import_measures do
+        [
+          attributes_for(:measure, :prohibitive),
+        ]
+      end
+    end
   end
 end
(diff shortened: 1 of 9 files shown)


spec/helpers/declarable_helper_spec.rb
```diff
@@ -1,4 +1,38 @@
...
+      let(:search) { build(:search, :with_country) }
+
+      it { is_expected.to eq("<p>\n  The import of commodity 0101300020 from Italy may be prohibited, depending on the additional code used.\n</p>") }
+    end
+
+    context 'when prohibitive' do
+      let(:declarable) { build(:commodity, :with_prohibitive_measures, goods_nomenclature_item_id: '0101300020') }
+      let(:search) { build(:search, :with_country) }
+
+      it { is_expected.to eq("<p>\n  The import of commodity 0101300020 from Italy is prohibited.\n</p>") }
```
jebw, line 32: Rather then trying to match html by exact characters, `have_css` is more flexible and will work in this context - eg

`it { is_expected.to have_css 'p', text: /import.*\d{10}.*Italy/`

spec/factories/commodity_factory.rb
```diff
@@ -57,5 +57,21 @@
...
+        [
+          attributes_for(:measure, :conditionally_prohibitive),
+        ]
+      end
+    end
+
+    trait :with_prohibitive_measures do
+      import_measures do
+        [
+          attributes_for(:measure, :prohibitive),
```
jebw, line 72: `attributes_for_list(:measure, 1, :prohibitive)` will return an array of 1 if that helps?

jebw (approved):
Looks good, some suggestions around the test code

Regarding the html in the translation file - part of me wonders if this needs to be in the translation file? The helper method could just render different partials?

PR #1041, opened 2022-09-16
HOTT-1975: Fix page HTML

maverickvi:
### Jira link

[HOTT-<1975>](https://transformuk.atlassian.net/browse/HOTT-1975)

### What?

I have added/removed/altered:

- [ ] Fixed page HTML

### Why?

I am doing this because:

- We had unwanted p tag in the view

Screenshot 2022-09-16 at 10 30 30


alexdesi (approved):
Easy peasy  ✅

PR #1042, opened 2022-09-16
HOTT-1904: New improvements on RoR tab

alexdesi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1904

### What?
-[x] Add two new info boxes and hide 
-[x] Check RoO button if there are no preferential duties.

### Why?
We want to tell people that when they do not need to bother with a preferential tariff
AND
there is no preference and no quota for the selected country & comm code combination, there is no value in taking users through the pain of the rules of origin wizard.


jebw (approved):
LGTM

PR #1043, opened 2022-09-16
HOTT-1932: Fixes prohibitive measures stw link bug

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-1932

### What?

I have added/removed/altered:

- [x] Altered logic of declarable helper to support mixture of measure situations
- [x] Added specs

### Why?

I am doing this because:

- When there are prohibitive and conditionally prohibitive measures we
should always pick the prohibitive message


alexdesi (approved):
it looks good ✅

PR #1044, opened 2022-09-16
HOTT-1952 Simplify exception handling

jebw:
### Jira link

HOTT-1952

### What?

I have added/removed/altered:

- [x] Added spec for errors pages to ensure they can render and render the correct content
- [x] Added specs to ensure raised exceptions in controller code are rescued by `ActionDispatch::ShowExceptions` middleware correctly
- [x] Removed custom exception rescue behaviour and use the built in rails exception handling
- [x] Configured the rails exception handling to behave correctly for the additional exceptions we were previously overriding
- [x] Enabled error reporting to sentry for exceptions rescued by the 500 and 404 pages 
- [x] Added a shared context to enable rescuing exceptions for a single spec

### Why?

I am doing this because:

- We previously had to make updates to the errors in two places but this wasn't obvious and often only one would happen
- Nothing was checking the error pages themselves rendered correctly
- Developers would not see the default rails exception pages, which would obscure the source of errors - most notable a missing or misnamed partial would result in the branded Page Not Found screen - confusing for new developers.
- We were previously only reporting really low level errors to sentry

- Makes changes to our handling and reporting of exceptions

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -17,18 +17,6 @@ class ApplicationController < ActionController::Base
 
   layout :set_layout
 
-  rescue_from Faraday::ServerError, Errno::ECONNREFUSED do
-    request.format = :html
-    render_500
-  end
-
-  rescue_from(Faraday::ResourceNotFound, WizardSteps::UnknownStep,
-              ActionView::MissingTemplate, ActionController::UnknownFormat,
-              AbstractController::ActionNotFound, URI::InvalidURIError) do |_e|
-    request.format = :html
-    render_404
-  end
-
   def url_options
     return super unless search_invoked?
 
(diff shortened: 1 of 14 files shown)


app/controllers/application_controller.rb
```diff
@@ -52,22 +40,6 @@ def url_options
...
-  end
-
-  def render_404
-    disable_search_form
-    render template: 'errors/not_found',
-           status: :not_found,
-           formats: :html
-    false
-  end
-
```
alexdesi, line 70: nice cleaning! 🧹 🧹 🧹

spec/controllers/chapters/changes_controller_spec.rb
```diff
@@ -28,10 +28,10 @@
   end
 
   describe 'chapter is not valid at given date', vcr: { cassette_name: 'chapters_changes#index_0100000000_1970-01-01' } do
-    before do
+    let :request_page do
       get :index, params: { chapter_id: chapter.short_code, as_of: Date.new(1970, 1, 1) }, format: :atom
     end
 
-    it { is_expected.to respond_with(:not_found) }
```
alexdesi, line 35: I think it does kind of make sense to have an expectation on the response in the controller, it's like a small integration test, that tests the ExceptionWrapper too.

alexdesi (approved):
✅ nice cleaning up!
I've just added a comment on a controller spec.

spec/controllers/chapters/changes_controller_spec.rb
jebw: Yeh - I dropped this because it wont work (the request object gets reused across the rest of the specs, which means it wont pass without turning off 'normal' exception handling in specs for the whole test environment :disappointed: I'd had an unsuccessful attempt at enabling that just for these specs, I may retry

PR #1045, opened 2022-09-16
HOTT-2004 Fix and specs for search banner on errors

jebw:
### Jira link

HOTT-2004

### What?

I have added/removed/altered:

- [x] Fixed bug when search banner is enabled

### Why?

I am doing this because:

- the `@search` variable is not assigned on error pages and so the search should not be shown

- Low, affects currently disable feature

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -63,7 +63,7 @@
       <%= render 'shared/govuk_logo' %>
       <%= yield :proposition_header %>
     </div>
-    <%= render 'shared/search_banner' if TradeTariffFrontend.search_banner? %>
+    <%= render 'shared/search_banner' if TradeTariffFrontend.search_banner? && @search %>
     <%= render 'news_items/header_banner', news_item: NewsItem.cached_latest_banner %>
   </header>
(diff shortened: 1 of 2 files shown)


app/views/layouts/application.html.erb
```diff
@@ -63,7 +63,7 @@
       <%= render 'shared/govuk_logo' %>
       <%= yield :proposition_header %>
     </div>
-    <%= render 'shared/search_banner' if TradeTariffFrontend.search_banner? %>
+    <%= render 'shared/search_banner' if TradeTariffFrontend.search_banner? && @search %>
```
alexdesi, line 66: If we want to be safe, I guess we should use:
`defined?(@search) && @search.present?`

Should not a problem with the current Ruby MRI version 
`@search` returns nill in case of:
- not declared
- it contains nil

but things could change

alexdesi: Ok, ignore my comment 😸 
we actually commonly use instance variables, which might not be even declared, in the views.

alexdesi (approved):
✅ It looks good!

PR #1046, opened 2022-09-16
HOTT-1958 Dont redirect maintenance page

jebw:
### Jira link

HOTT-1958

### What?

I have added/removed/altered:

- [x] Use exception handling to show maintenance page
- [x] Bump priority of before action checking for maintenance mode
- [x] Support bypass env var to allow testing whilst still in maintenance mode

### Why?

I am doing this because:

- If we redirect the user to `/503` then even after we remove maintenance mode, if they refresh their page they will still be show the 503 page leaving them unaware the site is back available
- If the backend is uncontactable for any reason, then the user would see the 500 page instead of the 503 page even when the site has been placed into maintenance mode
- We need to be able to test if the site is working correctly whilst still in maintance mode for the general public

- Changes how our maintenance page works - which needs to be reliable when we are making the service unavailable for any reason

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,14 +5,14 @@ class ApplicationController < ActionController::Base
   include TradeTariffFrontend::ViewContext::Controller
   include ApplicationHelper
 
+  before_action :maintenance_mode_if_active
+
   around_action :set_locale
 
   before_action :set_cache
   before_action :set_last_updated
   before_action :set_path_info
-
   before_action :set_search
-  before_action :maintenance_mode_if_active
   before_action :bots_no_index_if_historical
 
   layout :set_layout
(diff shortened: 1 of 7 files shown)


app/controllers/application_controller.rb
```diff
@@ -112,8 +112,11 @@ def meursing_lookup_result
   protected
 
   def maintenance_mode_if_active
-    if ENV['MAINTENANCE'].present? && action_name != 'maintenance'
-      redirect_to '/503'
+    if ENV['MAINTENANCE'].present? &&
+        (ENV['MAINTENANCE_BYPASS'].blank? ||
+         params[:maintenance_bypass].blank? ||
+         ENV['MAINTENANCE_BYPASS'] != params[:maintenance_bypass])
```
alexdesi: it's pretty difficult to follow this logical expression,

is it the same as this?
```
        !(ENV['MAINTENANCE_BYPASS'].present? && params[:maintenance_bypass].present? &&
         ENV['MAINTENANCE_BYPASS']  == params[:maintenance_bypass])
```

not sure, we should check that maintenance_bypass is also True, and not just present 🤔

alexdesi (approved):
good one! thanks for the changes ✅

PR #1062, opened 2022-09-23
HOTT-2018 change presentation of blocking periods

jebw:
### Jira link

HOTT-2018

### What?

I have added/removed/altered:

- [x] Show a combined row 'n/a' row when there are neither blocking nor suspension periods
- [x] Otherwise show only the applicable blocking suspension periods rows
- [x] Added a generalised helper for display of date ranges to replace the quota period specific ones
- [x] Added a view spec to test the quota definition overlay with different data scenarios

### Why?

I am doing this because:

- It simplifies the page for users
- We no longer required the suspension and blocking period specific helpers
- The view spec ensures the page renders with different data scenarios

### Have you? (optional)

### Screenshots

Screenshot from 2022-09-23 11-33-24
Screenshot from 2022-09-23 11-32-59
Screenshot from 2022-09-23 11-32-36
Screenshot from 2022-09-23 10-45-09

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -130,10 +130,7 @@ module ApplicationHelper
   end
 
   def govuk_date_range(from, to)
-    from = from.to_date if from.respond_to?(:to_date)
-    to = to.to_date if to.respond_to?(:to_date)
-
-    "#{from.to_formatted_s(:long)} to #{to.to_formatted_s(:long)}"
+    "#{from.to_date.to_formatted_s(:long)} to #{to.to_date.to_formatted_s(:long)}"
   end
 
   def paragraph_if_content(content)
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -129,6 +129,13 @@ module ApplicationHelper
     end
   end
 
+  def govuk_date_range(from, to)
+    from = from.to_date if from.respond_to?(:to_date)
+    to = to.to_date if to.respond_to?(:to_date)
+
+    "#{from.to_formatted_s(:long)} to #{to.to_formatted_s(:long)}"
+  end
+
   def paragraph_if_content(content)
     tag.p(content) if content.present?
   end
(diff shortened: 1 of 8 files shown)


app/helpers/application_helper.rb
```diff
@@ -129,6 +129,13 @@ def present_from_to(from, to)
     end
   end
 
+  def govuk_date_range(from, to)
+    from = from.to_date if from.respond_to?(:to_date)
+    to = to.to_date if to.respond_to?(:to_date)
```
alexdesi: Is it possible that  `from`  responds to `to_date` but not `to`? 🤔 
(or vice-versa)

is it acceptable?

if not a guard clause might be a better choice
```
return unless from.respond_to?(:to_date) && and to.respond_to?(:to_date)

"#{from.to_formatted_s(:long)} to #{to.to_formatted_s(:long)}"

```

spec/models/order_number/definition_spec.rb
```diff
@@ -38,4 +38,84 @@
...
+
+    context 'with no start or end dates' do
+      let :definition do
+        build :definition, blocking_period_start_date: nil,
+                           blocking_period_end_date: nil
+      end
+
+      it { is_expected.to be false }
+    end
+  end
```
alexdesi, line 120: very readable tests 👍 
I just noticed that maybe we  could remove some of the tests,
I guess we could tests only 2 conditions, true (when both are true)
and false (when one is false)
instead of
true true
true false
false true
false false

Up to you anyway 😉

alexdesi (approved):
nice one! ✅ 

spec/models/order_number/definition_spec.rb
jebw: I'd disagree with this -> my reasoning being the tests should testing combinations of input conditions not outcomes

There are 4 possible combinations of inputs
There are 2 possible outcomes

If I was writing a test for `String#blank?` - it only has `true` and `false` as outcomes but if I only test it with an empty string I'm still not proving that it also returns `true` for a string consistenting only of whitespace.

An implementation as follows would pass the test suite but actually return `false` for a whitespace string - this would mean the test suite was incomplete?

```ruby
def blank?
  nil? || length.zero?
end
```

PR #1063, opened 2022-09-26
HOTT-2002: Added guides to search results

maverickvi:
### Jira link

[HOTT-<2002>](https://transformuk.atlassian.net/browse/HOTT-2002)

### What?

I have added/removed/altered:

- [ ] Added guides to search results

### Why?

I am doing this because:

- this allows us to display useful information to users alongside search results

### Have you? (optional)

Screenshot 2022-09-26 at 14 08 45

### DIFF
--- a/app/views/beta/search_results/_sidebar.html.erb
+++ b/app/views/beta/search_results/_sidebar.html.erb
@@ -35,6 +35,6 @@
 
     <% end %>
 
-    <%= render "search/guide_section", guide: guide if guide %>
+    <%= render "search/guide_section", guide: guide if guide.present? %>
   </nav>
 </div>
--- a/app/views/beta/search_results/_sidebar.html.erb
+++ b/app/views/beta/search_results/_sidebar.html.erb
@@ -34,5 +34,7 @@
       </div>
 
     <% end %>
+
+    <%= render "search/guide_section", guide: guide if guide %>
   </nav>
 </div>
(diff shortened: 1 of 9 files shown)


willfish (approved):

spec/views/beta/search_results/show.html.erb_spec.rb
```diff
@@ -25,4 +25,20 @@
     it { is_expected.to render_template('beta/search_results/_search_results') }
     it { is_expected.not_to render_template('beta/search_results/_sidebar') }
   end
+
+  context 'when guide exists' do
+    let(:search_result) { build(:search_result) }
+
+    it { is_expected.to render_template('search/_guide_section') }
+
+    context "when guide doesn't exist" do
```
alexdesi: you are putting 
`when guide doesn't exist`
inside
`when guide exists`

app/views/beta/search_results/_sidebar.html.erb
```diff
@@ -34,5 +34,7 @@
       </div>
 
     <% end %>
+
+    <%= render "search/guide_section", guide: guide if guide %>
```
alexdesi: I guess it's better to use
`... if guide.present?` 
to cover also few more cases that are considered empty (like empty string)

spec/views/beta/search_results/show.html.erb_spec.rb
```diff
@@ -25,4 +25,20 @@
...
+
+    it { is_expected.to render_template('search/_guide_section') }
+  end
+
+  context "when guide doesn't exist" do
+    let(:search_result) { build(:search_result) }
+
+    before do
+      search_result.guide = nil
+    end
```
alexdesi, line 40: I guess this should be the same
```
let(:search_result) { build(:search_result, guide:nil ) }

it { is_expected.not_to render_template('search/_guide_section') }

```

and no need to `before` block

alexdesi (approved):
Nice one! ✅

PR #1083, opened 2022-10-10
HOTT-2017 show pending balance

jebw:
### Jira link

HOTT-2017

### What?

I have added/removed/altered:

- [x] Added an API to query the pending balance for a commodity
- [x] Changed the quota partial to conditionally include a table row for the pending balance
- [x] Lazy low the pending balance over XHR from the new API via a Stimulus controller

### Why?

I am doing this because:

- We want to show the pending balance on the quotas overlay
- Eager loading this data via the API is difficult to do efficiently so

### Have you? (optional)

- Should be low risk

### DIFF
--- a/app/services/pending_quota_balance_service.rb
+++ b/app/services/pending_quota_balance_service.rb
@@ -1,0 +1,49 @@
+class PendingQuotaBalanceService
+  attr_reader :declarable_id, :quota_order_number_id, :chosen_period
+
+  def initialize(declarable_id, quota_order_number_id, chosen_period)
+    @declarable_id = declarable_id
+    @quota_order_number_id = quota_order_number_id
+    @chosen_period = chosen_period
+  end
+
+  def call
+    pending_balance
+  end
+
+private
+
+  def declarable
+    @declarable ||= Commodity.find(declarable_id, as_of: chosen_period) # FIXME
+  end
+
+  def quota_measure
+    declarable.import_measures.find_by_quota_order_number(quota_order_number_id)
+  end
+
+  def definition
+    @definition ||= quota_measure&.order_number&.definition
+  end
+
+  def previous_period_declarable
+    @previous_period_declarable ||= Commodity.find declarable_id, as_of: previous_period # FIXME
+  end
+
+  def previous_quota_measure
+    previous_period_declarable.import_measures.find_by_quota_order_number(quota_order_number_id)
+  end
+
+  def previous_period_definition
+    previous_quota_measure&.order_number&.definition
+  end
+
+  def pending_balance
+    if definition.within_first_twenty_days? && declarable.has_safeguard_measure?
+      previous_period_definition&.balance
+    end
+  end
+
+  def previous_period
+    (definition.validity_start_date - 1.day).to_date
+  end
+end
--- a/app/services/pending_quota_balance_service.rb
+++ b/app/services/pending_quota_balance_service.rb
@@ -2,9 +2,7 @@ class PendingQuotaBalanceService
   attr_reader :declarable_id,
               :quota_order_number_id,
               :chosen_period,
-              :data_model,
-              :declarable,
-              :previous_period_declarable
+              :data_model
 
   def initialize(declarable_id, quota_order_number_id, chosen_period)
     @declarable_id = declarable_id
@@ -19,8 +17,8 @@ class PendingQuotaBalanceService
 
 private
 
-  def fetch_current_declarable
-    @declarable = data_model.find(declarable_id, as_of: chosen_period)
+  def declarable
+    @declarable ||= data_model.find(declarable_id, as_of: chosen_period)
   end
 
   def quota_measure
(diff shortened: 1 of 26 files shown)


app/services/pending_quota_balance_service.rb
```diff
@@ -0,0 +1,57 @@
...
+    previous_period_declarable.import_measures.find_by_quota_order_number(quota_order_number_id)
+  end
+
+  def previous_period_definition
+    previous_quota_measure&.order_number&.definition
+  end
+
+  def pending_balance
+    fetch_current_declarable
+    if definition.within_first_twenty_days? && declarable.has_safeguard_measure?
```
alexdesi, line 45: in theory `definition` could be `nil`, so `within_first_twenty_days?` would raise no method found

app/views/search/quotas/_definition.html.erb
```diff
@@ -1,6 +1,12 @@
 <tr class="govuk-table__row">
   <td class="govuk-table__cell" data-label="Order number">
-    <%= render partial: 'shared/quota_definition', locals: { order_number: definition.order_number, quota_definition: definition } %>
+    <%= render 'shared/quota_definition',
+               order_number: definition.order_number,
+               quota_definition: definition,
+               declarable_id: definition.all_goods_nomenclatures
+                                        .select { |gn| gn.respond_to?(:short_code) }
+                                        .first
+                                        &.short_code %>
```
alexdesi: The determination of `declarable_id` seems to be complex enough to deserve its own method:
```
all_goods_nomenclatures
  .select { |gn| gn.respond_to?(:short_code) }
  .first
  &.short_code
```

app/views/shared/_quota_definition.html.erb
```diff
@@ -45,6 +45,20 @@
               <%= quota_definition.measurement_unit %>
             </td>
           </tr>
+          <% if quota_definition.within_first_twenty_days? && declarable_id %>
+          <tr class="govuk-table__row hidden-pending-balance"
+              data-controller="pending-balance"
+              data-pending-balance-url-value="<%= pending_quota_balance_path declarable_id, quota_definition.order_number.number %>">
```
alexdesi: not important: maybe with parenthesis to make it more readable?
```
<%= pending_quota_balance_path(declarable_id, quota_definition.order_number.number) %>
```

spec/models/heading_spec.rb
```diff
@@ -67,4 +67,22 @@
       it { is_expected.to be_nil }
     end
   end
+
+  describe '#has_safeguarding_measure?' do
+    subject { build :heading, import_measures: [vat, measure] }
+
+    let(:vat) { attributes_for :measure, :vat }
+
+    context 'without safeguarding measures' do
```
alexdesi: is the description correct here? 
the safeguard seems to be present

spec/models/heading_spec.rb
```diff
@@ -67,4 +67,22 @@
...
+      let(:measure) { attributes_for :measure, :safeguard }
+
+      it { is_expected.to be_has_safeguard_measure }
+    end
+
+    context 'with safeguarding measure' do
```
alexdesi: same here (but opposite)

alexdesi (commented):
I've added a few minor comments, but it all looks very good to me.

app/services/pending_quota_balance_service.rb
jebw: Hmm - good point - I think that would occur if an API request was made for an order number which lacked a definition - in theory that feels like bad data but I'll just change it to return nil in that scenario :+1:

app/views/search/quotas/_definition.html.erb
jebw: :+1:

spec/models/heading_spec.rb
jebw: :facepalm:

spec/models/heading_spec.rb
jebw: :facepalm:

app/services/pending_quota_balance_service.rb
```diff
@@ -0,0 +1,57 @@
...
+      fetch_previous_declarable
+      previous_period_definition&.balance
+    end
+  end
```
alexdesi: I've just noticed this:
instead of calling explicitly `fetch_current_declarable`, and relying on the assignment, would it be possible just :

```
  def pending_balance
    if definition.within_first_twenty_days? && current_declarable.has_safeguard_measure?
      fetch_previous_declarable
      previous_period_definition&.balance
    end
  end
```
and

```
  def current_declarable
    @current_declarable = data_model.find(declarable_id, as_of: chosen_period)
  end
```
which also makes  Rubocop happier :)

alexdesi (approved):
Nice one! thanks for the changes ✅ 
I've added the last comment suggesting a tiny refactor for `pending_balance`, it might be worth it in case you got 5 min.

PR #1084, opened 2022-10-10
HOTT-2102: Add webchat sections on various pages.

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-2102

### What?
This PR encompasses all the sub-tickets in HOTT-2102.

Add webchat messages and links on various pages.

### Why?
I am doing this to improve the support provided to the users.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -132,4 +132,19 @@ module ApplicationHelper
   def paragraph_if_content(content)
     tag.p(content) if content.present?
   end
+
+  def webchat_message_type
+    case [controller_name, action_name]
+    when %w[search_results show]
+      :not_found
+    when %w[commodities show],
+         %w[headings show],
+         %w[subheadings show]
+         %w[chapters show],
+         %w[sections show]
+      :footer
+    when %w[pages help]
+      :help
+    end
+  end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -133,18 +133,7 @@ module ApplicationHelper
     tag.p(content) if content.present?
   end
 
-  def webchat_message_type
-    case [controller_name, action_name]
-    when %w[search_results show]
-      :not_found
-    when %w[commodities show],
-         %w[headings show],
-         %w[subheadings show]
-         %w[chapters show],
-         %w[sections show]
-      :footer
-    when %w[pages help]
-      :help
-    end
+  def webchat_visible_in_footer?
+    %w[commodities headings subheadings chapters sections].include?(controller_name)
   end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -132,8 +132,4 @@ module ApplicationHelper
   def paragraph_if_content(content)
     tag.p(content) if content.present?
   end
-
-  def webchat_visible_in_footer?
-    %w[commodities headings subheadings chapters sections].include?(controller_name)
-  end
 end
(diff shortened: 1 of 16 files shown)


alexdesi:
Screenshots:
In the footer of various pages:
Screenshot 2022-10-11 at 16 34 07

---

On the help page:
Screenshot 2022-10-11 at 16 33 05

---

In search results when no entries have been found:
Screenshot 2022-10-11 at 16 32 07

app/views/shared/webchat_message/_footer.html.erb
```diff
@@ -0,0 +1,7 @@
+<% if TradeTariffFrontend.webchat_enabled? %>
+    <div>
+        Use our
+        <%= link_to('Digital Assistant (opens in new tab)', "#{TradeTariffFrontend.webchat_url}", target: '_blank') %>
```
willfish: Small thing, but there's a bit of duplication with this link. Could you maybe add a WebchatHelper and consolidate there?

app/helpers/application_helper.rb
```diff
@@ -132,4 +132,8 @@ def present_from_to(from, to)
   def paragraph_if_content(content)
     tag.p(content) if content.present?
   end
+
+  def webchat_visible_in_footer?
```
willfish: we're missing tests for this chipolata

app/views/shared/webchat_message/_footer.html.erb
```diff
@@ -0,0 +1,7 @@
...
+        to get help on classifying your goods.
+    </div>
+<% end %>
```
willfish: Confused by the missing end of file terminator intermittently. Are you mixing/matching with a configured/non-configured editor?

alexdesi: I am tinkering with VisualStudio these days...
there must be some config to always add the last return :)

app/helpers/application_helper.rb
alexdesi: done 😄

alexdesi: I've added the tests for the webchat partials too

app/views/shared/webchat_message/_footer.html.erb
alexdesi: there you are:

```ruby
module WebchatHelper
  def webchat_link
    link_to('Digital Assistant (opens in new tab)', "#{TradeTariffFrontend.webchat_url}", target: '_blank', rel: 'noopener')
  end

  def webchat_visible_in_footer?
    %w[commodities headings subheadings chapters sections].include?(controller_name)
  end
end
```

willfish (approved):

spec/views/_webchat_message.html.erb_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  shared_examples_for 'a webchat link' do |partial_view|
+    subject(:rendered_page) do
+      render partial_view
+
+      rendered
+    end
+
+    context 'when ENV WEBCHAT_URL is set' do
+      before do
+        ENV['WEBCHAT_URL'] = 'http://webchat_url_test'
```
jebw, line 13: This changes this env var for all subsequent specs as well as this one. Probably not an issue for this particlar spec but this pattern can introduce flaky specs since the randomised ordering of specs will affect whether this breaks other subsequent specs?

A different approach is to use mocking?

```ruby
before do
  allow(ENV).to receive(:[]).and_call_original
  allow(ENV).to receive(:[]).with('WEBCHAT_URL').and_return('http://webchat_test_url')
end
```

jebw: Or even easier is stubbing the ruby code used to access the env var

```ruby
before do
  allow(TradeTariffFrontend).to receive(:webchat_url).and_return('http://webchat_test_url')
end
```

PR #1092, opened 2022-10-13
HOTT-2006 Enabled error reporting from Sentry

jebw:
### Jira link

HOTT-2006

### What?

I have added/removed/altered:

- [x] Report errors when they are rescued by rails' default error handling middleware

### Why?

I am doing this because:

- It can indicate legimate issues which we wish to address

- Changes our error reporting behaviour

### DIFF
--- a/config/initializers/sentry.rb
+++ b/config/initializers/sentry.rb
@@ -1,5 +1,7 @@
 Sentry.init do |config|
-  config.rails.report_rescued_exceptions = false
-
   config.breadcrumbs_logger = [:active_support_logger]
+
+  config.excluded_exceptions += %w[
+    Faraday::ResourceNotFound
+  ]
 end


alexdesi (approved):
✅ Approved but I am wondering if we should not raise exceptions for the 
 WizardSteps::UnknownStep and    TradeTariffFrontend::FeatureUnavailable.

This way our workmates from the future could ask why those exceptions are not reported on Sentry.

jebw:
yep, fair comment - its a cherry pick from a separate PR which adds those in

* WizardSteps::UnknownSteps  is already raised (by the gem), and rescued by ApplicationController but that will be going away in the future
* TradeTariffFrontend::FeatureUnavailable isn't yet used but will be in the branch its cherry picked from

I'll drop these from this PR and add them back into the other PR

PR #1094, opened 2022-10-14
HOTT-2140: Improvement for facet filter tag

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-2140

### What?
Move the Facet filter tag to a  different line - Only for desktop view.

### Why?
Cosmetic improvement.

# Outcomes:
## In the desktop view the facet filter is moved to the next line below.
Screenshot 2022-10-14 at 16 50 04

## In the mobile view the filter stays on the same line since the commodity code list has been moved to a block below.
Screenshot 2022-10-14 at 16 50 13

### DIFF
--- a/app/webpacker/src/stylesheets/_search.scss
+++ b/app/webpacker/src/stylesheets/_search.scss
@@ -32,3 +32,9 @@
     text-decoration: underline;
   }
 }
+
+div.facet-classifications-tag-container {
+  @media (max-width: $mobile-max-width) {
+    display: inline;
+  }
+}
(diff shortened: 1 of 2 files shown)


willfish (approved):

app/webpacker/src/stylesheets/_search.scss
```diff
@@ -32,3 +32,9 @@
     text-decoration: underline;
   }
 }
+
+div.facet-classifications-tag-container {
+  @media (max-width: $mobile-max-width) {
```
jebw: Should this be using the govuk responsive mixins?

jebw (approved):

app/webpacker/src/stylesheets/_search.scss
alexdesi: yes, I've replaced it with `govuk-media-query($until: tablet)`,
I've tested it and works well 👍

PR #1095, opened 2022-10-17
Fix flaky tests for webchat.

alexdesi:
### Jira link
No Jira ticket

### What?
Fix flaky tests for webchat.
This PR makes the tests independent of the value of the ENV var.
### Why?

I am doing this because:
We don't really like flaky tests 😜


jebw (approved):
LGTM

PR #1102, opened 2022-10-18
HOTT-2144: Adds no-hits page

willfish:
### Jira link

image

https://transformuk.atlassian.net/browse/HOTT-2144

### What?

I have added/removed/altered:

- [x] Added no hits page
- [x] Added specs

### Why?

I am doing this because:

- This is required for go live

### DIFF
--- a/app/views/beta/search_results/show.html.erb
+++ b/app/views/beta/search_results/show.html.erb
@@ -1,26 +1,9 @@
 <h1>Search results for '<%= @search_result.search_query_parser_result.corrected_search_query %>'</h1>
 
 <section id="beta-search-results-<%= @search_result.resource_id %>">
-  <div class="govuk-grid-row">
-    <% if @search_result.facet_filter_statistics.any? %>
-      <div class="govuk-grid-column-one-third">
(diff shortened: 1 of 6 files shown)


app/views/beta/search_results/show.html.erb
```diff
@@ -1,26 +1,9 @@
...
-        <% if @search_result.multiple_headings_view? %>
-          <%= render partial: 'multiple_headings_search_results' %>
-        <% else %>
-          <%= render partial: 'search_results' %>
-        <% end %>
-      </div>
-    <% end %>
-  </div>
+  <% if @search_result.hits.commodities.present? %>
+    <%= render 'with_hits' %>
```
jebw, line 5: General comment but if you pass the data into partials rather then relying on globals its easier to reuse partials in other locations in the future - keeps the api surface between view and partial well defined? 

```ruby
<%= render 'with_hits', search_result: @search_result %>
```

jebw (approved):

app/views/beta/search_results/show.html.erb
willfish: I'll do this separately but good shout

PR #1104, opened 2022-10-18
HOTT-2144: Search results bread crumbs

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2144

image

### What?

I have added/removed/altered:

- [x] Added bread crumbs to search results

### DIFF
--- a/app/views/beta/search_results/show.html.erb
+++ b/app/views/beta/search_results/show.html.erb
@@ -1,3 +1,12 @@
+<% content_for :top_breadcrumbs do %>
+  <%= generate_breadcrumbs(
+    t('breadcrumb.search_results'),
+    [
+      [t('breadcrumb.home'), home_path]
+    ]
+  ) %>
+<% end %>
+
 <h1>Search results for '<%= @search_result.search_query_parser_result.corrected_search_query %>'</h1>
 
 <section id="beta-search-results-<%= @search_result.resource_id %>">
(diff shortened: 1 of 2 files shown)


app/views/beta/search_results/show.html.erb
```diff
@@ -1,3 +1,12 @@
+<% content_for :top_breadcrumbs do %>
+  <%= generate_breadcrumbs(
+    t('breadcrumb.search_results'),
+    [
+      [t('breadcrumb.home'), home_path]
+    ]
+  ) %>
+<% end %>
+
 <h1>Search results for '<%= @search_result.search_query_parser_result.corrected_search_query %>'</h1>
```
alexdesi, line 10: I think we should use:
```
<%= page_header "Search results for
                &lsquo;#{h @search_result.search_query_parser_result.corrected_search_query}&rsquo;".html_safe
%>
```

this way includes  the switch button too:
Screenshot 2022-10-18 at 12 08 10

alexdesi: here is how I implemented it, in case you want to reuse some parts: https://github.com/trade-tariff/trade-tariff-frontend/pull/1103/files#diff-87a700faf2c560811479f70fb39aa863eb17d5dd6072198a4520a0f5cd3cd5d9R1-R3

willfish: Well, I guess when we merge yours we'll get that so don't need to do it here?

PR #1109, opened 2022-10-19
HOTT-2048: Created measure type page

maverickvi:
### Jira link

[HOTT-<2048>](https://transformuk.atlassian.net/browse/HOTT-2048)

Important: Backend PR needs to be merged before this frontend PR can be merged.

### What?

I have added/removed/altered:

- [ ] Created measure type page

### Why?

I am doing this because:

- We need to display measure types and preference codes data on the frontend

### Have you? (optional)

Screenshot 2022-10-19 at 15 45 44

Screenshot 2022-10-19 at 15 46 04


app/models/measure_type.rb
```diff
@@ -28,10 +28,20 @@ def description
...
+  def description_only
+    attributes['description']
+  end
+
   def supplementary_unit_import_only?
     id.in?(SUPPLEMENTARY_IMPORT_ONLY_MEASURE_TYPES)
   end
 
+  def md_details
+    File.read("db/measure_type_detail_texts/#{TradeTariffFrontend::ServiceChooser.service_name}/#{id}.md")
```
willfish: Rather than read a file on every request, I'd preload these into an object at boot keyed on the measure type id

app/models/measure_type.rb
```diff
@@ -28,10 +28,20 @@ def description
...
 
+  def description_only
+    attributes['description']
+  end
+
   def supplementary_unit_import_only?
     id.in?(SUPPLEMENTARY_IMPORT_ONLY_MEASURE_TYPES)
   end
 
+  def md_details
```
willfish:
```suggestion
  def details_text
```

willfish: Not sure about abbreviations tbh. It feels weird that the client needs to know the format, too

app/models/measure_type.rb
```diff
@@ -28,10 +28,20 @@ def description
...
+    File.read("db/measure_type_detail_texts/#{TradeTariffFrontend::ServiceChooser.service_name}/#{id}.md")
+  rescue Errno::ENOENT
```
willfish: I don't like exceptions to control decisions generally. How do you feel about just explicitly checking if the file exists?

app/models/measure_type.rb
willfish:
```suggestion
  def details_text
    Govspeak::Document.new(markdown_details_text, sanitize: true).to_html.html_safe if markdown_details_text
  end
```

config/routes.rb
```diff
@@ -62,6 +62,8 @@
 
   get 'geographical_areas/:id', to: 'geographical_areas#show', as: :geographical_area
 
+  get '/measure_type/:measure_type_id/:preference_code_id', to: 'measure_type#show', as: :measure_type
```
willfish: This feels weird to me. So the thing that uniquely identifies the measure type is the measure type id. Everything else should be a query param IMO

jebw: I would strongly disagree on this point - the page requires a dual primary key - the combination of :measure_type_id and :perference_code_id, it cannot be rendered with either just one or just the other. Making either part optional (which query params should be) is counter intuitive and results in an unrenderable page. I also think having half the primary key in one format (url component) and half in another (query param) would be wierd.

An alternative to using a slash for the separator on primary key components would be to use a hypen - thats less common but at least keeps both primary keys in the same domain

jebw: The better approach maybe to just add a preferences code segment to the URL - that way its more descriptive of what the page contains

`/measure_type/:measure_type_id/preference_code/:preference_code_id`

willfish: Your last comment was exactly what I was about to suggest, yeah. Let's go with that

willfish:
```
/measure_types/:measure_type_id/preference_codes/:id
```

willfish (approved):

PR #1110, opened 2022-10-20
HOTT-1952 Handle non html types in exceptions app

jebw:
This led to an odd behaviour previously where an unhandled exception would cascade up to the errors app, then trigger a ActionView::MissingTemplate exception in turn triggering the (now removed) rescue_from in ApplicationController and result in the 404 page

### Jira link

HOTT-1952

### What?

I have added/removed/altered:

- [x] Added error handler behaviour for non html content types

### Why?

I am doing this because:

- Previously these were triggering an ActionView::MissingTemplate resulting in the user seeing a 404 page for an unhandled exception on non HTML formats (eg JSON)

- Low, changes a unhandled exception in a JSON endpoint to return a 500 json document instead of a 404 html document

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -7,14 +7,26 @@ class ErrorsController < ApplicationController
   before_action :disable_search_form, :disable_switch_service_banner
 
   def not_found
-    render status: :not_found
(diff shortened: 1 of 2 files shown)


willfish (approved):
Amazing

PR #1112, opened 2022-10-21
HOTT-2162: Adds ability to use original search query

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2169

### What?

image

I have added/removed/altered:

- [x] Adds ability to use the original search query

### Why?

I am doing this because:

- This is required for go live


alexdesi (approved):
lgtm ✅

PR #1123, opened 2022-10-25
HOTT-2166: Fixed header CSS with search banner set to false

maverickvi:
### Jira link

[HOTT-<2166>](https://transformuk.atlassian.net/browse/HOTT-2166)

### What?

I have added/removed/altered:

- [ ] Fixed header CSS with search banner set to false

### Why?

I am doing this because:

- We had unwanted padding under the header when window width was greater than 1200px

### Have you? (optional)

without changes:
Screenshot 2022-11-09 at 09 33 31

with changes:
Screenshot 2022-11-09 at 09 33 23


app/views/layouts/_base.html.erb
```diff
@@ -19,7 +19,7 @@
     <a href="<%= home_path %>" class="govuk-header__link govuk-header__link--service-name" id="proposition-name">
       <%= trade_tariff_heading %>
     </a>
-    <nav id="proposition-menu"  aria-label="Top Level Navigation" class="govuk-header__navigation">
+    <nav id="proposition-menu"  aria-label="Top Level Navigation" class="govuk-header_">
```
jebw: this looks odd `govuk-header_` is unlikely to be a valid class name so it can probably be removed

jebw (approved):
LGTM :raised_hands:

PR #1136, opened 2022-10-31
HOTT-2030 Show the Rules of Origin button in the

jebw:
correct circumstances

### Jira link

HOTT-2030

### What?

I have added/removed/altered:

- [x] Changed the logic around what to show on Rules of Origin tab
- [x] UX tweak to use GovUK inset text for 'no tariff' message

### Why?

I am doing this because:

- If there were multiple preferential tariffs, we were telling the user that there weren't any - this primarily affected Vietnam

### Have you? (optional)

- Low, feature flagged off

### DIFF
--- a/app/views/rules_of_origin/_tab.html.erb
+++ b/app/views/rules_of_origin/_tab.html.erb
@@ -5,7 +5,7 @@
       <%= country_flag_tag country_code, alt: "Flag for #{country_name}" %>
     </h2>
 
-    <% if rules_of_origin_schemes.present? %>
+    <% if rules_of_origin_schemes.any? %>
       <p>
         Preferential tariff treatment reduces the duties you are required to pay when importing or exporting goods to or from <%= country_name %>.
       </p>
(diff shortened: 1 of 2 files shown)


app/views/rules_of_origin/_tab.html.erb
```diff
@@ -5,7 +5,7 @@
       <%= country_flag_tag country_code, alt: "Flag for #{country_name}" %>
     </h2>
 
-    <% if rules_of_origin_schemes.present? %>
+    <% if rules_of_origin_schemes.any? %>
```
alexdesi, line 8: Nice, I've read `any?` is faster than `present?` (since it uses limit 1 in the SQL generated)

alexdesi (approved):
It sounds good! ✅ 
since we added some logic, should we add a test for it?
or is it something we leave to @bhavanibandla

app/views/rules_of_origin/_tab.html.erb
jebw: Yeh - in this instance its operating on an array _but_ `#present?` checks for `#nil?` first and tries to be clever - `#any?` will only work if data is an enumerable - this always should be one, if its not then it'll raise which is good

jebw: Re test coverage - good shout but just been looking over the spec and it does have coverage for single, multiple and none scheme scenarios so I _think_ we're good

I have found there's a couple of 'old' specs for the old legacy tab still in there - they are passing because they check for absence rather then presence so I've adding a commit dropping those

PR #1141, opened 2022-11-01
HOTT-2196 Fix for no end date on quota definitions

jebw:
### Jira link

HOTT-2196

### What?

I have added/removed/altered:

- [x] Allow for missing end date in `govuk_date_range` helper

### Why?

I am doing this because:

- It is used on the quota definitions popup, and there are quotas without end dates

### Have you? (optional)

- Low, increases flexibility of date range helper and fixes an existing issue


app/helpers/application_helper.rb
```diff
@@ -130,7 +130,11 @@ def present_from_to(from, to)
   end
 
   def govuk_date_range(from, to)
-    "#{from.to_date.to_formatted_s(:long)} to #{to.to_date.to_formatted_s(:long)}"
+    if to.present?
+      "#{from.to_date.to_formatted_s(:long)} to #{to.to_date.to_formatted_s(:long)}"
```
alexdesi, line 134: just to check: do we need "From " at the start of this?

spec/helpers/application_helper_spec.rb
```diff
@@ -342,6 +342,18 @@
 
       it { is_expected.to eql '1 January 2022 to 1 June 2022' }
     end
+
+    context 'with blank :to date' do
```
alexdesi: nitpicking:
Does 
`context 'when :to date is blank' do`
read better? 🤔

spec/helpers/application_helper_spec.rb
```diff
@@ -342,6 +342,18 @@
...
+      subject { govuk_date_range from, '' }
+
+      it { is_expected.to eql 'From 1 January 2022' }
+    end
+
+    context 'with nil :to date' do
```
alexdesi: same as before

alexdesi (approved):
It looks good ✅ 
Just a couple of nitpickings

app/helpers/application_helper.rb
jebw: Not for our current usage though appreciate that conflicts with how GovUK guidelines say we should be formatting dates

PR #1147, opened 2022-11-07
HOTT-2181 Show errors for invalid date

jebw:
### Jira link

HOTT-2181

### What?

I have added/removed/altered:

- [x] Added a FindCommodity model to validate the params used in the find commodity form
- [x] Changed the FindCommodity form to submit back to the FindCommodities controller, then redirect on to search if valid

### Why?

I am doing this because:

- Currently we do not (cannot) validate submissions on the find commodities form. If the user submits an unparseable date this results in a 500 error

### Have you? (optional)

- High, changes to homepage and search behaviour

### Screenshots

image


maverickvi (dismissed):

app/controllers/find_commodities_controller.rb
```diff
@@ -5,5 +5,16 @@ def show
     @no_shared_search = true
     @hero_story = News::Item.latest_for_home_page
     @recent_stories = News::Item.updates_page.slice(0, 3)
+
+    @find_commodity = FindCommodity.new(search_params)
+    return unless @find_commodity.performing_search? && @find_commodity.valid?
+
+    redirect_to perform_search_path redirect_params
```
alexdesi: I am in favour of being explicit to make it easier to understand and maintain: 
```
if @find_commodity.performing_search? && @find_commodity.valid?
  render :show
else
  redirect_to perform_search_path redirect_params
end
```

app/models/find_commodity.rb
```diff
@@ -0,0 +1,49 @@
...
+  def date=(date)
+    @year = date.year
+    @month = date.month
+    @day = date.day
+  end
+
+private
+
+  def date_is_valid
+    unless date_parts_are_valid?
```
alexdesi: I think we can just use:
```
unless Date.valid_civil?( year.to_i, month.to_i, day.to_i)
```
`valid_civil?` is a clear and documented function

app/controllers/application_controller.rb
```diff
@@ -71,11 +71,11 @@ def search_invoked?
 
   def set_search
     # rubocop:disable Naming/MemoizedInstanceVariableName
-    @search ||= Search.new(search_attributes)
+    @search ||= Search.new(search_params)
```
alexdesi, line 74: just checking, do we need to set this an instance var?
I noticed that we are using `f` in the view (but I may miss other uses)

config/routes.rb
```diff
@@ -179,7 +179,10 @@
                 module: 'commodities'
     end
   end
-  get '/find_commodity', to: 'find_commodities#show'
```
alexdesi, line 182: This seems to be more clear than:
```
  resource :find_commodity, only: %i[show]
  resolve('FindCommodity') { route_for(:find_commodity) }
```
Are there functionalities that I am not considering?
Or are they equivalent?

alexdesi (commented):
There are lots of improvements here! 
I've added some minor comments

config/routes.rb
jebw: @alexdesi They are not directly equivalent

* the `resource` line *is* effectively equivalent to a single `GET` line _but_ is the favoured approach within the rails framework, I think because using resource makes following `resource` conventions the default pattern if this is ever gets extended for other actions
* the `resolve` line is the solution for rails defaulting to plural routes rather than singular routes (see https://guides.rubyonrails.org/routing.html#singular-resources)
  * when given a `FindCommodity` model instance, by default rails would look for a route of `/find_commodities/:id`
  * Because this is singular though there is no `:id` and we instead want it to just use `find_commodity_path`

app/controllers/find_commodities_controller.rb
```diff
@@ -5,5 +5,21 @@ def show
     @no_shared_search = true
     @hero_story = News::Item.latest_for_home_page
     @recent_stories = News::Item.updates_page.slice(0, 3)
+
+    @find_commodity = FindCommodity.new(search_params)
+
+    if search_params.blank?
+      nil
```
alexdesi, line 12: This does nothing :thinking:

spec/controllers/search_controller_search_spec.rb
```diff
@@ -151,20 +159,22 @@
...
-      end
+          response
+        end
 
-      let(:year)    { 'errr' }
-      let(:month)   { 'er' }
-      let(:day)     { 'er' }
+        let(:year)    { 'errr' }
+        let(:month)   { 'er' }
+        let(:day)     { 'er' }
```
alexdesi, line 172: the indentation looked right to me,
but I might be wrong :thinking:

spec/controllers/search_controller_search_spec.rb
```diff
@@ -151,20 +159,22 @@
         it { expect(assigns(:search)).to be_a(Search) }
         it { is_expected.to redirect_to(sections_path) }
       end
-    end
 
-    context 'when date param components are invalid' do
-      subject(:do_response) do
-        post :search, params: { year:, month:, day: }
+      context 'when date param components are invalid' do
```
alexdesi, line 163: is the indentation incorrect?

alexdesi (approved):
nice one! :rocket:

app/controllers/find_commodities_controller.rb
```diff
@@ -5,5 +5,19 @@ def show
...
+
+    return unless @find_commodity.valid?
+
+    redirect_to perform_search_path redirect_params
+  end
+
+  private
+
+  def redirect_params
+    search_params.select { |_k, v| v.present? }
```
alexdesi, line 21: I've checked,
compact works also with hashes.
So
`search_params.compact`
does the same job.

PR #1148, opened 2022-11-07
HOTT-2186: Uppercased country code before hitting the backend API

maverickvi:
### Jira link

[HOTT-<2186>](https://transformuk.atlassian.net/browse/HOTT-2186)

### What?

I have added/removed/altered:

- [ ] Uppercased country code before hitting the backend API

### Why?

I am doing this because:

- the backend API requires country_code parameters to be uppercased when calling the GeographicalArea endpoint. This fix could also be done on the backend if we decided to change the API requirements and uppercased all country_code parameters by default.

Screenshot 2022-11-07 at 15 19 23

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -148,6 +148,6 @@ class ApplicationController < ActionController::Base
   end
 
   def country
-    params['country']
+    params['country'].try(:upcase)
   end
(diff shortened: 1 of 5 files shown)


jebw (approved):
LGTM :raised_hands:

PR #1160, opened 2022-11-08
HOTT-2213: Removed preferential treatment text

maverickvi:
### Jira link

[HOTT-<2213>](https://transformuk.atlassian.net/browse/HOTT-2213)

### What?

I have added/removed/altered:

- [ ] Removed preferential treatment text

### Why?

I am doing this because:

- We no longer need it

Screenshot 2022-11-08 at 14 07 21


alexdesi (approved):
It looks good! ✅

PR #1173, opened 2022-11-18
HOTT-2224 Reintroducing Hyphens

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-2224

### What?

Add hyphens on the heading page to help to identify the level of the commodity.

### Why?
Provide a better user experience.


alexdesi:
Screenshots:
Screenshot 2022-11-18 at 13 38 47

Screenshot 2022-11-18 at 13 38 01

app/webpacker/src/stylesheets/_commodity-tree.scss
```diff
@@ -186,19 +186,21 @@ article.tariff {
       position:relative;
       word-wrap: break-word;
 
+    // Creates the hiphen lines for the commodities
```
alexdesi: ops, misspell here: h**y**phen

app/views/shared/_commodity_tree.html.erb
```diff
@@ -14,7 +14,7 @@
       </div>
     </div>
     <ul class="commodities js-commodities">
-      <%= render commodities %>
+      <%= render commodities, initial_indent: commodities.first.number_indents %>
```
jebw: What happens when there are no commodities? This will trigger an exception presumably - can this be `commodities.first&.number_indents` instead - thanks

app/webpacker/src/stylesheets/_commodity-tree.scss
jebw: misspelling

app/webpacker/src/stylesheets/_commodity-tree.scss
```diff
@@ -186,6 +186,24 @@ article.tariff {
       position:relative;
       word-wrap: break-word;
 
+    // Creates the hiphen lines for the commodities 
+      @for $i from 2 through 50 {
+        &.level-#{$i} {
+            $url_list: ();
+            $position_list: ();
+            $dashes_spacing: 20px;
```
jebw: When we discussed this we concluded, `$dashes_spacing` must always be the same value as `$token_list`

Please either chain the variable from `$token_width` (ie `$dashes_spacing: $token-width`) or drop it entirely and just use the `$token_width` variable instead on line 198. Thanks

jebw (changes requested):
Looks good, nice work on solving the subheadings issue. Couple of minor issues  - can you  address the duplicated variable, correct the comment spelling and drop the unused image. :+1:

app/webpacker/src/stylesheets/_commodity-tree.scss
alexdesi: thanks for the reminder, I'll go with
`$dashes_spacing: $token-width`

I think it makes clearer the purpose of the variables (even if the value is the same)

app/webpacker/src/stylesheets/_commodity-tree.scss
alexdesi: fixed

app/webpacker/src/stylesheets/_commodity-tree.scss
```diff
@@ -186,6 +186,24 @@ article.tariff {
       position:relative;
       word-wrap: break-word;
 
+    // Creates the hiphen lines for the commodities 
+      @for $i from 2 through 50 {
```
alexdesi: put 20 as max level

alexdesi: fixed.

app/views/shared/_commodity_tree.html.erb
alexdesi: I've just checked, and this is something that does not happen now, thanks to the `rescue_from Faraday::ResourceNotFound` on `GoodsNomenclaturesController`.

Anyway, it's a change that makes sense, I am happy to do it, and default `initial_indent` to one, in case of empty commodities.

app/helpers/commodities_helper.rb
```diff
@@ -9,8 +9,11 @@ def leaf_position(commodity)
     end
   end
 
-  def commodity_level(commodity)
-    "level-#{commodity.number_indents}"
+  def commodity_level(commodity, initial_indent)
+    initial_indent ||= 1
```
alexdesi, line 13: we need `initial_indent ||= 1`, 
since `def commodity_level(commodity, initial_indent=1)`
behaves in a slightly different way: you can explicitly pass nil.

jebw (approved):
LGTM

PR #1175, opened 2022-11-21
HOTT-2245: Hide import trade summary if basic_third_country_duty is nil

maverickvi:
### Jira link

[HOTT-<2245>](https://transformuk.atlassian.net/browse/HOTT-2245)

### What?

I have added/removed/altered:

- [ ] Hidden import trade summary box if basic_third_country_duty is nil

### Why?

I am doing this because:

- There was a bug when rendering the import trade summary partial

Screenshot 2022-11-21 at 10 49 35

Screenshot 2022-11-21 at 10 40 59

### DIFF
--- a/app/views/rules_of_origin/_import_trade_summary.html.erb
+++ b/app/views/rules_of_origin/_import_trade_summary.html.erb
@@ -1,4 +1,4 @@
-<div class="govuk-inset-text">
+<div class="govuk-inset-text import-trade-summary">
     <h3 class="govuk-heading-s">How rules of origin could affect the import duty payable</h3>
 
     <ul class="govuk-list govuk-list--bullet">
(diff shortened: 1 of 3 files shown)


jebw (dismissed):
LGTM :+1:

jebw (changes requested):
Can we add a spec for when we encounter this scenario

https://github.com/trade-tariff/trade-tariff-frontend/blob/main/spec/views/rules_of_origin/_tab.html.erb_spec.rb#L83

eg, just something simple like the existing one for when present

```ruby
context 'when preferential quota duty is present' do
  let(:import_trade_summary) { attributes_for(:import_trade_summary, :with_quota_duty, basic_third_country_duty: nil) }

  it { expect(rendered_page).not_to have_css '.preferential-quota-duty' }
end
```

jebw (approved):
LGTM :raised_back_of_hand:

PR #1185, opened 2022-11-25
HOTT-2259 Move the critical footnotes after the header.

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-2259

### What?
Move the critical footnotes after the header to make it more evident.

### DIFF
--- a/app/views/headings/show.html.erb
+++ b/app/views/headings/show.html.erb
@@ -1,19 +1,21 @@
 <% content_for :title, goods_nomenclature_title(@heading) %>
 
-
 <% content_for :head do %>
   <meta name="description" content="<%= @heading %>">
   <link rel='alternate' type='application/json' href='<%= heading_url(@heading, format: :json) %>' title='Heading information page in JSON format' />
 <% end %>
-
+  
 <% if @heading.declarable? %>
   <%= render 'commodities/header', commodity_code: (@heading.page_heading + '000000') %>
-<% else %>
-  <%= page_header @heading.page_heading %>
 <% end %>
 
-<% @heading.critical_footnotes.each do |footnote| %>
-  <%= render 'footnotes/critical_warning', footnote: footnote %>
+<%= page_header do %>
+  <% @heading.critical_footnotes.each do |footnote| %>
+    <%= render 'footnotes/critical_warning', footnote: footnote %>
+  <% end %>
+  <h1 class="<%= css_heading_size(@heading.page_heading) %>">
+    <%= sanitize @heading.page_heading %>
+  </h1>
 <% end %>
 
 <%= render 'shared/context_tables/heading' %>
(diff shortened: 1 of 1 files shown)


app/views/headings/show.html.erb
```diff
@@ -1,19 +1,14 @@
 <% content_for :title, goods_nomenclature_title(@heading) %>
 
-
 <% content_for :head do %>
   <meta name="description" content="<%= @heading %>">
   <link rel='alternate' type='application/json' href='<%= heading_url(@heading, format: :json) %>' title='Heading information page in JSON format' />
 <% end %>
-
-<% if @heading.declarable? %>
```
alexdesi, line 9: I've removed this, because 
this branch of the if (heading.declarable?) is never reached,
since if it is a declarable heading, the controller would had redirect it to the commodity view.

alexdesi:
Screenshot of the implementation:
Screenshot 2022-11-25 at 15 23 24

app/views/shared/_page_header.html.erb
```diff
@@ -7,11 +7,11 @@
     <%= switch_service_button if show_switch_service %>
   </span>
 
+  <%= extra_content %>
```
jebw: This is significantly changing the behaviour of the `_page_header` partial which will affect the display of other pages using `extra_content`

The  partial is already quite flexible, and if it is insufficiently flexible then its better to either not use it in the header page or failing that improve its flexibility further.

app/views/headings/show.html.erb
jebw: I'm thinking this assertion is might be wrong? Doesn't a declarable heading render `show`, but then it renders the same `measures` partials as `commodities/show` ?

jebw: Looks like this is rendering the declarable content on line 25 of this file?

jebw (changes requested):
Couple of points I'm not sure about with this PR

1. Looks like `headings/show` _does_ render declarable headings
2. The change to `shared/_page_header` will impact other pages but I'm pretty sure you can do these changes without modfying that partial anyway?

app/views/headings/show.html.erb
jebw: Looking at the helper method which drives this - if you supply a nil header _but_ do supply a block then it will render just the caption + block which is what you need for this I think?

alexdesi: We've checked that actually the code inside the IF is never executed.

alexdesi: but I am happy to move it to a different PR!

app/views/shared/_page_header.html.erb
alexdesi: yes, as discussed we actually do not need to move the extra_content 👍

app/views/headings/show.html.erb
alexdesi: In the end, I thought it was OK to do It in this PR but on a different commit.

jebw (approved):
LGTM

PR #1187, opened 2022-11-28
HOTT-2254: Enable measure of type 464

maverickvi:
### Jira link

[HOTT-<2254>](https://transformuk.atlassian.net/browse/HOTT-2254)

### What?

I have added/removed/altered:

- [x] Enable measure of type 464 on commodities page

### Why?

I am doing this because:

- We were asked to enable it

Screenshot 2022-11-28 at 11 27 22


jebw (approved):
LGTM :+1:

PR #1203, opened 2022-12-02
HOTT-2256: Remember selected tab on commodities page when country is …

maverickvi:
…selected

### Jira link

[HOTT-<2256>](https://transformuk.atlassian.net/browse/HOTT-2256)

### What?

I have added/removed/altered:

- [ ] Added a hidden field to remember which tab the user has selected

### Why?

I am doing this because:

- We want to remember the selected tab when switching countries to deliver a better user experience

Screenshot 2022-12-02 at 13 31 39

### DIFF
--- a/app/controllers/trading_partners_controller.rb
+++ b/app/controllers/trading_partners_controller.rb
@@ -15,7 +15,7 @@ class TradingPartnersController < ApplicationController
     if @trading_partner.valid?
       redirect_to goods_nomenclature_path(
         country: @trading_partner.country,
-        anchor: trading_partner_params["anchor"],
+        anchor: trading_partner_params['anchor'],
       )
     elsif should_not_render_errors?
       redirect_to goods_nomenclature_path
--- a/app/controllers/trading_partners_controller.rb
+++ b/app/controllers/trading_partners_controller.rb
@@ -15,6 +15,7 @@ class TradingPartnersController < ApplicationController
     if @trading_partner.valid?
       redirect_to goods_nomenclature_path(
         country: @trading_partner.country,
+        anchor: trading_partner_params["anchor"],
       )
     elsif should_not_render_errors?
       redirect_to goods_nomenclature_path
@@ -30,6 +31,6 @@ class TradingPartnersController < ApplicationController
   end
 
   def trading_partner_params
-    params.fetch(:trading_partner, {}).permit(:country)
+    params.fetch(:trading_partner, {}).permit(:country, :anchor)
   end
 end
--- a/app/controllers/trading_partners_controller.rb
+++ b/app/controllers/trading_partners_controller.rb
@@ -10,12 +10,12 @@ class TradingPartnersController < ApplicationController
   end
 
   def update
-    @trading_partner = TradingPartner.new(trading_partner_params)
+    @trading_partner = TradingPartner.new(country: trading_partner_params[:country])
 
     if @trading_partner.valid?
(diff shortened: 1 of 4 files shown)


app/controllers/trading_partners_controller.rb
```diff
@@ -15,6 +15,7 @@ def update
     if @trading_partner.valid?
       redirect_to goods_nomenclature_path(
         country: @trading_partner.country,
+        anchor: trading_partner_params['anchor'],
```
willfish: Shouldn't this be part of the goods_nomenclature_path construction? Is the referer blank for some reason?

```ruby
  def anchor
    referer&.fragment
  end
```

app/models/trading_partner.rb
```diff
@@ -3,6 +3,7 @@ class TradingPartner
   include ActiveModel::Attributes
 
   attribute :country, :string
+  attribute :anchor, :string
```
willfish: AFAIK, this attribute is never used?

app/views/shared/_country_picker.html.erb
```diff
@@ -2,6 +2,7 @@
   <fieldset class="govuk-fieldset country-picker js-country-picker govuk-!-font-size-16">
     <div class="country-picker-container">
       <%= f.hidden_field :render_errors, value: false %>
+      <%= f.hidden_field :anchor, value: nil %>
```
willfish, line 5: You shouldn't need this? Unless @jebw knows something I don't?

maverickvi: This is being set here: $(this.selectElement).siblings('input[name$="[anchor]"]').val(window.location.hash.substring(1)); in app/webpacker/src/javascripts/commodities.js

app/models/trading_partner.rb
maverickvi: If I don't set this I get attribute errors

app/controllers/trading_partners_controller.rb
maverickvi: anchor is provided by Rails

app/views/shared/_country_picker.html.erb
willfish: So we're setting a hidden field with some javascript. I guess my point is I can't conceive of why we need to when either the referrer or the request url in the rails controller action is aware of this? Isn't it a) duplicating what we've already got in the wrong place and b) therefore adds unnecessary extras.

A page anchor also has 0 to do with the trading partner so it smells a bunch

Why can't you just use the goods_nomenclature_path method to do this for you? Is there a bug in it?

app/models/trading_partner.rb
willfish: I suspect because you're permitting attributes that get assigned to the form model that aren't used by the form model or its consumers.

app/controllers/trading_partners_controller.rb
willfish: So why can't we use that?

app/views/shared/_country_picker.html.erb
willfish: I get this now:

1. The click event to change tab, updates the anchor client side but isn't visible to the server
2. We need to make sure it is visible to the server so that we can redirect to the right tab on the commodities page
3. This does not touch the trading partners page (even though it could but we'd need the referrer anchor for this to work on the javascript side)

willfish (approved):

PR #1201, opened 2022-12-01
HOTT-2230: Adds switching beta search on/off

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2230

### What?

I have added/removed/altered:

- [x] Adds ability to enable/disable beta search via the UI
- [x] Adds coverage for this new feature

### Why?

I am doing this because:

- This is required as part of the go-live requirements for beta search

PR #1205, opened 2022-12-05
Revert "HOTT-2230: Adds switching beta search on/off"

willfish:
Reverts trade-tariff/trade-tariff-frontend#1201


willfish:
We need to make sure that this feature only applies to dev/staging after getting some clarification from Matt.

PR #1211, opened 2022-12-05
HOTT-2282: Fix heading/subheading overview measure alignment

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2282

image

### What?

I have added/removed/altered:

- [x] Removed broken flex display configuration

### Why?

I am doing this because:

- This only seems to negatively affect the alignment of the overview measure cells in a desktop browser and seems to add no value/prevents us from aligning properly

### DIFF
--- a/app/webpacker/src/stylesheets/_commodity-tree.scss
+++ b/app/webpacker/src/stylesheets/_commodity-tree.scss
@@ -242,11 +242,6 @@ article.tariff {
         @media (min-width: 641px) {
           float: right;
           width: 514px;
-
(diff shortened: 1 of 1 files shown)


jebw (approved):
LGTM

Looks like it was originally done with fixed widths, then hacked to tried and solve the issues you get when you using fixed widths 

https://github.com/trade-tariff/trade-tariff-frontend/commit/9b4028e9c5f49eeaff87c15456e3b5e08df37973

PR #1212, opened 2022-12-05
HOTT-2294 show collection name

jebw:
### Jira link

HOTT-2294

### What?

I have added/removed/altered:

- [x] Fixed display of collection name when filtering by collection

### Why?

I am doing this because:

- Saying 'All collections' when its not is misleading for the user

### Have you? (optional)

- None - epic branch

### DIFF
--- a/app/models/news/collection.rb
+++ b/app/models/news/collection.rb
@@ -19,5 +19,15 @@ module News
     def to_param
       slug.presence || id
     end
+
+    def matches_param?(collection_id)
+      if collection_id.to_s.match? %r{\A\d+\z}
+        id == collection_id.to_i
+      elsif slug.present?
+        slug == collection_id
+      else
+        false
+      end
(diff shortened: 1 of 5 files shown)


maverickvi (approved):

app/models/news/collection.rb
```diff
@@ -19,5 +19,15 @@ def id
     def to_param
       slug.presence || id
     end
+
+    def matches_param?(collection_id)
```
willfish, line 23: Isn't this something backend can do?

jebw: At present, no

I _could_ add a `#show` action to do it _but_ I'm already requesting _all_ collections from the backend so adding another api request (and supporting API on the backend) seems a pretty complicated solution to 'find this item in the array I already have')

Will go just iterating the list I already have for now

PR #1228, opened 2022-12-12
HOTT-2297: Adds presentation of quota closed and transferred event

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2297

### What?

I have added/removed/altered:

- [x] Added view for quota closed and transferred events
- [x] Removed 20 day requirement for fetching pending safeguard quota events
- [x] Adds specs for new transfer events

### Why?

I am doing this because:

- This is required to enable transfer events to present in the quota modal UI
- The requirement of showing pending balance events only within 20 days of the definition period no longer applies

### DIFF
--- a/app/models/order_number/definition.rb
+++ b/app/models/order_number/definition.rb
@@ -35,6 +35,8 @@ class OrderNumber
     has_one :order_number
     has_many :measures
 
+    has_one :incoming_quota_closed_and_transferred_event, class_name: 'QuotaClosedAndTransferredEvent'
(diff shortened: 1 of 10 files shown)


willfish:
**Pending balances still show regardless of the current date of the safeguard quota:**

image

**Transferred balances show when there is an event and we're past the closing date**

image

**Transferred balances on the closing date**

image

maverickvi (approved):

PR #1236, opened 2022-12-15
HOTT-2370 Check homepage renders as part of deployment

jebw:
If it doesn't render, we shouldn't switch to the new app

### Jira link

HOTT-2370

### What?

I have added/removed/altered:

- [x] Check the home page as well as the healthcheck

### Why?

I am doing this because:

- If the home page isn't functional for any reason we shouldn't be deploying the app
- Would catch rendering errors for example
  - Eg something in layout relying upon data not deployed on prod for example

- Changes deployment pipeline with additional robustness check

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,11 +46,27 @@ commands:
           name: "Verify new version is working on dark URL."
           command: |
             sleep 15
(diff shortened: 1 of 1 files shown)


willfish (approved):
Awesome

PR #1242, opened 2022-12-19
HOTT-2388: Hide first quarter pending balances

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2388

### What?

I have added/removed/altered:

- [x] Alter logic of when to show pending balances to hide them during the first quarter

### Why?

I am doing this because:

- Balances can only be transferred for a given quota year and not between quota years


jebw (approved):
LGTM

PR #1268, opened 2023-01-13
HOTT-2472: Adds credibility check measure section

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2472

### What?

I have added/removed/altered:

- [x] Added credibility check measure group

### Why?

I am doing this because:

- This is required for showing credibility measures in the frontend and is a client request


app/views/measures/grouped/_credibility_checks.html.erb
```diff
@@ -0,0 +1,3 @@
+<%# <div class="measure-inset"> %>
```
willfish, line 1: This is just a placeholder for now and we can populate content for it when HMRC have provided it. We do expect them too, soon

willfish:
image


image

maverickvi (approved):

PR #1293, opened 2023-01-27
HOTT-2352: Added table to display additional code and duty

maverickvi:
### Jira link

[HOTT-<TODO>](https://transformuk.atlassian.net/browse/HOTT-2352)

Important: This PR needs to be merged 1 day after https://github.com/trade-tariff/trade-tariff-backend/pull/1010 to allow time for nightly rake tasks to finish

### What?

I have added/removed/altered:

- [ ] Added table to display additional code and duty expression

### Why?

I am doing this because:

- this will help users understand which code belongs to a duty

Screenshot 2023-01-31 at 14 25 58


app/views/commodities/_commodity.html.erb
```diff
@@ -22,8 +22,26 @@
...
+          <% if tcd.count > 1 %>
+            <table class='table'>
+              <thead>
+                <tr>
+                  <th>Code</th>
+                  <th>Duty</th>
+                </tr>
+              </thead>
+              <tbody>
+              <% tcd.each do |t| %>
```
willfish:
```suggestion
              <% tcd.each do |m| %>
```

app/webpacker/src/stylesheets/_commodity-tree.scss
```diff
@@ -554,3 +555,8 @@ article.tariff {
     }
   }
 }
+
+.additional-code-table {
+  font-size: 14px;
```
jebw: rather then duplicate govuk sizes, we should defer to them 
- see https://github.com/trade-tariff/trade-tariff-frontend/blob/main/app/webpacker/src/stylesheets/_commodity-codes.scss#L89

And for the margins
- https://design-system.service.gov.uk/styles/spacing/#applying-spacing-in-your-own-css

jebw (changes requested):
Looks good but lets DRY out the css incase GovUK tweak things in the future :raised_hands:

app/views/commodities/_commodity.html.erb
```diff
@@ -23,7 +23,10 @@
         <% end %>
         <div class="duty" aria-describedby="commodity-duty-title">
           <% tcd = commodity.overview_measures.third_country_duties %>
-          <% if tcd.count > 0 %>
+
+          <% if tcd.count > 1 %>
+            <%= render '/commodities/additional_code_table', measures: tcd %>
```
jebw: This should not require a leading slash - I'm not sure about when in a view (and obviously rendering a partial) but in a controller that actually refers to a specific template file rather a relative one

app/views/commodities/_commodity.html.erb
```diff
@@ -23,7 +23,10 @@
         <% end %>
         <div class="duty" aria-describedby="commodity-duty-title">
           <% tcd = commodity.overview_measures.third_country_duties %>
-          <% if tcd.count > 0 %>
+
+          <% if tcd.count > 1 %>
```
jebw, line 27: This could be `if tcd.many?`, and the second one `tcd.any?`

spec/views/commodities/_additional_code_table.html.erb_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'spec_helper'
+
+RSpec.describe 'commodities/_additional_code_table', type: :view do
+  subject(:rendered_page) do
+    render 'commodities/additional_code_table', measures: [measure, measure]
+    rendered
+  end
+
+  let(:measure) { build :measure, :vat, :with_additional_code }
```
jebw, line 9: could use `build_list` here ?

```ruby
let(:measures) { build_list :measure, 2, :vat, :with_additional_code }
```

jebw (changes requested):
Couple of suggestions for coding polish but the only detail I'm keen to see change is to remove the leading slash from the partial name

Not convinced this is actually a problem but seems better to stick to the standard convention

jebw (approved):
Approved - looks good, can you just remove the spurious `//here` you've added in `_commodity-tree.scss` before merging. Thanks

PR #1305, opened 2023-01-31
HOTT-2603: Temporarily hide non-prohibitive STW links

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2603

### What?

I have added/removed/altered:

- [x] Temporarily hide non-prohibitive STW links
- [x] Make sure we're showing the right message when no country is selected

### Why?

I am doing this because:

- STW is currently broken for these links


willfish:
Country selected but no prohibitive/conditionally prohibitive measures

image

No country selected

image

maverickvi (approved):

PR #1306, opened 2023-02-01
HOTT-2603: Feature flag STW integration

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2603

### What?

I have added/removed/altered:

- [x] Feature flag single trade window integration

### Why?

I am doing this because:

- We're in this weird state where the prohibitive state is not shown because of broken data from DIT and where the STW platform itself has issues. Effectively this means we've disabled all variants of STW messaging so to simplify/make configurable in the environment, I've added this

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -18,6 +18,7 @@ PORT=3001
 ROO_WIZARD=true
 SEARCH_BANNER=true
 SERVICE_DEFAULT=uk
+STW_ENABLED=false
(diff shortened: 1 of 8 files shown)


jebw (approved):
If I'm following this right this changes from the 'is this declarable' logic to a straight feature flag - would it make more sense as a combination of the two -> that way we can turn it off whenever we need to but don't need a code change to turn it back on?

PR #1308, opened 2023-02-01
HOTT-2592:  Display measure additional codes with new messages

maverickvi:
### Jira link

[HOTT-<TODO>](https://transformuk.atlassian.net/browse/HOTT-2592)

### What?

I have added/removed/altered:

- [ ] Displayed measure additional codes with new messages

### Why?

I am doing this because:

- To make it clearer for users to understand if restrictions apply or do not apply on a measure

### Have you? (optional)

Screenshot 2023-02-06 at 11 59 37


app/models/measure_type.rb
```diff
@@ -48,6 +48,13 @@ def safeguard?
     SAFEGUARD_TYPES.include?(id)
   end
 
+  def abbreviation
+    return 'Prohibition' if description.scan(/Prohibition/i).present?
```
jebw, line 52: Works fine but for future reference, `.match?(/Prohibition/i)` would be quicker 

-> `#scan` will need to scan the entire string for all matches where as `#match?` can just return as soon as it finds the first match

app/models/additional_code.rb
```diff
@@ -4,6 +4,8 @@ class AdditionalCode
   include ApiEntity
   include HasGoodsNomenclature
 
+  RESIDUAL_CODES = %w[49 98 99].freeze
```
willfish, line 7: We ought to be able to do this:

```ruby
enum :additional_code, {
   residual: %w[49 98 99]
}
```

willfish: You may need to add the additional code field, here: https://github.com/trade-tariff/trade-tariff-backend/blob/main/app/serializers/api/v2/additional_code_serializer.rb

jebw (approved):
LGTM :raised_hands: very clean

willfish (changes requested):
We're adding logic to the view, we should add tests for this

app/views/measures/_measure.html.erb
```diff
@@ -25,15 +25,18 @@
...
+          <% end %>
+          <strong><%= measure.additional_code.code %></strong>
+        </p>
 
-      <% if measure.additional_code.formatted_description.present? %>
-        <span class='table-line'>
-          <%= sanitize measure.additional_code.formatted_description %>
-        </span>
-      <% end %>
+        <p class="govuk-!-font-size-16"><%= sanitize measure.additional_code.formatted_description if measure.additional_code.formatted_description.present? %></p>
```
jebw: Just spotted this - conditional is at wrong level, this will generate an empty `<p>` tag instead of no `<p>` tag if `formatted_description` is not `present?`

willfish (approved):

PR #1312, opened 2023-02-06
HOTT-2299: Fixed country dropdown on commodities page

maverickvi:
### Jira link

[HOTT-<TODO>](https://transformuk.atlassian.net/browse/HOTT-2299)

### What?

I have added/removed/altered:

- [ ] Fixed country dropdown on commodities page

### Why?

I am doing this because:

- it was rendering incorrectly and pushing page content downwards

### Have you? (optional)

Screenshot 2023-02-06 at 15 19 46


willfish (approved):

jebw (changes requested):
I think this is wrong - this is replacing the inline css to behave like overlay _but_ the issue is we want the overlay behaviour, so we should change the javascript to use the overlay config option

https://alphagov.github.io/accessible-autocomplete/#other-options

app/webpacker/src/javascripts/commodities.js
```diff
@@ -752,6 +752,7 @@
                 showAllValues: true,
                 confirmOnBlur: true,
                 alwaysDisplayArrow: true,
+                displayMenu: 'overlay',
```
willfish, line 755: Do we have any javascript tests for this? Is it appropriate to have coverage here?

willfish (approved):
LGTM assuming @jebw is happy

app/webpacker/src/javascripts/commodities.js
maverickvi: This is provided by the add-on so we can just assume they have throughly tested this feature

willfish: Yeah, that's fair enough - this is essentially configuration

jebw (approved):
LGTM :raised_hands:

PR #1320, opened 2023-02-08
HOTT-2622: Adds UI for deriving goods nomenclatures

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2622

### What?

I have added/removed/altered:

- [x] Added UI for goods nomenclatures which derive from the missing goods nomenclatures
- [x] Add validity periods polymorphic relationship

### Why?

I am doing this because:

- This is required to enable users to see where a goods nomenclature has moved to

Screenshot 2023-02-13 at 15 42 46


app/views/shared/_derived_goods_nomenclature.html.erb
```diff
@@ -0,0 +1,28 @@
+<% if deriving_goods_nomenclatures.count == 1 %>
```
willfish:
```suggestion
<% if deriving_goods_nomenclatures.one? %>
```

app/views/shared/_derived_goods_nomenclature.html.erb
```diff
@@ -0,0 +1,28 @@
...
+  <% deriving_goods_nomenclatures.each do |goods_nomenclature| %>
```
willfish: Each feels weird, here:

```suggestion
  <% deriving_goods_nomenclatures.first do |goods_nomenclature| %>
```

app/views/shared/_derived_goods_nomenclature.html.erb
```diff
@@ -0,0 +1,28 @@
...
+    <p>Goods previously classified under commodity <%= @commodity_code %>
+    were moved to code <%= link_to(goods_nomenclature.goods_nomenclature_item_id, polymorphic_path(goods_nomenclature)) %> 
+    (<%= goods_nomenclature.formatted_description %>) on <%= goods_nomenclature.validity_start_date&.to_formatted_s(:long) %></p>
+  <% end %>
+<% elsif deriving_goods_nomenclatures.count > 1 %>
```
willfish:
```suggestion
<% elsif deriving_goods_nomenclatures.many? %>
```

willfish:
I don't think you need the if check when rendering the partial on every page. You can move that check into the partial and its corresponding view spec and then the only test we need on the 404 views is to say the template was rendered

willfish:
LGTM :sheep: :it:

maverickvi (approved):

PR #1330, opened 2023-02-14
HOTT-2664: Disable switching on news items pages

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2664

### What?

I have added/removed/altered:

- [x] Disable service switching on the news items page

### Why?

I am doing this because:

- This is a change requested by Matt

### DIFF
--- a/app/controllers/news_items_controller.rb
+++ b/app/controllers/news_items_controller.rb
@@ -1,6 +1,7 @@
 class NewsItemsController < ApplicationController
   before_action :disable_search_form,
-                :disable_last_updated_footnote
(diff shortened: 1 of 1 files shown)


jebw (commented):
From the call this morning I thought it was decided this was meant to be just disabling for the #Show action

That was the one where matt was saying it doesn't make sense to switch service for? The user may want to swap from a listing of UK stories to XI stories?

willfish:
Spoke with Matt and he was happy with this @jebw

maverickvi (approved):

PR #1333, opened 2023-02-16
HOTT-2381: Simplify search references

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2381

### What?

I have added/removed/altered:

- [x] The referenced_id now returns something compatible for the az without any real adjustments

### Why?

I am doing this because:

- This is required since we've simplified the search references in the backend

### DIFF
--- a/.erb-lint.yml
+++ b/.erb-lint.yml
@@ -1,10 +1,2 @@
 ---
 EnableDefaultLinters: true
-linters:
-  ErbSafety:
-    enabled: true
-  Rubocop:
-    enabled: true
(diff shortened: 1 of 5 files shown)


.erb-lint.yml
```diff
@@ -1,10 +1,2 @@
 ---
 EnableDefaultLinters: true
-linters:
```
willfish, line 3: This configuration was actually breaking erblint

jebw (approved):
LGTM

PR #1334, opened 2023-02-16
HOTT-2652: Silence sentry noise for country scrapers

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2652
https://engine-le.sentry.io/issues/3929079904/?project=5557002&referrer=slack

### What?

I have added/removed/altered:

- [x] Removed sentry alerts for missing country flag
- [x] Removed sentry alerts for missing guide images

### Why?

I am doing this because:

- More signal, less noise
- The guide images are controlled by us on all fronts and if someone is scraping images of guides and they put in the wrong text we don't want to be alerted to that


jebw (approved):
:+1:

PR #1352, opened 2023-03-01
HOTT-2734: Handle keydown events for li items

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2734

### What?

I have added/removed/altered:

- [x] Added handling for hitting Enter/Tab on li items (suggestions) as well as Enter/Tab on the input

### Why?

I am doing this because:

- This fixes autosubmitting when a suggestion is selected

### DIFF
--- a/app/webpacker/src/javascripts/commodities.js
+++ b/app/webpacker/src/javascripts/commodities.js
@@ -798,15 +798,29 @@
                 $(element).parents('form').find('.js-commodity-picker-target').val($(ev.target).val());
               });
 
-              $(element).on('keydown', 'input[type="text"]', function(ev) {
-                if (ev.key == 'Enter') {
-                  ev.preventDefault() ;
+              // Avoid the default keydown behaviour of the autocomplete library and
+              // auto submit on Enter or Tab ourselves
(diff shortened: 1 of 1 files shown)


maverickvi (approved):

app/webpacker/src/javascripts/commodities.js
```diff
@@ -833,15 +847,7 @@
                   if (obj) {
                     $(element).parents('form:first').find('.js-commodity-picker-target').val(obj.id);
 
-                    $(document).trigger('tariff:chooseSearchQueryOption', [{
```
willfish, line 836: So this event just does not exist

app/webpacker/src/javascripts/commodities.js
```diff
@@ -833,15 +847,7 @@
...
-                      params: {
-                        data: obj
-                      }
-                    }]);
-
-                    if (typeof($(element).data('nosubmit')) == 'undefined') {
```
willfish, line 842: As I understand it, our preventing default behaviour is bypassing this. I'm going to revert it, though, just to avoid additional changes

PR #1366, opened 2023-03-09
HOTT-2774: Stop filtering out filtered facets

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2774

### What?

I have added/removed/altered:

- [x] Fixed issue with overzealous filtering of facets with a single classification

### Why?

I am doing this because:

- We use these facets to display any applied classifications

### DIFF
--- a/app/models/beta/search/facet_filter_statistic_collection.rb
+++ b/app/models/beta/search/facet_filter_statistic_collection.rb
@@ -29,8 +29,7 @@ module Beta
           facet_filter_statistic.facet_count,
         )
 
-        facet_filter_statistic.facet_classification_statistics.many? &&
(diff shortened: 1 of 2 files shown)


app/models/beta/search/facet_filter_statistic_collection.rb
```diff
@@ -29,8 +29,7 @@ def facet_filter_applicable?(facet_filter_statistic)
           facet_filter_statistic.facet_count,
         )
 
-        facet_filter_statistic.facet_classification_statistics.many? &&
```
willfish, line 32: If we require all facet filters to have multiple classifications then we can't see ones that we've clicked to filter by the classification in the UI and cause an error

maverickvi (approved):

PR #1373, opened 2023-03-15
HOTT-2729: Display validity dates on commodity and declarable heading…

maverickvi:
…s pages

### Jira link

[HOTT-<TODO>](https://transformuk.atlassian.net/browse/HOTT-2729)

### What?

I have added/removed/altered:

- [ ] Displayed validity dates on commodity and declarable headings pages

### Why?

I am doing this because:

- Displaying these dates on the frontend will be useful for TT users

Screenshot 2023-03-15 at 11 35 20

Screenshot 2023-03-15 at 11 35 41

### DIFF
--- a/app/views/shared/context_tables/_commodity.html.erb
+++ b/app/views/shared/context_tables/_commodity.html.erb
@@ -27,7 +27,23 @@
     </dd>
     <dd class="govuk-summary-list__actions"></dd>
   </div>
-  <%= commodity_validity_dates(declarable) %>
+  <% if declarable.validity_start_date.present? && declarable.validity_end_date.present? %>
+    <div class="govuk-summary-list__row">
+      <dt class="govuk-summary-list__key">Commodity valid between</dt>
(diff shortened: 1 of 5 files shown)


app/views/shared/context_tables/_commodity.html.erb
```diff
@@ -17,6 +17,7 @@
     </dd>
     <dd class="govuk-summary-list__actions"></dd>
   </div>
+  <%= commodity_validity_dates(declarable) %>
```
willfish: Not saying your approach is wrong, but shouldn't we just be consistent with the other rows?

willfish: Maybe its worth refactoring them to be consistent with the helper approach? And maybe we should add a new helper for the goods nomenclature context summary lists?

willfish:
Smoke tests need updating?

willfish (approved):

PR #1374, opened 2023-03-15
HOTT-2849: Show accumulated excise units when no supplementary unit

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2853

image

### What?

I have added/removed/altered:

- [x] Added support for showing excise measurement units in supplementary unit row

### Why?

I am doing this because:

- This is required to better flag to traders that they need to input excise measurement units

### DIFF
--- a/app/models/concerns/declarable.rb
+++ b/app/models/concerns/declarable.rb
@@ -78,17 +78,16 @@ module Declarable
   end
 
   def supplementary_unit_description
-    supplementary_measures.first&.supplementary_unit_description || 'Supplementary unit'
+    import_measures.supplementary.first&.supplementary_unit_description || 'Supplementary unit'
   end
 
   def supplementary_unit
-    measure = supplementary_measures.first
-    supplementary_component = measure&.measure_components&.first
-
-    if supplementary_component
-      "#{supplementary_component.measurement_unit&.description} (#{measure.duty_expression&.base})"
+    if supplementary_component.present?
+      supplementary_component.unit_for_classification
+    elsif supplementary_excise_measures?
+      excise_unit_for_classification
     else
-      'No supplementary unit required.'
+      no_supplementary_unit_classification
     end
   end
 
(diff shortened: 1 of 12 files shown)


app/models/concerns/declarable.rb
```diff
@@ -104,8 +103,28 @@ def has_safeguard_measure?
...
   private
 
-  def supplementary_measures
-    @supplementary_measures ||= import_measures.supplementary
+  def supplementary_component
+    @supplementary_component ||= import_measures.supplementary.first&.measure_components&.first
+  end
+
+  def supplementary_excise_measures?
+    import_measures.excise.erga_omnes.any? && import_measures.excise.erga_omnes.any?(&:measurement_units?)
```
jebw: I think the first `any?` is unnecessary, if theren't aren't any then the second expression wont try to call `measurement_units?` and will just return false

If there are some `erga_omnes` then it'll end up running the second `any?` anyway?

app/models/concerns/declarable.rb
```diff
@@ -104,8 +103,28 @@ def has_safeguard_measure?
...
+  end
+
+  def excise_unit_for_classification
+    excise_units = import_measures.excise.erga_omnes.each_with_object({}) do |measure, acc|
+      measure.measure_components.each do |component|
+        acc[component.measurement_unit.resource_id] = component.measurement_unit.description.downcase
```
jebw: Is unit description ever going to be nil? Eg because of missing descriptions on the backend? Should this be `&.downcase`

jebw (approved):
LGTM :+1:  - couple of minor comments

app/models/concerns/declarable.rb
willfish: Good shout

app/models/concerns/declarable.rb
willfish: I mean its technically possible but these are never nil in my experience. I might just add a to_s

PR #1376, opened 2023-03-16
HOTT-2841-2842: Updated the CSS on commodities page

maverickvi:
### Jira link

[HOTT-<TODO>
](https://transformuk.atlassian.net/browse/HOTT-2841)
https://transformuk.atlassian.net/browse/HOTT-2842

### What?

I have added/removed/altered:

- [ ] Updated CSS on commodities page

### Why?

I am doing this because:

- we need to make improvements to the page design

Screenshot 2023-03-16 at 13 17 56

### DIFF
--- a/app/webpacker/src/stylesheets/_commodity-tree.scss
+++ b/app/webpacker/src/stylesheets/_commodity-tree.scss
@@ -571,6 +571,3 @@ article.tariff {
   white-space: nowrap;
 }
 
-.quotas-blurb {
-  font-size: 16px;
-}
--- a/app/webpacker/src/stylesheets/_commodity-tree.scss
+++ b/app/webpacker/src/stylesheets/_commodity-tree.scss
@@ -570,4 +570,3 @@ article.tariff {
 .non-breaking-heading {
   white-space: nowrap;
 }
-
--- a/app/webpacker/src/stylesheets/_commodity-tree.scss
+++ b/app/webpacker/src/stylesheets/_commodity-tree.scss
@@ -570,3 +570,7 @@ article.tariff {
 .non-breaking-heading {
   white-space: nowrap;
 }
+
+.quotas-blurb {
+  font-size: 16px;
+}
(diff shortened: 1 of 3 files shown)


app/webpacker/src/stylesheets/_commodity-tree.scss
```diff
@@ -570,3 +570,7 @@ article.tariff {
 .non-breaking-heading {
   white-space: nowrap;
 }
+
+.quotas-blurb {
```
willfish: We have so much special snowflake css styles :(

willfish (approved):

app/webpacker/src/stylesheets/_commodity-tree.scss
jebw: This class duplicates what is already provided by the GDS design system  - use `govuk-!-font-size-16`

and I'd recommend reading through the design system docs to learn whats already provided by our design system

jebw: Highly specific classes like this should be a real edgecase we try to avoid at all costs

If we need custom css, we should have it as part of a generic element, eg something like `small-text`

_but_ that actuall already exists and has a font-size of 16px, its `govuk-body-s`

see https://design-system.service.gov.uk/styles/typography/#body-small

jebw (changes requested):

app/webpacker/src/stylesheets/_measure-inset.scss
```diff
@@ -27,6 +27,10 @@
     li:not(:last-child) {
       margin-bottom: 5px;
     }
+
+    ul {
+      padding-left: 27px;
```
jebw, line 32: Not  sure how specific this sizing needs to be but if it can flex by a couple of pixels you could use whats provided by the design system since they have padding classes for 25px?

maverickvi: 25px doesn't align properly so I have to stick with 27px

PR #1377, opened 2023-03-17
HOTT-2885: Remove basic auth now that we're rate limiting

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2885

### What?

I have added/removed/altered:

- [x] Removed basic auth capability now that we rate limit

### Why?

I am doing this because:

- I have no intention of ever enabling this again


jebw (commented):
Personally I think we should keep it - we've already pulled it out once then put it back in. GDS docs say that non-prod apps should be restricted I believe - I doubt they'd ever care but this does go against that.

If we find we do want to restrict access to dev/staging for any purpose this easily enables that with minimal code

willfish:
Fair enough @jebw. I'll close this in favour of just changing configuration

PR #1386, opened 2023-03-22
HOTT-2909: Client side cookie management

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2909

### What?

I have added/removed/altered:

- [x] Adds a cookies manager class and specs
- [x] Adds cookie banner controller and specs
- [x] Plumb together cookie banner in view
- [x] Adds policy controller and specs
- [x] Plumb together policy controller in view
- [x] Use client-side cookies with google tag manager
- [x] Use client side cookies with tree
- [x] Remove server-side implementation
- [x] Adds eslint rules
- [x] Adds js-cookie and other helpful packages
- [x] Updates babel config to precompile private methods

### Why?

I am doing this because:

- This is how we should have done this originally and enables us to just configure the CDN to cache responses


config/routes.rb
```diff
@@ -85,11 +85,8 @@
   resource :trading_partners, only: %i[show update]
 
   namespace :cookies do
-    resource :policy, only: %i[show create update]
-    resource :hide_confirmation, only: %i[create]
+    resource :policy, only: %i[show]
   end
-  resolve('Cookies::Policy') { %i[cookies policy] }
-  get 'cookies', to: redirect(path: '/cookies/policy')
```
willfish, line 92: Wondering why this was added and whether we need this.

maverickvi (approved):

PR #1393, opened 2023-03-28
BAU: Fixes deploys by using an s3 repo

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [x] Fixes issue with deployment by using s3 bucket to install cf

### Why?

I am doing this because:

- This api is down


.circleci/config.yml
```diff
@@ -23,7 +40,8 @@ commands:
         type: string
     steps:
       - checkout
-      - tariff/cf-install:
+      - aws-cli/install
```
jebw: Not needed anymore but not a problem either

jebw (approved):
LGTM

PR #1394, opened 2023-04-03
HOTT-2930: Show only descendants of a subheading

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2930

**Before**

image

**After**

image

### What?

I have added/removed/altered:

- [x] Altered the commodities we pass to the commodity tree to exclude the subheading itself

### Why?

I am doing this because:

- This was noted as a bit confusing for users and the subheading is clearly visible in the classification

### DIFF
--- a/app/models/subheading.rb
+++ b/app/models/subheading.rb
@@ -21,6 +21,16 @@ class Subheading < GoodsNomenclature
                 :formatted_description,
                 :declarable
 
+  def descendants
+    included_subheading = commodities.find { |c| c.goods_nomenclature_sid == goods_nomenclature_sid }
+
+    if included_subheading.present?
+      included_subheading.children
+    else
+      []
+    end
+  end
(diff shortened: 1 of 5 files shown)


app/models/subheading.rb
```diff
@@ -21,6 +21,16 @@ class Subheading < GoodsNomenclature
                 :formatted_description,
                 :declarable
 
+  def descendants
+    included_subheading = commodities.find { |c| c.goods_nomenclature_sid == goods_nomenclature_sid }
+
+    if included_subheading.present?
```
jebw: this could just be a ternary? or even safe navigation, eg

```
included_subheading ? included_subheading.children : []
```

```
included_subheading&.children || []
```

jebw (approved):
LGTM - small code suggestion but very minor

PR #1406, opened 2023-04-04
HOTT-2984:  Sanitise stop press titles and content titles

maverickvi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2984

### What?

I have added/removed/altered:

- [x] Sanitised stop press titles and content titles

### Why?

I am doing this because:

- We want to escape dangerous html and allow safe html

### Have you? (optional)

Screenshot 2023-04-04 at 14 31 13

### DIFF
--- a/spec/factories/news/item_factory.rb
+++ b/spec/factories/news/item_factory.rb
@@ -76,13 +76,11 @@ FactoryBot.define do
     end
 
     trait :with_unsafe_html do
-      unsafe_html = "<script>alert('Hello, world!')</script>"
-
-      title { "Title #{unsafe_html}" }
+      title { "Title <script>alert('Hello, world!')</script>" }
 
       content do
         <<~CONTENT
-          ## Second heading #{unsafe_html}
+          ## Second heading <script>alert('Hello, world!')</script>
 
           This is a paragraph
         CONTENT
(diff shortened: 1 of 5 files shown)


spec/factories/news/item_factory.rb
```diff
@@ -74,5 +74,33 @@
         CONTENT
       end
     end
+
+    trait :with_unsafe_html do
+      unsafe_html = "<script>alert('Hello, world!')</script>"
```
jebw: This isn't really needed and tbh I'm not entirely sure the scope these variables are getting set in - I think its fine to duplicate this short string in the content and title attributes

spec/factories/news/item_factory.rb
```diff
@@ -74,5 +74,33 @@
...
+        <<~CONTENT
+          ## Second heading #{unsafe_html}
+
+          This is a paragraph
+        CONTENT
+      end
+    end
+
+    trait :with_safe_html do
+      safe_html = '<abbr>MFN</abbr>'
```
jebw: Likewise I think this is a bit uncommon, I'd just copy the `safe_html` contents into the other two strings

jebw (changes requested):
Looks great, just a minor tweak for the factory changes

jebw (approved):
LGTM :raised_hands:

willfish (changes requested):
How comes so much is changing in this PR?

maverickvi:
> 

The issue with how Github was displaying commits has been resolved by their team.

PR #1417, opened 2023-04-18
HOTT-3034: Adds labels to suggestions

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3034
https://github.com/trade-tariff/trade-tariff-backend/pull/1188

image
image

### What?

I have added/removed/altered:

- [x] Added labels to suggestion types
- [x] Added test coverage for the new labels

### Why?

I am doing this because:

- This makes the meaning of the suggestions clearer for users

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -99,7 +99,14 @@ class SearchController < ApplicationController
   def suggestions
     search_term = Regexp.escape(params[:term].to_s.strip)
     matched_suggestions = SearchSuggestion.all(q: search_term)
-    results = matched_suggestions.map { |s| { id: s.value, text: s.value, query: s.query } }
+    results = matched_suggestions.map do |s|
+      {
+        id: s.value,
+        text: s.value,
+        query: s.query,
+        formatted_suggestion_type: s.formatted_suggestion_type,
+      }
+    end
 
     render json: { results: }
   end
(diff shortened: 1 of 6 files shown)


app/controllers/search_controller.rb
```diff
@@ -99,7 +99,14 @@ def toggle_beta_search
   def suggestions
     search_term = Regexp.escape(params[:term].to_s.strip)
     matched_suggestions = SearchSuggestion.all(q: search_term)
-    results = matched_suggestions.map { |s| { id: s.value, text: s.value, query: s.query } }
+    results = matched_suggestions.map do |s|
+      {
+        id: s.value,
+        text: s.value,
+        query: s.query,
+        formatted_suggestion_type: s.formatted_suggestion_type,
```
willfish, line 107: So in js, we can set this value to an empty string or null and the result will be false/we won't show the labels. This highlights what I've tested in the UI that basically we're not going to show labels when the suggestion type is absent (e.g. if the backend isn't sufficiently deployed/the suggestions haven't been updated to include the type yet) 
```node
> !!{"formatted_suggestion_type": null}.formatted_suggestion_type
false
> !!{"formatted_suggestion_type": ""}.formatted_suggestion_type
false
> !!{}.formatted_suggestion_type
false
```

maverickvi (approved):

PR #1421, opened 2023-04-20
HOTT-3027: Display standard rate text on VAT measures without additio…

maverickvi:
…nal code

### Jira link

https://transformuk.atlassian.net/browse/HOTT-3027

### What?

I have added/removed/altered:

- [ ] Displayed standard rate text on VAT measures without additional code

### Why?

I am doing this because:

- This will make it clearer for users to understand which measures are "standard rate" measures

Screenshot 2023-04-20 at 13 23 34

### DIFF
--- a/app/views/measures/_measure.html.erb
+++ b/app/views/measures/_measure.html.erb
@@ -25,13 +25,14 @@
           declarable_id: declarable.short_code %>
       </span>
     <% end %>
-
     <% if measure.additional_code.present? %>
       <% if measure.measure_type.prohibitive? %>
         <%= render 'measures/additional_codes/prohibitive', measure: measure %>
       <% else %>
         <%= render 'measures/additional_codes/regular', measure: measure %>
       <% end %>
+    <% elsif !measure.additional_code.present? && measure.vat? %>
+      <span class="table-line">Standard rate</span>
     <% end %>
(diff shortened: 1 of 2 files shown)


app/views/measures/_measure.html.erb
```diff
@@ -25,13 +25,14 @@
           declarable_id: declarable.short_code %>
       </span>
     <% end %>
-
```
jebw, line 28: Why compress these blocks together? Seems like it reduces rather then increases readability?

jebw (approved):
LGTM

app/views/measures/_measure.html.erb
```diff
@@ -25,13 +25,14 @@
...
     <% if measure.additional_code.present? %>
       <% if measure.measure_type.prohibitive? %>
         <%= render 'measures/additional_codes/prohibitive', measure: measure %>
       <% else %>
         <%= render 'measures/additional_codes/regular', measure: measure %>
       <% end %>
+    <% elsif !measure.additional_code.present? && measure.vat? %>
```
willfish, line 34: Possibly this is my misunderstanding but aren't their 3 types of VAT additional code?

I don't think they're all meant to be labelled `Standard Rate`

PR #1424, opened 2023-04-21
HOTT-3066: Fixes bug in additional code table

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3066

### What?

I have added/removed/altered:

- [x] Fixes bug in additional codes table
- [x] Adds some coverage for null_object
- [x] Adds a bunch of business logic methods for understanding additional code measures
- [x] Extract, refactor and test overview measure behaviour

### Why?

I am doing this because:

- This is required to make sure we're handling situations where there are multiple third country measures of different types that are purely non-additional code measures

### DIFF
--- a/app/models/measure_type.rb
+++ b/app/models/measure_type.rb
@@ -28,6 +28,10 @@ class MeasureType
     id.in?(SUPPLEMENTARY_MEASURE_TYPES)
   end
 
+  def mfn?
+    id == '103'
+  end
+
   def description
     translated_description || attributes['description']
   end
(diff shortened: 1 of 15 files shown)


app/models/measure_type.rb
```diff
@@ -28,6 +28,10 @@ def supplementary?
     id.in?(SUPPLEMENTARY_MEASURE_TYPES)
   end
 
+  def mfn?
```
willfish: This is a bit of a terrible name. Maybe `mfn_no_authorised_use?`

jebw (approved):
LGTM

PR #1418, opened 2023-04-18
HOTT-2307: Enable markdown in commodity code descriptions

maverickvi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-2307

### What?

I have added/removed/altered:

- [x] Enabled markdown in commodity code descriptions

### Why?

I am doing this because:

- We were incorrectly displaying markdown

Screenshot 2023-04-18 at 16 38 02

PR #1429, opened 2023-04-27
Revert "HOTT-2307: Enable markdown in commodity code descriptions"

willfish:
Reverts trade-tariff/trade-tariff-frontend#1418


willfish:
Talking this over with Matt there are a few issues with this:

1. We haven't fixed a bunch of cases
2. The formatting needs to live in the backend and only the govspeak implementation should do rendering in the frontend.

PR #1443, opened 2023-05-09
BAU: Fixes failing specs

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [x] Fixes broken atom search specs

### Why?

I am doing this because:

- These are missing a request in the VCR cassette


jebw (approved):
LGTM

PR #1446, opened 2023-05-11
HOTT-3166 Garbage collect wizard sessions

jebw:
### Jira link

HOTT-3166

### What?

I have added/removed/altered:

- [x] Added a custom wizard store which garbage collects old journeys

### Why?

I am doing this because:

- We are overflowing the cookie size because our sessions are long lived

- Removes user session data

### DIFF
--- a/app/models/rules_of_origin/wizard_store.rb
+++ b/app/models/rules_of_origin/wizard_store.rb
@@ -3,6 +3,17 @@ module RulesOfOrigin
     PREFIX = 'rules_of_origin-'.freeze
     MAX_JOURNEYS = 5
 
+    class << self
+      def gc_store_data(backing_store)
+        backing_store.keys
+                     .select { |k| k.start_with? PREFIX }
+                     .sort_by { |k| backing_store[k]['modified'] || '2022-01-01T00:00:00Z' }
+                     .reverse
+                     .slice(MAX_JOURNEYS..)
+                     &.each { |k| backing_store.delete k }
+      end
+    end
+
     def initialize(backing_store, key)
       @backing_store = backing_store
       @backing_store["#{PREFIX}#{key}"] ||= {}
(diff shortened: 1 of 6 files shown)


app/models/rules_of_origin/wizard_store.rb
```diff
@@ -0,0 +1,30 @@
...
+        backing_store.keys
+                     .select { |k| k.start_with? PREFIX }
+                     .sort_by { |k| backing_store[k]['modified'] || '2022-01-01T00:00:00Z' }
+                     .reverse
+                     .slice(MAX_JOURNEYS..)
+                     &.each { |k| backing_store.delete k }
+      end
+    end
+
+    def initialize(backing_store, key)
```
willfish, line 17: Shouldn't this be in the normal scope of the class?

app/models/rules_of_origin/wizard_store.rb
```diff
@@ -0,0 +1,30 @@
...
+    MAX_JOURNEYS = 5
+
+    class << self
+      def gc_store_data(backing_store)
+        backing_store.keys
+                     .select { |k| k.start_with? PREFIX }
+                     .sort_by { |k| backing_store[k]['modified'] || '2022-01-01T00:00:00Z' }
+                     .reverse
+                     .slice(MAX_JOURNEYS..)
+                     &.each { |k| backing_store.delete k }
```
willfish, line 13: I suspect & isn't needed, here?

willfish (approved):
LGTM

app/models/rules_of_origin/wizard_store.rb
jebw: :eyes:

app/models/rules_of_origin/wizard_store.rb
jebw: after discussion its not an issue

app/models/rules_of_origin/wizard_store.rb
jebw: Unfortunately slice returns nil, not an empty array if no matches

```
Loading development environment (Rails 7.0.4.3)
[1] pry(main)> [1,2,3,4].slice(5..)
=> nil
[2] pry(main)> 
```

PR #1451, opened 2023-05-15
HOTT-3197 Don't match bad urls to RoO wizard

jebw:
### Jira link

HOTT-3197

### What?

I have added/removed/altered:

- [x] Added constraints to the url mapping for the rules of origin wizard

### Why?

I am doing this because:

- Incorrect urls under `/rules_of_origin/` always got routed to the wizard, which then did an invalid redirect on the incomplete url

- Low, should only impact users following invalid urls

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -95,12 +95,13 @@ Rails.application.routes.draw do
   resources :news_items, only: %i[index show], path: '/news'
 
   namespace :rules_of_origin, path: nil do
-    get '/rules_of_origin/:commodity/:country', to: 'steps#index', as: :steps
-    get '/rules_of_origin/:commodity/:country/:id', to: 'steps#show', as: :step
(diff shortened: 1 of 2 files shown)


config/routes.rb
```diff
@@ -95,12 +95,13 @@
   resources :news_items, only: %i[index show], path: '/news'
 
   namespace :rules_of_origin, path: nil do
-    get '/rules_of_origin/:commodity/:country', to: 'steps#index', as: :steps
-    get '/rules_of_origin/:commodity/:country/:id', to: 'steps#show', as: :step
-    patch '/rules_of_origin/:commodity/:country/:id', to: 'steps#update', as: nil
-    get '/rules_of_origin/:commodity', constraints: { commodity: /\d{10}/ },
-                                       to: 'product_specific_rules#index',
-                                       as: :product_specific_rules
+    with_options constraints: { commodity: /\d{10}/ } do
```
willfish, line 98: TIL

willfish (approved):

PR #1480, opened 2023-06-09
HOTT-3336 Show cds proofs info on measures overlay

jebw:
### Jira link

HOTT-3336

### What?

I have added/removed/altered:

- [x] Show CDS proof info on measures tab for relevant measures

### Why?

I am doing this because:

- If helps our users submit valid declarations

### Have you? (optional)

- Adds a call to the all rules of origin schemes endpoint on the commodities page

### Screenshots

Screenshot from 2023-06-09 11-36-21
Screenshot from 2023-06-09 11-36-28
image

### DIFF
--- a/app/models/measure_type.rb
+++ b/app/models/measure_type.rb
@@ -2,6 +2,7 @@ require 'api_entity'
 
 class MeasureType
   RENDERED_MEASURE_TYPE_DETAILS = { 'xi' => {}, 'uk' => {} }.freeze
+  CDS_PROOFS_OF_ORIGIN = %w[141 142 143 145 146 147].freeze
 
   delegate :service_name, to: TradeTariffFrontend::ServiceChooser
 
@@ -46,6 +47,10 @@ class MeasureType
     return 'Restriction' if description.scan(/Restriction/i).present?
 
     'Control'
+  end
+
+  def cds_proofs_of_origin?
+    CDS_PROOFS_OF_ORIGIN.include? id
   end
 
   private
--- a/app/models/rules_of_origin/scheme.rb
+++ b/app/models/rules_of_origin/scheme.rb
@@ -56,8 +56,6 @@ class RulesOfOrigin::Scheme
   end
 
   def applies_to_geographical_area?(area)
-    area_code = area.is_a?(GeographicalArea) ? area.geographical_area_id : area
-
-    countries.include? area_code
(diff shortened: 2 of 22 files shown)


app/models/measure_type.rb
```diff
@@ -48,6 +49,10 @@ def abbreviation
     'Control'
   end
 
+  def cds_proofs_of_origin?
+    CDS_PROOFS_OF_ORIGIN.include? id
```
willfish, line 53: I'm wondering if this should be `resource_id`?

app/models/rules_of_origin/scheme.rb
```diff
@@ -54,4 +54,10 @@ def proof_codes
   def cds_proof_info?
     proof_intro.present? || proof_codes&.any?
   end
+
+  def applies_to_geographical_area?(area)
+    area_code = area.is_a?(GeographicalArea) ? area.geographical_area_id : area
```
willfish: I think you can just treat this like a duck and remove the defensiveness?

willfish: Ah, you're not actually using the area as area_code variant so can probably just remove it?

app/models/measure_type.rb
jebw: hmm - or possibly `measure_type_id` - I'll change it :+1:

app/models/rules_of_origin/scheme.rb
jebw: Ok

spec/factories/measure_factory.rb
```diff
@@ -16,7 +16,7 @@
     end
 
     geographical_area do
-      attributes_for(:geographical_area, id: geographical_area_id)
+      attributes_for(:geographical_area, id: geographical_area_id, geographical_area_id:)
```
willfish, line 19: The api responding with 3 geo area ids is a bit mad but clearly historical

jebw: Yeh, just figured I'd fix up the factory - confused me initially

willfish (dismissed):

app/models/measure_type.rb
jebw: Nah - this needs to be `id` - `resource_id` is kinda internal and there isn't a `measure_type_id`

maverickvi (approved):

PR #1482, opened 2023-06-12
HOTT-3364: XI notes links go to XI instead of UK site

maverickvi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3364

### What?

I have added/removed/altered:

- [ ] XI notes links go to XI instead of UK site

### Why?

I am doing this because:

- XI notes were linking to UK

Screenshot 2023-06-12 at 16 39 17

### DIFF
--- a/app/helpers/service_helper.rb
+++ b/app/helpers/service_helper.rb
@@ -109,7 +109,7 @@ module ServiceHelper
   def insert_service_links(note)
     return note if uk_service_choice? || note.nil?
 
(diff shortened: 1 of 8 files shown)


app/helpers/service_helper.rb
```diff
@@ -106,6 +106,12 @@ def replace_service_tags(content)
     end
   end
 
+  def insert_service_links(note)
```
willfish: Can't we just add this at the point of generating links?

willfish (approved):

PR #1490, opened 2023-06-21
HOTT-1355 Upgrade Faraday to v2

jebw:
### Jira link

HOTT-1355

### What?

I have added/removed/altered:

- [x] Upgraded to last v1 version
- [x] Added new retry gem
- [x] Upgraded to v2 and dropped middleware gem

### Why?

I am doing this because:

- We want to use a current version 
- We want to subsequently add the faraday caching plugin

- Significant - last time this was attempted it caused performance problems but they were never profiled so not sure what I should be addressing. Will have hotfix to rollback on hand


willfish (approved):
Eeeeeeeeeeeeeeeeeeeeeeeeeeeeee

PR #1493, opened 2023-06-23
HOTT-3370 Add caching to api client

jebw:
### Jira link

HOTT-3370

### What?

I have added/removed/altered:

- [x] Added Faraday-http-cache plugin
- [x] Dropped existing manual caching in redis

### Why?

I am doing this because:

- Any http request which is in our cache not older then the cache expiry using the cached copy without ever contacting the backend
  - cache expiry is provided by the backend, this is anticipated to be 5 minutes
- Any request which is in the cache but older then the cache expiry will result in a query to the backend but if the backend determines the cached response is still fresh, then the frontend will receive a very fast 304 response and continue to use its cached copy. Resetting the cache expiry
- Any request which the backend determines to no longer be fresh will be returned as normal but cached to avoid future repeat requests

- Enables caching - pages on the frontend will potentially be 5 minutes out of date with the backends data


willfish (approved):
I love this :heart:

PR #1501, opened 2023-06-29
Fix title tag on chapters to use non capitalized version

shjohnson:
https://transformuk.atlassian.net/browse/HOTT-3521

### What?

I have added/removed/altered:

- [ x] Changed the `description_plain` too return the `formatted_description`

### DIFF
--- a/app/helpers/service_helper.rb
+++ b/app/helpers/service_helper.rb
@@ -6,7 +6,7 @@ module ServiceHelper
   def goods_nomenclature_title(goods_nomenclature)
     t(
       'title.goods_nomenclature',
-      goods_description: goods_nomenclature.description_plain,
+      goods_description: goods_nomenclature.formatted_description,
       service_name:,
     )
   end
--- a/app/models/section.rb
+++ b/app/models/section.rb
@@ -10,7 +10,7 @@ class Section
                 :title,
                 :chapters,
                 :section_note,
-                :description_plain
+                :formatted_description
 
   attr_reader :chapter_to,
               :chapter_from
--- a/spec/helpers/service_helper_spec.rb
+++ b/spec/helpers/service_helper_spec.rb
@@ -44,7 +44,7 @@ RSpec.describe ServiceHelper, type: :helper do
   end
 
   describe '#goods_nomenclature_title' do
-    let(:commodity) { build(:commodity, description_plain: 'Live horses, asses, mules and hinnies') }
+    let(:commodity) { build(:commodity, formatted_description: 'Live horses, asses, mules and hinnies') }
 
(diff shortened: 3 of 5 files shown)


app/helpers/service_helper.rb
```diff
@@ -6,7 +6,7 @@ def default_title
   def goods_nomenclature_title(goods_nomenclature)
     t(
       'title.goods_nomenclature',
-      goods_description: goods_nomenclature.description_plain,
+      goods_description: goods_nomenclature.formatted_description,
```
willfish, line 9: So I think the history of this is:

1. Ata added a new attribute to the backend which removed certain characters from the description that shows in the title (he chose to call this `description_plain`)
2. He updated the frontend to reflect this for all titles for all goods nomenclature
3. For some reason its not working specifically for headings
4. Sections seem to be unaffected as far as I can see on staging

So I think we still need `description_plain` everywhere but could do with knowing why it isn't showing properly for headings

Looking at the v2 api response from the backend (which is what the frontend consumes) I can see the `description_plain` attribute is missing https://dev.trade-tariff.service.gov.uk/api/v2/headings/0101

So my best guess is that simply updating the backend application to properly surface the `description_plain` attribute should resolve this issue on the frontend

The method is implemented on the parent class of all goods nomenclature (Chapter, Heading, Subheading, Commodity are all children of this class) https://github.com/trade-tariff/trade-tariff-backend/blob/main/app/models/goods_nomenclature.rb#L30-L31C45 and surfaced to the api resource via https://github.com/trade-tariff/trade-tariff-backend/blob/main/app/serializers/api/v2/headings/heading_serializer.rb#L11-L16

So we should safely be able to add `description_plain` to the list of attributes :)

shjohnson: Thanks Will, i can see that its missing like you described below but it would still mean that i have to change how the description_plain is being presented as currently its all in capitals right?

shjohnson: I think the worry with changing the API was more that if something is consuming it and expects capitals then it could cause issues

willfish: Good shout re the api users expecting the capitalised form. Technically the description_plain field actually predates Ata's change and I believe he merely extended it. I don't know whether people in the api expect it and have no way of actually knowing sadly so we'd need to handle the capitalisation client side.

We will want to make sure that the Heading, Subheading and Commodity goods nomenclatures do use the description plain since the alternative is having weird characters in the titles and at least 11468 goods nomenclatures on the XI service.

shjohnson:
Closed in favour of this backend change: https://github.com/trade-tariff/trade-tariff-backend/pull/1365

PR #1503, opened 2023-06-29
HOTT-3513: Add build job for new AWS environment

amberstarlight:
### Jira link

[HOTT-3513](https://transformuk.atlassian.net/browse/HOTT-3513)

### What?

I have added/removed/altered:

- Added a new CircleCI job to build and push images to the new AWS ECR repository.
- Broken this out into a shell script, which I've checked against `shellcheck`.

### Why?

- Multiple things needed to change here:
  - moving away from using the `dev-build` parameter (i.e. just using the short SHA for the tag), and
  - having to use a different context for the new AWS ECR repository (as it's on a different account).

### Other notes

I would have liked to add a `.pre-commit-config.yml` to this repository, but even running the basic things like end-of-file-fixer made changes to a staggering number of files across the repository. That resulted in a _giant_ diff that would take a significant amount of effort to review.

I think we should make an ADR (and, ostensibly, a new ticket on the backlog) to get this implemented at some stage for consistency across the repo. We could break it up per subdirectory in several smaller PRs, or do it all at once in a co-ordinated strike.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -372,6 +372,25 @@ jobs:
             docker tag $IMAGE_NAME:$DOCKER_TAG $ECR_REPO/$IMAGE_NAME:$DOCKER_TAG
             docker push $ECR_REPO/$IMAGE_NAME:$DOCKER_TAG
 
+  # new aws env
+  build_and_push:
+    environment:
+      IMAGE_NAME: tariff-frontend
+    parameters:
+      environment:
+        type: string
+    docker:
+      - image: cimg/ruby:3.2.2-node
+    steps:
+      - checkout
+      - setup_remote_docker:
+          version: 20.10.11
+          docker_layer_caching: false
+      - aws-cli/install
+      - run:
+          name: Build and push image
+          command: ./.circleci/ecr.sh << parameters.environment >>
+
   smoketest_dev:
     docker:
       - image: "cypress/base:16.5.0"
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -372,6 +372,25 @@ jobs:
...
 
+  # new aws env
+  build_and_push:
+    environment:
+      IMAGE_NAME: tariff-frontend
+    parameters:
+      environment:
+        type: string
+    docker:
+      - image: cimg/ruby:3.2.2-node
```
willfish, line 383: I wonder if we could move to a smaller image?

willfish (dismissed):
LGTM

amberstarlight:
So after looking at this again, I've realised that we do need separate AWS ECR repositories for each image; there's only one namespace available.

Thus, I've [opened a PR](https://github.com/trade-tariff/trade-tariff-platform-aws-terraform/pull/43) on the Terraform repo that will create these. Once that's merged, I can revisit _this_ PR and fix the error.

.circleci/config.yml
amberstarlight: It has to build the application - `cimg/ruby:3.2.2-node` happens to have all the binaries we need for doing that. Is there any other image we could use to build?

willfish: Ah k

willfish (approved):
Feel free to ignore the test failures. They are totally unrelated to your change. I'll fix these separately

PR #1505, opened 2023-07-03
HOTT-3512: Add service-level Terraform

amberstarlight:
### Jira link

[HOTT-3512](https://transformuk.atlassian.net/browse/HOTT-3512)

### What?

I have added/removed/altered:

- Added the service-level Terraform to this repository, under `terraform/`.
- Added a pre-commit file under `terraform/` which is scoped to this subdirectory only.

### Why?

I am doing this because:

- We're now at the point where we can add the Terraform itself to control the service deployment and sub-resources.

> **Note**
> This PR **_DOES NOT_** add pipeline changes to deploy this into any environments, it is just the Terraform itself.

### Extra Info

When this is merged, I will add `[no ci]` or equivalent so that it does not trigger any pipeline, as there's no changes to make.


terraform/locals.tf
```diff
@@ -0,0 +1,9 @@
+locals {
+  api_service_backend_url_options = {
+    uk = "https://${var.base_domain}/uk/"
+    xi = "https://${var.base_domain}/xi"
+  }
+
+  # TODO: Revisit this when PaaS is no more
```
willfish: So this is actually a hard requirement coming from the plek gem which is a nested dependency of one of our other govuk gems https://github.com/alphagov/plek

willfish: So we'll never get rid of it :(

terraform/main.tf
```diff
@@ -0,0 +1,136 @@
...
+
+  docker_image = data.aws_ssm_parameter.ecr_url.value
+  docker_tag   = var.docker_tag
+  skip_destroy = true
+
+  service_environment_config = [
+    # TODO: discuss this internally via HOTT-3555
+    {
+      name  = "API_SERVICE_BACKEND_OPTIONS"
+      value = jsonencode(local.api_service_backend_url_options)
```
willfish, line 26: Yeah, I guess this will get discovered later.

terraform/locals.tf
amberstarlight: Makes sense. I'll remove the TODO 😢

willfish (approved):
Excited to get this rolling out :tada:

PR #1515, opened 2023-07-05
HOTT-3514: Add CI jobs to deploy Terraform

amberstarlight:
### Jira link

[HOTT-3514](https://transformuk.atlassian.net/browse/HOTT-3514)

### What?

I have added/removed/altered:

- Heavily modifed the CircleCi `config.yml`:
  - Removed jobs/commands that have been ported across to the `trade-tariff` Orb
  - Added `terraform` jobs

### Why?

I am doing this because:

- The time has come where we can start to deploy this app to the new AWS environment.

### Notes

So far I've been adding `[ci skip]` to all my pushes on this branch, so as to avoid redeploying the app several times. I'll want to do a run on this before we merge to see what the pipeline looks like from a branch that isn't main.

It's likely that we actually want to make changes to this - as in, we may want to deploy over development for every PR, but I've had my eyes full of YAML for the past three days and I can no longer think about them in great detail without a break from it.


.circleci/config.yml
```diff
@@ -447,80 +398,74 @@ jobs:
...
-    steps:
-      - cf_purge_reviews
-
   deploy_main_to_staging:
-    docker:
-      - image: cimg/ruby:3.2.2-node
+    executor:
+      name: tariff/ruby
     environment:
-      SENTRY_ENVIRONMENT: "staging"
```
willfish, line 490: Assuming this should remain staging?

amberstarlight: Yep, looks like I did some copypasta

.circleci/config.yml
```diff
@@ -2,98 +2,76 @@ version: 2.1
...
-            cf delete "$CF_APP-<< parameters.environment_key >>" -f
-
-            # Switch name of "dark" version to claim correct name
-            cf rename "$CF_APP-<< parameters.environment_key >>-dark" "$CF_APP-<< parameters.environment_key >>"
+            cf map-route \
+              "$CF_APP-<< parameters.environment_key >>-dark" \
+              london.cloudapps.digital \
+              -n "$CF_APP-<< parameters.environment_key >>-dark"
+
+            cf add-network-policy "$CF_APP-<< parameters.environment_key >>-dark" \
```
willfish: Any thoughts on splitting the dark app name onto a new line?

.circleci/config.yml
```diff
@@ -447,80 +423,74 @@ jobs:
...
-    environment:
-      SENTRY_ENVIRONMENT: "development"
-    steps:
-      - cf_purge_reviews
-
   deploy_main_to_staging:
-    docker:
-      - image: cimg/ruby:3.2.2-node
+    executor:
+      name: tariff/ruby
```
willfish, line 460: Should this be a ruby-node executor?

.circleci/config.yml
```diff
@@ -447,80 +423,74 @@ jobs:
...
+          environment_key: staging
+          app_domain_prefix: staging
+      - tariff/sentry-release:
+          environment: staging
 
   deploy_release_to_staging:
-    docker:
-      - image: cimg/ruby:3.2.2-node
+    executor:
+      name: tariff/ruby
```
willfish, line 478: Same again re ruby-node executor

.circleci/config.yml
amberstarlight: So this uses the orb `ruby` executor, which [defaults](https://github.com/trade-tariff/trade-tariff-ci-orb/blob/main/src/executors/ruby.yml) to `3.2.2-node`.

.circleci/config.yml
amberstarlight: As above, so below 🙏🏻

.circleci/config.yml
willfish: Oh. I guess the inverse of this is that when we only want a smaller ruby image we end up with node as well?

.circleci/config.yml
amberstarlight: A good idea! Done in https://github.com/trade-tariff/trade-tariff-frontend/pull/1515/commits/748a017e06ee67a24a7ea9968f1015a6b5beb093 :)

willfish: :tada:

willfish (approved):

PR #1532, opened 2023-07-11
HOTT-3476: Exchange rates periods UI

maverickvi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3476

### What?

I have added/removed/altered:

- [ ] Created exchange rates periods UI

### Why?

I am doing this because:

- We need to display this page on the TradeTariff front end

Screenshot 2023-07-11 at 16 54 53


app/models/exchange_rates/period_list.rb
```diff
@@ -0,0 +1,15 @@
...
+
+class ExchangeRates::PeriodList
+  include ApiEntity
+
+  attr_accessor :year
+
+  has_many :exchange_rate_years, class_name: 'ExchangeRates::Year'
+  has_many :exchange_rate_periods, class_name: 'ExchangeRates::Period'
+
+  def self.find(year, opts = {})
```
willfish, line 11: You could just use `.all(year: 2022)` and get rid of this find method which feels weird for a list anyway :)

maverickvi: I have discussed this with Will and we are going to stick with the .find override because the year params need to be optional on the frontend views and backend API.

app/helpers/exchange_rates_helper.rb
```diff
@@ -0,0 +1,5 @@
+module ExchangeRatesHelper
+  def filter_years(years, year_to_hide)
```
alexdesi, line 2: `filter_years` is a generic name,
what about
`visible_years`
`only_visible_years`
`exclude_hidden_years`
?

app/helpers/exchange_rates_helper.rb
```diff
@@ -0,0 +1,5 @@
...
+    years.reject { |year| year.year == year_to_hide || year.year == ExchangeRates::Year::HIDDEN_YEAR }
```
alexdesi, line 3: just to ask, is the year a year? 😄  or is it a date class?
`year.year` sounds a bit strange

spec/factories/exchange_rates/period_list_factory.rb
```diff
@@ -0,0 +1,11 @@
+FactoryBot.define do
+  factory :exchange_rate_period_list, class: 'ExchangeRates::PeriodList' do
+    transient do
+      year { 2020 }
```
alexdesi: is this transient used? 🤔

spec/models/exchange_rates/period_list_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+        },
+        {
+          id: '2022-exchange_rate_year',
+          type: 'exchange_rate_year',
+          attributes: {
+            year: 2022,
+          },
+        },
+      ],
+    }
```
alexdesi: this is quite a large data structure,
it make sense to me to put on a fixture `period_list_blablabla.json` and 
and load it using `file_fixture`:
```
def json_data(filename:)
  file_content = file_fixture("#{filename}.json").read
  JSON.parse(file_content, symbolize_names: true)
end
```

app/helpers/exchange_rates_helper.rb
maverickvi: Year class has year attribute ex 2021

app/helpers/exchange_rates_helper.rb
maverickvi: I think filter_years makes more sense because we provide a year to filter out and we also remove years we don't want

spec/factories/exchange_rates/period_list_factory.rb
maverickvi: I have removed this.

spec/models/exchange_rates/period_list_spec.rb
maverickvi: I have limited the examples to have just 1 relationship to make this smaller. I have left it in the file as it makes it easier to read the test

alexdesi (dismissed):
it looks good to me 🚀

willfish (approved):

PR #1533, opened 2023-07-13
BAU: Pivot the CDN app out of the orb

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [x] Added a route for the frontend CDN

### Why?

I am doing this because:

- We can consistently between all apps pivot the origin since these are
unique but the CDN route is specific to the frontend

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,8 +35,10 @@ commands:
 
     steps:
       - checkout
+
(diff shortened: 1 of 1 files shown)


amberstarlight (dismissed):

amberstarlight (approved):
💚 green pipe

PR #1534, opened 2023-07-14
Amend README.md

alexdesi:
### What?

Remove API_SERVICE_BACKEND_DEFAULT from the readme.md, since it is not used anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,22 +4,19 @@ __Now maintained at https://github.com/trade-tariff/trade-tariff-frontend__
 
 https://www.trade-tariff.service.gov.uk/trade-tariff/sections
 
-This is the front-end application for:
+This is the front-end application for [Trade Tariff Backend](https://github.com/trade-tariff/trade-tariff-backend), which provides the Backend APIs.
 
-* [Trade Tariff Backend](https://github.com/trade-tariff/trade-tariff-backend)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,18 +8,13 @@ This is the front-end application for:
 
 * [Trade Tariff Backend](https://github.com/trade-tariff/trade-tariff-backend)
 
-This application requires one or more Trade Tariff Backend APIs to be running and the following env variables set: `API_SERVICE_BACKEND_DEFAULT`, `API_SERVICE_BACKEND_URL_OPTIONS`.
+To run TTF (Trade Tariff Frontend) you need:
 
-For example:
+- to run Trade Tariff Backend APIs
+- to set `API_SERVICE_BACKEND_URL_OPTIONS`
```
willfish: These are configured by default so we treat this as convention over configuration and leave it undocumented to avoid confusion?

alexdesi: Good point, maybe we can just mention it as an important variable, but we does not need to be changed in order to run the FE (and it looks like now)

willfish: Yeah. This is actually misleading because the reality is by convention we run:

1. The duty calculator on port 3002
2. The frontend on 3001
3. The backend on 3000

So all of these links would be required to point at 3000 anyway.

willfish (approved):

PR #1548, opened 2023-07-19
Hott 3645 additional codes for draught relief

alexdesi:
### Jira link
Add new measure condition popups on additional codes that feature draught relief (DR).

### What?
Add new measure condition popups on additional codes that feature draught relief (DR).

### Why?
This is part of the many updates related to the VAT/Excise for Alcoholic drinks


willfish (dismissed):

willfish (approved):

willfish:
Oh terraform planning is broken. Hold off for a bit

PR #1561, opened 2023-07-31
HOTT-3477: Create view page for exchange rates in given year

shjohnson:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3477

### What?

image

I have added/removed/altered:

- [x] Created UI for viewing the exchange rates in particular month/year


app/models/exchange_rates/exchange_rate_file.rb
```diff
@@ -0,0 +1,12 @@
+require 'api_entity'
+
+class ExchangeRates::ExchangeRateFile
```
maverickvi: You can remove this and use app/models/exchange_rates/file.rb

app/views/exchange_rates/show.html.erb
```diff
@@ -0,0 +1,94 @@
...
+
+   <dt class="gem-c-metadata__term">From:</dt>
+   <dd class="gem-c-metadata__definition">
+     <strong><a href="https://www.gov.uk/government/organisations/hm-revenue-customs">HM Revenue &amp; Customs</a></strong>
+   </dd>
+   <dt class="gem-c-metadata__term">Published:</dt>
+   <dd class="gem-c-metadata__definition"><%= @published_date %></dd>
+   <dt class="gem-c-metadata__term">Download:</dt>
+   <dd class="gem-c-metadata__definition">
+     <% files.each do |file| %>
```
maverickvi: You can use the code in here app/views/exchange_rates/_document_detail.html.erb and put it in a view partial so it can be used in both places if possible

spec/factories/exchange_rates/exchange_rate_file_factory.rb
```diff
@@ -0,0 +1,10 @@
+FactoryBot.define do
+  factory :exchange_rate_file, class: 'ExchangeRates::ExchangeRateFile' do
```
maverickvi: You can remove this and use spec/factories/exchange_rates/file_factory.rb

app/views/exchange_rates/show.html.erb
shjohnson: I think this is different from document detail but i will put in a partial if can be used somewhere else.

maverickvi (approved):

PR #1568, opened 2023-08-03
HOTT-3717: Add file links, publication_date and file_size to period list frontend

maverickvi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3717

### What?

I have added/removed/altered:

- [ ] Added file links, publication_date and file_size to period list frontend

### Why?

I am doing this because:

- We need access to these attributes and links on the views

Screenshot 2023-08-03 at 11 06 14

### DIFF
--- a/app/models/exchange_rates/file.rb
+++ b/app/models/exchange_rates/file.rb
@@ -7,4 +7,8 @@ class ExchangeRates::File
                 :file_size,
                 :publication_date,
                 :format
+
+  def file_size_in_kb
+    sprintf('%.1f', file_size.to_f / 1024)
+  end
 end
--- a/app/models/exchange_rates/file.rb
+++ b/app/models/exchange_rates/file.rb
@@ -1,0 +1,10 @@
+require 'api_entity'
+
+class ExchangeRates::File
+  include ApiEntity
+
+  attr_accessor :file_path,
+                :file_size,
+                :publication_date,
+                :format
+end
--- a/app/models/exchange_rates/period.rb
+++ b/app/models/exchange_rates/period.rb
@@ -4,4 +4,6 @@ class ExchangeRates::Period
   include ApiEntity
 
   attr_accessor :year, :month
+
+  has_many :files, class_name: 'ExchangeRates::File'
 end
--- a/app/models/exchange_rates/period_list.rb
+++ b/app/models/exchange_rates/period_list.rb
@@ -16,6 +16,6 @@ class ExchangeRates::PeriodList
   def publication_date
     date_str = exchange_rate_periods.first.files.first&.publication_date
     date = Date.parse(date_str) if date_str
-    date&.strftime('%d %B %Y')
+    date&.to_formatted_s(:long)
   end
 end
--- a/app/models/exchange_rates/period_list.rb
+++ b/app/models/exchange_rates/period_list.rb
@@ -12,4 +12,10 @@ class ExchangeRates::PeriodList
     opts[:year] = year
     super(nil, opts)
   end
+
+  def publication_date
+    date_str = exchange_rate_periods.first.files.first&.publication_date
+    date = Date.parse(date_str) if date_str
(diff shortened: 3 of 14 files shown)


app/models/exchange_rates/file.rb
```diff
@@ -0,0 +1,14 @@
...
+
+class ExchangeRates::File
+  include ApiEntity
+
+  attr_accessor :file_path,
+                :file_size,
+                :publication_date,
+                :format
+
+  def file_size_in_kb
```
willfish, line 11: :raised_hands:

app/models/exchange_rates/period.rb
```diff
@@ -4,4 +4,6 @@ class ExchangeRates::Period
   include ApiEntity
 
   attr_accessor :year, :month
+
+  has_many :files, class_name: 'ExchangeRates::File'
```
willfish, line 8: Do you need the nesting here? I can't remember. I guess this just gets constantized?

app/models/exchange_rates/period_list.rb
```diff
@@ -12,4 +12,10 @@ def self.find(year, opts = {})
     opts[:year] = year
     super(nil, opts)
   end
+
+  def publication_date
+    date_str = exchange_rate_periods.first.files.first&.publication_date
+    date = Date.parse(date_str) if date_str
+    date&.strftime('%d %B %Y')
```
willfish: Is there a govuk locale which we could use here instead of strftime?

app/views/exchange_rates/_document_detail.html.erb
```diff
@@ -3,14 +3,18 @@
...
     <div class="attachment-details">
         <h3 class="govuk-heading-s govuk-!-font-size-27 govuk-!-font-weight-regular exchange-rates-period-heading">
           <%= Date::MONTHNAMES[period.month] %> <%= period.year %> monthly exchange rates
         </h3>
 
-        <p><strong><a class="govuk-link" href="/todo">View online</a></strong></p>
+        <p><strong><a class="govuk-link" href="/todo">View online</a></strong> | 
+        <% period.files.each_with_index do |file, index| %>
+          <strong><a href="<%= file.file_path %>"><%= file.format.upcase %></strong> <%= file.file_size_in_kb %> KB</a>
+          <% unless index == period.files.size - 1 %>|<% end %>
```
willfish: Wondering if we could put the pipe before the unless?

```suggestion
          <%= "|" unless index == period.files.size - 1 %>
```

lib/trade_tariff_frontend.rb
```diff
@@ -24,6 +24,7 @@ def accessible_api_endpoints
       geographical_areas
       chemical_substances
       simplified_procedural_code_measures
+      exchange_rates
```
willfish: Are we wanting both to expose the api _and_ to consume it in the frontend? If we expose it we'll want to add some documentation

app/models/exchange_rates/period.rb
maverickvi: yeah, I need it here

app/views/exchange_rates/_document_detail.html.erb
maverickvi: done

lib/trade_tariff_frontend.rb
maverickvi: I have removed this from accessible_api_endpoints and left it in public_api_endpoints because we need that to allow file downloading. There are upcoming tickets to expose the API and documentation will be added with those

app/models/exchange_rates/period_list.rb
maverickvi: I can't seem to find this so used .strftime('%d %B %Y')

maverickvi: found it

willfish (dismissed):

willfish (approved):

PR #1578, opened 2023-08-08
HOTT-3603: Fix supplementary unit when country is not selected.

willfish:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-3603

### What?

The supplementary unit should only be displayed if the user has selected a country with a supplementary unit (or one of the countries from the group) indicated.

### Why?
Currently, it shows a value that is related only to a country, which is wrong if no country is picked.


alexdesi:
Screenshot:
Screenshot 2023-08-08 at 17 29 04

willfish:
:sheep: :it:

willfish:
:sheep: :it:

alexdesi (approved):

PR #1581, opened 2023-08-09
HOTT-3778: Fix subheading in the A-Z index page.

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-3778

### What?
The product line suffix is something that we should not ever show on the UI of the OTT.

image

### DIFF
--- a/app/helpers/commodities_helper.rb
+++ b/app/helpers/commodities_helper.rb
@@ -216,17 +216,25 @@ module CommoditiesHelper
   end
 
   def abbreviate_code(code)
-    case code.gsub(/0*\z/, '').length
+    only_code = code_without_subheading(code)
+
+    case only_code.gsub(/0*\z/, '').length
+    when 9..10
+      only_code
     when 7..8
-      code.slice(0, 8)
+      only_code.slice(0, 8)
     when 5..6
-      code.slice(0, 6)
+      only_code.slice(0, 6)
     else
-      code
+      only_code
     end
   end
 
   def commodity_ancestor_id(index)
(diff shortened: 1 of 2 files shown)


app/helpers/commodities_helper.rb
```diff
@@ -216,17 +216,25 @@ def abbreviate_commodity_code(commodity)
   end
 
   def abbreviate_code(code)
-    case code.gsub(/0*\z/, '').length
+    only_code = code_without_subheading(code)
```
willfish, line 219: Out of interest, is this used elsewhere?

app/helpers/commodities_helper.rb
```diff
@@ -216,17 +216,25 @@ def abbreviate_commodity_code(commodity)
...
+
+    case only_code.gsub(/0*\z/, '').length
+    when 9..10
```
willfish, line 222: Do you need this extra condition now you're splitting out the producline suffix from the goods nomenclature item id?

willfish (approved):
Seems fine to me

app/helpers/commodities_helper.rb
alexdesi: if you mean `only_code`?  it is used on line#221

app/helpers/commodities_helper.rb
alexdesi: I guess you mean this could be capture in the `else`,
yes, that's true ... I think I wrote it just to be more explicit,
even if the result is the same as the `else` branch, the case is a different one.

PR #1583, opened 2023-08-10
BAU: Revamp footnote search UI

willfish:
### Jira link

BAU

image

### What?

I have added/removed/altered:

- [x] Migrate to using government supplied form builder
- [x] Added validations, models, etc, to simplify searching for footnotes

### Why?

I am doing this because:

- The backend api has gotten streamlined and no longer returns 32 MB of data


app/forms/footnote_search_form.rb
```diff
@@ -1,29 +1,41 @@
...
+  def type
+    code.to_s[0..1]
   end
 
-  delegate :present?, to: :instance_variables
-
   def to_params
     {
-      code:,
+      code: code.to_s[2..],
```
alexdesi: From the `type` on L#35, I deduce this is code without type, isn't it?
Maybe a small method well named to return `code.to_s[2..]` ?

app/services/footnote_search_service.rb
```diff
@@ -0,0 +1,13 @@
...
+  def initialize(query)
+    @query = query
+  end
+
+  def call
+    if @query.any?
+      Footnote.search(@query)
+    else
+      []
+    end
```
alexdesi: nitpick: maybe using clause guard?

```
def call
  return [] if @query.empty?

  Footnote.search(@query)
end
```

config/locales/en.yml
```diff
@@ -223,6 +223,14 @@ en:
       additional_code_description_label_html: "<strong>Additional code description</strong>"
       additional_code_description_hint_text: "Use this field if you know a key phrase included in the additional code description."
       submit: "Search"
+  footnote_search:
+    form:
+      legend_text: "Search"
+      footnote_label_html: "<strong>Footnote code</strong>"
```
alexdesi, line 229: I guess we should avoid storing html tags in the locales

willfish: I'm never sure about this. I think a single strong tag isn't painful and locales have a built in idiom for converting html locales to be safe with the `_html` key suffix.

alexdesi: Ok, I think we should not store them, ... but we are already doing it, so it's fine 😄

alexdesi (approved):
🚀

PR #1591, opened 2023-08-17
HOTT-3611: Support controlling showing proofs for specific area

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3611

### What?

I have added/removed/altered:

- [x] Added support for overriding when we should show proofs for specific areas
- [x] Used the enum idiom for simple business logic on the measure type
- [x] Extracted the rules of origin condition popup logic into a helper and not in the view
 
### Why?

I am doing this because:

- The override is necessary to force showing proofs for specific geographical area ids, only
- The enum idiom is just much cleaner and more consistent with the frontend application
- The logic for this view was misplaced, complex and hard to read in the middle of the view

### DIFF
--- a/spec/factories/rules_of_origin/scheme_factory.rb
+++ b/spec/factories/rules_of_origin/scheme_factory.rb
@@ -26,6 +26,7 @@ FactoryBot.define do
     links { attributes_for_list :rules_of_origin_link, link_count }
     articles { attributes_for_list :rules_of_origin_article, article_count }
     proofs { attributes_for_list :rules_of_origin_proof, proof_count }
+    show_proofs_for_geographical_areas { [] } # if emtpy it is ignored
     origin_reference_document { attributes_for :rules_of_origin_origin_reference_document }
 
     rule_sets do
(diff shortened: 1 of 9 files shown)


spec/factories/rules_of_origin/scheme_factory.rb
```diff
@@ -26,6 +26,7 @@
     links { attributes_for_list :rules_of_origin_link, link_count }
     articles { attributes_for_list :rules_of_origin_article, article_count }
     proofs { attributes_for_list :rules_of_origin_proof, proof_count }
+    show_proofs_for_geographical_areas { [] } # if emtpy it is ignored
```
willfish: :+1:

spec/models/rules_of_origin/scheme_spec.rb
```diff
@@ -221,6 +221,10 @@
       it { is_expected.to eql [] }
     end
 
+    context 'with show_proofs_for_geographical_areas' do
```
willfish: This all feels unnecessary. Can't we just extend the list of attributes on the response_data that already exists? There's nothing behavioural, here?

alexdesi: yes, we could expand the spec for the `applies_to_geographical_area?(area)`

PR #1603, opened 2023-08-25
HOTT-3779: Fix identifier class width

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3779

XI after

image

UK after

image

XI before

image

UK before

image

### What?

I have added/removed/altered:

- [x] Consolidates umbrella/regular identifier types
- [x] Removes duplicate scss definitions in print/tree scss files
- [x] Annotates umbrella/regular tree nodes with the service
- [x] Adds custom handling for the service to apply a specific width for the identifier
- [x] Removes inline width value that deviates from the variable pattern
- [x] Removes completely unused/complex ol li-selecting scss (we use ul for the tree items)

### Why?

I am doing this because:

- This whole thing was a mess and needed consolidating
- This fixes a specific issue with the display of the identifier column on the heading/subheading pages

### DIFF
--- a/app/helpers/commodities_helper.rb
+++ b/app/helpers/commodities_helper.rb
@@ -82,9 +82,32 @@ module CommoditiesHelper
         concat(content_tag(:div, vat_overview_measure_duty_amounts(commodity), class: 'vat', aria: { describedby: 'commodity-vat-title' }))
       end
 
-      concat(content_tag(:div, third_country_overview_measure_duty_amounts(commodity), class: 'duty', aria: { describedby: 'commodity-duty-title' }))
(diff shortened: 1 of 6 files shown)


app/helpers/commodities_helper.rb
```diff
@@ -82,9 +82,32 @@ def overview_measure_duty_amounts_for(commodity)
...
+          class: 'supplementary-units',
+          aria: { describedby: 'commodity-supplementary-title' },
+        ),
+      )
+
+      concat(
+        content_tag(
+          :div,
+          segmented_commodity_code(abbreviate_commodity_code(commodity), coloured: true),
+          class: "identifier service-#{TradeTariffFrontend::ServiceChooser.service_name}",
```
willfish, line 107: This is the only change in this helper (I'm annotating the service in the class of the identifier div)

maverickvi (approved):

willfish:
After on mobile:

image

Before on mobile:

image

PR #1624, opened 2023-09-04
HOTT-3634: Disable exchange rates on XI

maverickvi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3634

### What?

I have added/removed/altered:

- [x] Disable exchange rates on XI

### Why?

I am doing this because:

- We don't want users on XI to access exchange rates

Screenshot 2023-09-04 at 17 36 10


spec/controllers/exchange_rates_controller_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+RSpec.describe ExchangeRatesController, type: :controller do
+  before { TradeTariffFrontend::ServiceChooser.service_choice = nil }
+
+  context 'with xi as the service choice' do
+    before do
+      allow(TradeTariffFrontend::ServiceChooser).to receive(:xi?).and_return(true)
+      TradeTariffFrontend::ServiceChooser.service_choice = 'xi'
+    end
+
+    context 'when accessing the index action' do
```
alexdesi: I think we should use describe here,
something like:
`describe('GET #index')`

spec/controllers/exchange_rates_controller_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+      it 'raises an exception' do
+        expect { get :index }.to raise_exception(TradeTariffFrontend::FeatureUnavailable)
+      end
+    end
+
+    context 'when accessing the show action' do
```
alexdesi: and here:
`describe('GET #show')`

spec/controllers/exchange_rates_controller_spec.rb
maverickvi: using routes now so no need for specs

spec/controllers/exchange_rates_controller_spec.rb
maverickvi: using routes now so no need for specs

willfish (approved):

PR #1631, opened 2023-09-06
Add simple 404 page for when no data is present for a year requested

shjohnson:
### Jira link

HOTT-3910

### What?

I have added/removed/altered:

- [ ] Added 404 page for when the api returns no data for requested year

### DIFF
--- a/app/controllers/exchange_rates_controller.rb
+++ b/app/controllers/exchange_rates_controller.rb
@@ -9,6 +9,8 @@ class ExchangeRatesController < ApplicationController
     @monthly_exchange_rate = ExchangeRates::MonthlyExchangeRate.find(
       "#{year}-#{month}",
     )
+
+    render :show_404, status: :not_found unless @monthly_exchange_rate
   end
 
   private
--- a/app/views/exchange_rates/show_404.html.erb
+++ b/app/views/exchange_rates/show_404.html.erb
@@ -1,0 +1,14 @@
+<% content_for :top_breadcrumbs do %>
+   <%= generate_breadcrumbs "Exchange rates #{@monthly_exchange_rate.month_and_year_name}",
+                            [ [t('breadcrumb.home'), home_path],
+                            [t('breadcrumb.tools'), tools_path],
+                            [t('breadcrumb.exchange_rates'), exchange_rates_path], ] %>
+ <% end %>
+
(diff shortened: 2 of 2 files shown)


app/views/exchange_rates/show_404.html.erb
```diff
@@ -0,0 +1,14 @@
...
+   <%= generate_breadcrumbs "Exchange rates #{@monthly_exchange_rate.month_and_year_name}",
+                            [ [t('breadcrumb.home'), home_path],
+                            [t('breadcrumb.tools'), tools_path],
+                            [t('breadcrumb.exchange_rates'), exchange_rates_path], ] %>
+ <% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <p>
+      There are no exchnage rates present for the year requested please use the back navigation for years with data.
```
willfish, line 11:
```suggestion
      There are no exchange rates present for the year requested please use the back navigation for years with data.
```

shjohnson: Actually this has been pushed with old content, but yes i still make the mistake when typing `exchange` 😶‍🌫️

app/controllers/exchange_rates_controller.rb
```diff
@@ -9,6 +9,8 @@ def show
     @monthly_exchange_rate = ExchangeRates::MonthlyExchangeRate.find(
       "#{year}-#{month}",
     )
+
+    render :show_404, status: :not_found unless @monthly_exchange_rate
```
willfish, line 13: Does this work with your backend change?

So as I understand it, this will:

1. Fetch a response from the backend
2. Try and parse the json api structured json
3. Try and instantiate a monthly exchange rate `new parse_jsonapi(response)`
4. Fail with some unhandled exception
5. Propagate and not render this page

https://github.com/trade-tariff/trade-tariff-frontend/blob/main/lib/api_entity.rb#L234-L238

willfish: I wonder if we need to extend the ApiEntity implementation to handle 404 responses in the `find` method? I guess we could have it raise some generic `ApiEntity::NotFound` exception

shjohnson: sorry this is the wrong PR 1 sec

PR #1645, opened 2023-09-13
Add simple 404 page for when no data is present for a year requested

shjohnson:
### Jira link

HOTT-3910

### What?

I have added/removed/altered:

- [ ] Added 404 page for when the api returns no data for requested year

Example screenshot below
Screenshot 2023-09-13 at 17 39 45

### DIFF
--- a/app/models/exchange_rates/period_list.rb
+++ b/app/models/exchange_rates/period_list.rb
@@ -9,6 +9,8 @@ class ExchangeRates::PeriodList
   has_many :exchange_rate_periods, class_name: 'ExchangeRates::Period'
 
   def publication_date
+    raise Faraday::ResourceNotFound if exchange_rate_periods.empty?
+
     date_str = exchange_rate_periods.first.files.first&.publication_date
     date = Date.parse(date_str) if date_str
     date&.to_formatted_s(:long)
--- a/app/controllers/exchange_rates_controller.rb
+++ b/app/controllers/exchange_rates_controller.rb
@@ -1,6 +1,6 @@
 class ExchangeRatesController < ApplicationController
-  before_action :disable_search_form,
-                :disable_switch_service_banner
+  before_action :disable_search_form, :disable_switch_service_banner
+  rescue_from Faraday::ResourceNotFound, with: :render_404
 
   def index
     @period_list = ExchangeRates::PeriodList.find(params[:year])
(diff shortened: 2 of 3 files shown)


app/models/exchange_rates/period_list.rb
```diff
@@ -9,6 +9,8 @@ class ExchangeRates::PeriodList
   has_many :exchange_rate_periods, class_name: 'ExchangeRates::Period'
 
   def publication_date
+    raise Faraday::ResourceNotFound if exchange_rate_periods.empty?
```
willfish, line 12: Oh, now that you're using this exception I'm wondering if this is actually automatically propagated given what you've plumbed together in the backend?

shjohnson: I was just using similar to what we have elsewhere but i couldnt find anything else.

willfish: Yeah, I think that Jeremy added middleware propagation of 404s (amongst other status code errors) https://github.com/trade-tariff/trade-tariff-frontend/blob/main/app/services/client_builder.rb#L52

You actually ought to be able to just remove this line with your backend change

willfish: There's actually generic handling of Faraday::ResourceNotFound https://github.com/trade-tariff/trade-tariff-frontend/blob/main/config/initializers/exceptions.rb#L2

willfish: Not saying we should use the generic view, etc, since your 404 is more informative

shjohnson: Yeah i thought custom message for the message was the idea? Im happy either way although this looks a little more obvious what is going on and is general catch i guess? Like if we had loads of response errors we were raising it would be different. Can you also approve the backend change by the way :).

willfish: Yeah, custom message is great. Just a by-the-by thing

willfish: So basically, I think we can remove this propagation, here, and let the rescue_from/client do the work

shjohnson: I tested that but didnt work for me 😶‍🌫️

shjohnson: Even though no results are returned from the API with 404 i still need to check here to stop it blowing up

willfish: Oh, that's strange. I guess the middleware isn't propagating 404s :confused:

willfish (approved):

PR #1662, opened 2023-09-27
auto rubocop and manual changes

shjohnson:
### Jira link

HOTT-'Need to find this'

### What?

I have added/removed/altered:

- [ ] All the Rubocop changes and linting errors


app/controllers/healthcheck_controller.rb
```diff
@@ -1,6 +1,6 @@
 require 'api_entity'
 
-class HealthcheckController < ActionController::Base
+class HealthcheckController < ApplicationController
```
willfish: Actually, its worth double checking that this doesn't break anything after you've deployed this

https://dev.trade-tariff.service.gov.uk/healthcheck

shjohnson: Nice, yeah there is also some rebasing to do with this branch aswell

amberstarlight (approved):
Changes look sensible, Rubocop is happy, and all tests pass. LGTM.

PR #1663, opened 2023-09-29
HOTT-3976: Refactor exchange rate average page.

alexdesi:
### What?
This PR remove the page index_average_rates and reuse the template used for the spot/monthly pages.

### Jira link
https://transformuk.atlassian.net/browse/HOTT-3976

### Why?
To remove duplications and make the exchange rate pages consistent.


app/controllers/exchange_rates_controller.rb
```diff
@@ -3,19 +3,13 @@ class ExchangeRatesController < ApplicationController
 
   def index
     case type
-    when 'spot', 'scheduled'
+    when 'spot', 'scheduled', 'average'
```
alexdesi: I am adding a before filter instead of this case/when.

alexdesi: also those strings should be constants

willfish (dismissed):

willfish (approved):

PR #1666, opened 2023-10-04
HOTT-3809: UI tweaks for exchange rates

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-3809

### What?

I have added/removed/altered:

- [x] Tweaked the breadcrumbs for easier navigation of the difference rate types
- [x] Fixed the type label in the view online page

### Why?

I am doing this because:

- These were picked up in a demo to Chris

### DIFF
--- a/app/models/exchange_rate_collection.rb
+++ b/app/models/exchange_rate_collection.rb
@@ -5,7 +5,14 @@ class ExchangeRateCollection
 
   set_singular_path '/exchange_rates/:id'
 
-  attr_accessor :month, :year
+  attr_accessor :month, :year, :type
+
+  enum :type, {
+    monthly: %w[scheduled],
+    annual: %w[average spot],
+    average: %w[average],
+    spot: %w[spot],
+  }
 
   has_many :exchange_rate_files, class_name: 'ExchangeRates::File'
   has_many :exchange_rates, class_name: 'ExchangeRates::ExchangeRate'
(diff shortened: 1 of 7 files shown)


app/models/exchange_rate_collection.rb
```diff
@@ -5,7 +5,14 @@ class ExchangeRateCollection
 
   set_singular_path '/exchange_rates/:id'
 
-  attr_accessor :month, :year
+  attr_accessor :month, :year, :type
+
+  enum :type, {
+    monthly: %w[scheduled],
```
alexdesi, line 11: As you know this sounds a bit strange,
since scheduled also means Spot.
But that fix is probably for another occasion.

app/views/exchange_rates/index.html.erb
```diff
@@ -1,8 +1,10 @@
 <% content_for :top_breadcrumbs do %>
-   <%= generate_breadcrumbs "Exchange rates #{@period_list.year}",
-                            [ [t('breadcrumb.home'), home_path],
-                            [t('breadcrumb.tools'), tools_path],
-                            [t('breadcrumb.exchange_rates'), exchange_rates_path], ] %>
+  <%= generate_breadcrumbs t("breadcrumb.#{@period_list.type}_rates") ,
```
alexdesi, line 2: a space that precedes the comma,
also, is indentation right?

alexdesi (approved):
👍

PR #1691, opened 2023-10-25
Fix validation form on the /additional_code_search page.

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-4154

### What?
Fix wrong error messages on the /additional_code_search page.

### DIFF
--- a/app/forms/additional_code_search_form.rb
+++ b/app/forms/additional_code_search_form.rb
@@ -14,8 +14,10 @@ class AdditionalCodeSearchForm
   attr_writer :page
 
   def validate_code
+    sanitize_code
+
     if code.present?
-      errors.add(:code, :invalid) unless code =~ /\A([A-Z]|[0-9]){4}\z/
+      errors.add(:code, :length) unless code.length == 4
       errors.add(:code, :wrong_type) unless type.in?(self.class.possible_types)
     elsif description.blank?
       errors.add(:code, :blank)
@@ -53,5 +55,14 @@ class AdditionalCodeSearchForm
     def additional_code_type_ids
       @additional_code_type_ids ||= AdditionalCodeType.all.map(&:additional_code_type_id).sort
     end
+  end
+
+  private
+
+  def sanitize_code
+    return if code.blank?
+
+    code.strip!
+    code.upcase!
   end
 end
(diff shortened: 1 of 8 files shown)


app/forms/additional_code_search_form.rb
```diff
@@ -54,4 +56,13 @@ def additional_code_type_ids
       @additional_code_type_ids ||= AdditionalCodeType.all.map(&:additional_code_type_id).sort
     end
   end
+
+  private
+
+  def sanitize_code
```
willfish:
```suggestion
  def sanitize_code!
```

app/forms/additional_code_search_form.rb
```diff
@@ -14,8 +14,10 @@ class AdditionalCodeSearchForm
   attr_writer :page
 
   def validate_code
+    sanitize_code
+
     if code.present?
-      errors.add(:code, :invalid) unless code =~ /\A([A-Z]|[0-9]){4}\z/
```
willfish, line 18: This was checking length and specific characters before. I don't imagine its a huge deal no longer doing the two together but this is a different validation

spec/forms/additional_code_search_form_spec.rb
```diff
@@ -1,3 +1,5 @@
+require 'spec_helper'
```
willfish, line 1: You shouldn't need this because zeitwerk should be loading this lazily

willfish:
I'd add a small check of the sanitize code method - maybe one of the examples could include some white space. Also, if we're doing this here can we also do the other forms that take a composite code value (e.g. footnote/certificate search)

app/forms/additional_code_search_form.rb
alexdesi: It is just redundant.
Another validation checks if the first char is a valid type. And it works better. It checks if the type is a valid possible_types (and not just generic A-Z).

alexdesi:
Indeed, I've just checked and we got a similar error on the footnote_search page.
Since the code to fix is in different places, even if it is similar, I've create a new ticket: https://transformuk.atlassian.net/browse/HOTT-4256

willfish:
You'll want another for certificate search.

willfish (approved):

PR #1699, opened 2023-11-03
HOTT-4318 Dropped no longer needed webdrivers gem

jebw:
### Jira link

HOTT-4318

### What?

I have added/removed/altered:

- [x] Dropped webdrivers gem

### Why?

I am doing this because:

- it tries to automatically install chromedriver but that is no longer possible, now the developer should install a copy of chrome for testing, eg via homebrew (or install from their repos for linux)

- Low, only affects rspec


jebw:
Superceded by #1701

PR #1705, opened 2023-11-07
HOTT-4344: How to page error fix

maverickvi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-4344

### What?

I have added/removed/altered:

- [x] Fixed missing template error

### Why?

I am doing this because:

- we need to display 404 error when page is missing

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -49,7 +49,9 @@ class PagesController < ApplicationController
   ].freeze
 
   def howto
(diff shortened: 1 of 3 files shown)


jebw (dismissed):
This will error if the user requests a format of html?

How about checking if its blank or html?

willfish (dismissed):

willfish (dismissed):

willfish (approved):

PR #1709, opened 2023-11-13
HOTT-4326: Enable SES permissions

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-4326

### What?

I have added/removed/altered:

- [x] Added permissions to send emails via SES

### Why?

I am doing this because:

- This is required for the feedback emails


amberstarlight (dismissed):
LGTM

PR #1717, opened 2023-11-14
Fix and refactor additional_code_search_form.

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-4385

### What?
The form generated a 404 when the code field starts space.

While fixing the tests, I've made a small refactor: I separated the sanitize and validated concerns (now we do not need to sanitize the form before calling sanitise).

### Why?
We got Sentry errors: https://engine-le.sentry.io/issues/4632324212/?alert_rule_id=3445731&alert_type=issue&notification_uuid=e92937c0-d941-4b61-9dc6-9b05674c1172&project=5557002&referrer=slack

Screenshot of the error:
 
Screenshot from 2023-11-14 14-15-14

### DIFF
--- a/app/forms/additional_code_search_form.rb
+++ b/app/forms/additional_code_search_form.rb
@@ -13,10 +13,14 @@ class AdditionalCodeSearchForm
 
   attr_writer :page
 
+  def initialize(*args)
+    super(*args)
+
+    sanitize_code!
+  end
+
   def validate_code
     if code.present?
-      sanitize_code!
-
       errors.add(:code, :length) unless code.length == 4
       errors.add(:code, :invalid_characters) unless code =~ /\A([A-Z]|[0-9]){4}\z/
       errors.add(:code, :wrong_type) unless type.in?(self.class.possible_types)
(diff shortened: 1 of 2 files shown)


app/forms/additional_code_search_form.rb
```diff
@@ -13,10 +13,14 @@ class AdditionalCodeSearchForm
 
   attr_writer :page
 
+  def initialize(*args)
```
willfish, line 16: Do we need a constructor, here? We can just move this into the top of validate_code?

app/forms/additional_code_search_form.rb
```diff
@@ -13,10 +13,14 @@ class AdditionalCodeSearchForm
...
+    super(*args)
+
+    sanitize_code!
+  end
+
   def validate_code
     if code.present?
```
willfish, line 23: Or just do 
```suggestion
    if code.nil?
```

app/forms/additional_code_search_form.rb
alexdesi: I tried it,
but in that case `@form.to_params` would return unsanitised data.

We'd need to call `@form.valid?` to sanitise the data. 

In other words,
it is better to separate the concerns:
- we sanitise the form in  `initialization`
- we validate it  `validate_code`

willfish:
You need to rebase

willfish (approved):

PR #1720, opened 2023-11-15
Fix all warnings about date conversions.

alexdesi:
### What?

Fix all warnings about date conversions:
DEPRECATION WARNING: Using a :default format for Date#to_s is deprecated. Please use Date#to_fs instead.
### Why?

I am doing this because:
- It generates lots of noise when running spec tests
- This is deprecated and will be removed in the following Rails versions.

### DIFF
--- a/app/models/goods_nomenclature.rb
+++ b/app/models/goods_nomenclature.rb
@@ -17,7 +17,9 @@ class GoodsNomenclature
   has_many :ancestors, polymorphic: true
 
   def validity_start_date=(validity_start_date)
-    @attributes['validity_start_date'] = Date.parse(validity_start_date.to_s) if validity_start_date.present?
+    return unless validity_start_date.present?
+
+    @attributes['validity_start_date'] = validity_start_date.to_date
   end
 
   def validity_start_date
@@ -25,7 +27,9 @@ class GoodsNomenclature
   end
 
   def validity_end_date=(validity_end_date)
-    @attributes['validity_end_date'] = Date.parse(validity_end_date.to_s) if validity_end_date.present?
+    return unless validity_end_date.present?
+
+    @attributes['validity_end_date'] = validity_end_date.to_date
   end
 
   def validity_end_date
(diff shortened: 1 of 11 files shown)


app/models/goods_nomenclature.rb
```diff
@@ -17,15 +17,19 @@ class GoodsNomenclature
   has_many :ancestors, polymorphic: true
 
   def validity_start_date=(validity_start_date)
-    @attributes['validity_start_date'] = Date.parse(validity_start_date.to_s) if validity_start_date.present?
+    return if validity_start_date.blank?
+
+    @attributes['validity_start_date'] = validity_start_date.to_date
```
alexdesi, line 22: ### Why did I make this change?
- `to_s` is deprecated
-  `to_fs` does not work with String, and validity_start_date sometimes is a String (and sometimes a TimeZone)
- `date_to` works with string and other Date compatible classes.

app/models/goods_nomenclature.rb
```diff
@@ -17,15 +17,19 @@ class GoodsNomenclature
...
   end
 
   def validity_start_date
     @attributes['validity_start_date'].presence || NullObject.new
   end
 
   def validity_end_date=(validity_end_date)
-    @attributes['validity_end_date'] = Date.parse(validity_end_date.to_s) if validity_end_date.present?
+    return if validity_end_date.blank?
```
jebw, line 30: Much more readable

app/models/goods_nomenclature.rb
jebw: Just thinking this one through - it works great as is and replicates the existing behaviour _but_ it does highlight its not possible to change the value to nil

Our use case is read only so I don't see us having a model with a date assigned, then changing it back to nil so this isn't an issue here but for future reference if it was 

```ruby
@attributes['validity_start_date'] = validity_start_date.presence&.to_date
```

then it would handle blanks but also allow assigning to `nil`

jebw (approved):
LGTM :raised_hands:

PR #1735, opened 2023-11-28
Fix deprecation warnings

alexdesi:
### What?
- Replace implicit to_s to to_fs. (To fix warning deprecation)
- Remove gem responder.
- Fix WARNING: Using `expect { }.not_to raise_error(SpecificErrorClass)

### Why?
To keep code clean and updated.


jebw (approved):
LGTM - nice to drop the responders gem, its really not needed for our use case :+1:

PR #1736, opened 2023-11-28
BAU: Disable beta search in production

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [x] Disable beta search in production

### Why?

I am doing this because:

- This should be turned off when we go live with the migration since this functionality hasn't been tested

### DIFF
--- a/terraform/variables.tf
+++ b/terraform/variables.tf
@@ -45,3 +45,8 @@ variable "memory" {
   description = "Memory to allocate in MB. Powers of 2 only."
   type        = number
 }
+
+variable "beta_search_enabled" {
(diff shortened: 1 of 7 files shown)


terraform/variables.tf
```diff
@@ -45,3 +45,7 @@ variable "memory" {
   description = "Memory to allocate in MB. Powers of 2 only."
   type        = number
 }
+variable "beta_search_enabled" {
```
amberstarlight, line 49:
```suggestion

variable "beta_search_enabled" {
```

amberstarlight (approved):

PR #1737, opened 2023-11-28
BAU: Adds sentry configuration

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [x] Added sentry configuration

### Why?

I am doing this because:

- This is part of the standard development process


amberstarlight (approved):
Failing in prod as we haven't pushed secrets up that far, LGTM

PR #1738, opened 2023-11-28
Update to rails 7.1

alexdesi:
### Jira link
Hott 4457

### What?
Update rails to 7.1:
    - Run rails app:update to complete the update to Rails 7.1
    - Fix various compatibility issues with Rails 7.1
    - Fix issue with zeitwerk mode in lib/errors.rb (renamed to lib/unparseable_response_error.rb)
    - Fix some deprecation warnings.

### Why?
To get the latest features, security patches, and performance improvements.


Gemfile
```diff
@@ -1,13 +1,14 @@
 source 'https://rubygems.org'
 
-ruby File.read('.ruby-version')
+# ruby File.read('.ruby-version')
```
alexdesi: remove this comment

Gemfile
```diff
@@ -1,13 +1,13 @@
 source 'https://rubygems.org'
 
-ruby File.read('.ruby-version')
+ruby file: '.ruby-version'
```
willfish, line 3: TIL

bin/setup
```diff
@@ -1,25 +1,28 @@
 #!/usr/bin/env ruby
-require "fileutils"
+require 'fileutils'
```
willfish: Have you linted these munchkins? I'd probably preserve them 100% as they come + our config if so

config/application.rb
```diff
@@ -4,6 +4,8 @@
 # Pick the frameworks you want:
 require 'active_model/railtie'
 # require "active_job/railtie"
+
+# TODO: try to remove this:
 require 'active_record/attribute_assignment'
```
willfish, line 8: I don't think we can - I believe these handle the parsing of date parts into a single date attribute

config/application.rb
```diff
@@ -4,6 +4,8 @@
 # Pick the frameworks you want:
 require 'active_model/railtie'
 # require "active_job/railtie"
+
+# TODO: try to remove this:
```
willfish: Maybe just remove this comment?

willfish (dismissed):
So I can't see anything obvious - it would be good to properly validate this with a full regression suite in staging

Gemfile
```diff
@@ -16,7 +16,7 @@ gem 'faraday-net_http_persistent'
 gem 'faraday-retry'
 gem 'multi_json'
 gem 'net-http-persistent'
-gem 'routing-filter', github: 'svenfuchs/routing-filter'
+gem 'routing-filter', git: 'https://github.com/trade-tariff/routing-filter'
```
jebw: minor but this could stay as a github source?

bin/setup
```diff
@@ -1,25 +1,28 @@
...
-  system! "yarn install"
+  puts '== Installing node dependencies =='
+  system! 'yarn install'
 
   puts "\n== Removing old logs and tempfiles =="
-  system! "bin/rails log:clear tmp:clear webpacker:clobber"
+  system! 'bin/rails log:clear tmp:clear webpacker:clobber'
+
+  puts "\n== Restarting application server =="
+  system! 'bin/rails restart'
```
jebw: Can you drop this block please - my ideas around containerised development rely on the `bin/setup` in each repo script only doing 'setup'

And as @willfish said, if you don't lint this then it makes future diffs from `rails app:update` much easier

bin/setup
alexdesi: There were some small changes in the file, and Rubocop complains now, that's why I had to make those those changes.

config/environments/development.rb
```diff
@@ -48,14 +48,17 @@
...
   # Raises error for missing translations.
   # config.i18n.raise_on_missing_translations = true
 
   # Annotate rendered view with file names.
   config.action_view.annotate_rendered_view_with_filenames = true
 
-  # Uncomment if you wish to allow Action Cable access from any origin.
-  # config.action_cable.disable_request_forgery_protection = true
-
+  # Raise error when a before_action's only/except options reference missing actions
```
jebw: This comment should be the line before to the `raise_on_missing_callback_actions` I think?

config/environments/production.rb
```diff
@@ -31,10 +30,24 @@
   # config.action_dispatch.x_sendfile_header = "X-Sendfile" # for Apache
   # config.action_dispatch.x_sendfile_header = "X-Accel-Redirect" # for NGINX
 
+  # Assume all access to the app is happening through a SSL-terminating reverse proxy.
+  # Can be used together with config.force_ssl for Strict-Transport-Security and secure cookies.
+  # config.assume_ssl = true
+
+  # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
+  config.force_ssl = true
```
jebw: You'll want to check this one with @willfish and @amberstarlight  - they did some custom fix around cookies because they couldn't enable it previously I think. If it is enabled, I think the `.assume_ssl` option also needs enabling together with dropping the custom fix around cookie assignment

config/environments/production.rb
```diff
@@ -31,10 +30,24 @@
...
+
+  # Log to STDOUT by default
+  config.logger = ActiveSupport::Logger.new($stdout)
+    .tap  { |logger| logger.formatter = ::Logger::Formatter.new }
+    .then { |logger| ActiveSupport::TaggedLogging.new(logger) }
+
+  # Prepend all log lines with the following tags.
+  config.log_tags = [:request_id]
```
jebw: This can be commented out - we're defining a custom logger config lower down in the same file

jebw (dismissed):
Looks really good, some minor suggestions

Gemfile
alexdesi: yes, I'll use to the same format. :+1:

bin/setup
alexdesi: Removed!

about the linting, I'll double-check if Rubocop is happy with the double quote.

config/environments/development.rb
alexdesi: That's right!
Fixed, I've checked the `test.rb` too, which was correct.

config/environments/production.rb
alexdesi: good spot!  I missed that, thanks

config/environments/production.rb
```diff
@@ -35,7 +35,7 @@
   # config.assume_ssl = true
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  config.force_ssl = true
+  config.force_ssl = false
```
amberstarlight, line 38: Just as some background information for this;

- when we're doing container healthchecks, they're happening over [HTTP](https://github.com/trade-tariff/trade-tariff-platform-aws-terraform/blob/ab616fb3d45cbaa2dcdccb76c80b31da17cc93d9/environments/common/application-load-balancer/alb.tf#L34) currently.
- `force_ssl` responds to those healthcheck requests with 301s pointing to HTTPS.
- the healthcheck receives 301 instead of 200 and marks the app as unhealthy.
- the app never starts fully.

So we can change our load balancer healthchecks to HTTPS and then set `force_ssl = true` in all the Rails apps. I think that's a better way forward here. I'll prep a quick PR on the infrastructure repo to do this.

If we can do HTTPS healthchecks on all of our apps then this is a very easy fix.

willfish (dismissed):

willfish (approved):

PR #1743, opened 2023-11-29
HOTT-4269 SearchReferencePresenter updated for A-Z Index Page

HWallenberg:
as A-Z Index page should capitalise the first letter of the whole string (not every word) Otherwise, leave exactly as-is

### Jira link
https://transformuk.atlassian.net/browse/HOTT-4269

### What?

I have altered:

- [x] SearchReferencePresenter.rb to_s override to capitalize instead of titleize (and added a spec)

### Why?

I am doing this because:

- The A-Z Index Page was mutating titles of commodities (removing hyphens and making each word capitalised instead of just capitalising just the first word and leaving the rest as is.

### DIFF
--- a/spec/presenters/search_reference_presenter_spec.rb
+++ b/spec/presenters/search_reference_presenter_spec.rb
@@ -19,12 +19,9 @@ RSpec.describe SearchReferencePresenter do
   end
 
   describe '#to_s' do
-    it 'returns the correct string' do
-      expect(presented.to_s).to eq('Tomatoes')
-    end
+    let(:search_reference) { build :search_reference, title: 'semi-skimmed milk, not powdered' }
 
     it 'capitalizes the first letter of the string and leaves the rest as is' do
-      presented.title = 'semi-skimmed milk, not powdered'
       expect(presented.to_s).to eq('Semi-skimmed milk, not powdered')
     end
   end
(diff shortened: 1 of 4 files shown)


jebw (dismissed):
LGTM :raised_hands:

spec/presenters/search_reference_presenter_spec.rb
```diff
@@ -19,6 +19,13 @@
   end
 
   describe '#to_s' do
-    it { expect(presented.to_s).to eq('Tomatoes') }
+    it 'returns the correct string' do
+      expect(presented.to_s).to eq('Tomatoes')
+    end
+
+    it 'capitalizes the first letter of the string and leaves the rest as is' do
+      presented.title = 'semi-skimmed milk, not powdered'
```
alexdesi: just a little thing here,  
presented should be initialised somehow instead of direct assignment here.

alexdesi: Ops, one more thing,
the ticket also mentions something about the use of hyphens,
for example
"semi-skimmed milk"
instead of
"semi skimmed milk"

Should we fix those cases too?

alexdesi (approved):
nice one :rocket:

PR #1753, opened 2023-12-06
HOTT-4464 Fixing timing issues in JS specs

jebw:
### Jira link

HOTT-4464

### What?

I have added/removed/altered:

- [x] Increased the max time out for any one test in js specs from 2 to 5 seconds
- [x] Removed override code (`using_wait_until`) which dropped it from 2 to 1 seconds ??
- [x] Removed sleeps, these are replaced with calls to `has_css` - which, like `have_content`, will block until either the document element becomes available or max_wait_timeout is reached
- [x] Changed the cuprite config to wait longer for chrome to become available

### Why?

I am doing this because:

- The timing issues shouldn't require sleeps in specs to resolve - allowing Capybara to wait longer for a `have_css`/`have_content` condition to become true should resolve this whilst also allowing it to complete faster if it becomes true sooner
- The drops of wait timer were likely to increase failures rather than reduce them
- The cuprite config change is a bit speculative, its referenced in this github ticket against capybara - https://github.com/rubycdp/cuprite/issues/127. It shouldn't take longer as I understand it

- Low, should hopefully improve test suite reliability

### DIFF
--- a/spec/features/search_banner_spec.rb
+++ b/spec/features/search_banner_spec.rb
@@ -13,21 +13,16 @@ RSpec.describe 'Search banner', js: true do
         page.find('.tariff-search-banner .autocomplete__input').click
 
         page.find('.tariff-search-banner .autocomplete__input').set('gold')
-        sleep 1.5
 
+        expect(page).to have_css('.tariff-search-banner .autocomplete__option')
         expect(page.find('.tariff-search-banner .autocomplete__option:first-of-type').text).to eq('gold')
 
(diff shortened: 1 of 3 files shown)


alexdesi (approved):
It looks good to me, 
I didn't fully get how it works now without those "sleep" calls. Maybe the reason is those changes you made on the spec_helper.rb

jebw:
Updated the PR description to clarify how this works

PR #1754, opened 2023-12-06
HOTT-4233: Exchange rates accessibility changes

maverickvi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-4233

### What?

I have added/removed/altered:

- [ ] Made links more accessible

### Why?

I am doing this because:

- This will make the website more accessible for all types of users

Screenshot 2023-12-06 at 16 43 58

### DIFF
--- a/app/views/exchange_rates/_document_detail.html.erb
+++ b/app/views/exchange_rates/_document_detail.html.erb
@@ -12,21 +12,18 @@
         <%= Date::MONTHNAMES[period.month] %> <%= period.year %> <%= type %> exchange rates
       </h3>
 
-      <p>
-        <strong>
-          <% if period.has_exchange_rates %>
-            <%= link_to "View online", exchange_rate_collection_path("#{period.year}-#{period.month}", type: type), class: 'govuk-link' %>
-
-          <% end %>
-        </strong>
+      <p class="document-download">
+        <% if period.has_exchange_rates %>
+          <%= link_to "<strong>View online</strong>".html_safe, exchange_rate_collection_path("#{period.year}-#{period.month}", type: type), title: "View #{period.month} #{period.year} monthly exchange rates online", class: 'govuk-link' %>
+        <% end %>
 
         <% if period.files.any? %>
-          <%= " | " if period.has_exchange_rates %>
-          <% period.files.each_with_index do |file, index| %>
-            <strong><a href="<%= file.file_path %>"><%= file.format.upcase %></strong> <%= file.file_size_in_kb %> KB</a>
-              <%= ' | ' unless index == period.files.size - 1 %>
-            <% end %>
+          <% period.files.each do |file| %>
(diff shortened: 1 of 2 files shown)


amberstarlight (approved):
Looks much cleaner to my eyes.

app/views/exchange_rates/_document_detail.html.erb
```diff
@@ -12,21 +12,18 @@
...
+        <% end %>
 
         <% if period.files.any? %>
-          <%= " | " if period.has_exchange_rates %>
-          <% period.files.each_with_index do |file, index| %>
-            <strong><a href="<%= file.file_path %>"><%= file.format.upcase %></strong> <%= file.file_size_in_kb %> KB</a>
-              <%= ' | ' unless index == period.files.size - 1 %>
-            <% end %>
+          <% period.files.each do |file| %>
+            <a title="Download <%= period.month%> <%= period.year %> monthly exchange rates in <%= file.format.upcase %> format" href="<%= file.file_path %>">
```
jebw, line 22: Its probably worth using the `link_to` helper - it means when rails adds additional attributes to `<a>` tags for security purposes, then links such as this will be benefit - you can still use code blocks to separate the 'linked' content from the link_to helper, eg

```erb
<%= link_to file.file_path title: "Download <% ... %>" do %>
  <strong><%= file.format.upcase %></strong>
  <%= file.file_size_in_kb %>
<% end %>
```

app/views/exchange_rates/_document_detail.html.erb
```diff
@@ -12,21 +12,18 @@
...
-      <p>
-        <strong>
-          <% if period.has_exchange_rates %>
-            <%= link_to "View online", exchange_rate_collection_path("#{period.year}-#{period.month}", type: type), class: 'govuk-link' %>
-
-          <% end %>
-        </strong>
+      <p class="document-download">
+        <% if period.has_exchange_rates %>
+          <%= link_to "<strong>View online</strong>".html_safe, exchange_rate_collection_path("#{period.year}-#{period.month}", type: type), title: "View #{period.month} #{period.year} monthly exchange rates online", class: 'govuk-link' %>
```
jebw, line 17: Use of `.html_safe` is generally reserved for 'there's not another option' - it can cause problems if not used correctly and whilst this use is safe - usage of it in a code base tends to then lead to further usage.

Here it is possible to use regular rails helpers, this could be rewritten using either the `tag.strong('View online')` in place of the string or more neatly by using a code block onto the end of link_to

```erb
<% link_to exchange_rate_collection_path(), title: "View ..." do %>
  <strong>View online</strong>
<% end %>
```

The addition of `govuk-link` does no harm but isn't actually needed because we automatically add that to all links anyway (see https://github.com/trade-tariff/trade-tariff-frontend/blob/main/app/webpacker/packs/application.scss#L1)

jebw (commented):
Looks good but `.html_safe` is not really needed?

PR #1769, opened 2023-12-15
HOTT-4629: Improve exception handling

rasikasri:
### Jira link

[HOTT-4629](https://transformuk.atlassian.net/browse/HOTT-4629)

### What?

I have added/removed/altered:

- Added actions in the controller for the missing error-handling actions
- added generic view for errors: `error.html.erb` view

### Why?

I am doing this because:

- There are certain exceptions in which are mapped in the default Rails behaviour to actions we don’t currently support


app/controllers/errors_controller.rb
```diff
@@ -15,6 +15,17 @@ def not_found
     end
   end
 
+  # issue: this one is not handled by the controller
```
willfish: Remove me

app/models/exchange_rates/exchange_rate.rb
```diff
@@ -17,7 +17,7 @@ def formatted_validity_start_date
     validity_start_date&.to_date&.to_fs(:short)
   end
 
-  def formatted_validity_end_date
+  def formatted_validity_end_datee
```
willfish:
```suggestion
  def formatted_validity_end_date
```

config/environments/development.rb
```diff
@@ -11,8 +11,11 @@
   # Do not eager load code on boot.
   config.eager_load = false
 
+  # Raise exceptions instead of rendering exception templates.
+  config.action_dispatch.show_exceptions = true
+
   # Show full error reports.
-  config.consider_all_requests_local = true
+  config.consider_all_requests_local = false
```
alexdesi: The application will render the public error pages instead of detailed debug information.

config/environments/test.rb
```diff
@@ -22,14 +22,14 @@
     'Cache-Control' => "public, max-age=#{1.hour.to_i}",
   }
 
+  # Raise exceptions instead of rendering exception templates.
+  config.action_dispatch.show_exceptions = true
+
   # Show full error reports and disable caching.
-  config.consider_all_requests_local       = true
+  config.consider_all_requests_local       = false
```
alexdesi, line 29: The application will render the public error pages instead of detailed debug information.

config/environments/development.rb
alexdesi: It has pros and cons:
It makes it easier to manually test the behaviour as production.
On the other side, for certain debugging, we should set it to false.

spec/requests/application_controller_spec.rb
```diff
@@ -0,0 +1,51 @@
+RSpec.describe ApplicationController, type: :request do
```
alexdesi, line 1: I reconsidered the usefulness of these specs.
We can remove these. We are just (re)testing the mappings in the rails actionpack/lib/action_dispatch/middleware/exception_wrapper.rb are correct.

The useful tests are the `spec/requests/errors_controller_spec.rb`,
since the tests the routes and the rendered views.

config/environments/development.rb
```diff
@@ -11,6 +11,9 @@
   # Do not eager load code on boot.
   config.eager_load = false
 
+  # Raise exceptions instead of rendering exception templates.
+  config.action_dispatch.show_exceptions = true
```
willfish, line 15: I think we can remove this chipolata?

willfish: Nope, I'm wrong

willfish (approved):

PR #1772, opened 2023-12-19
HOTT-4725: Tactical UI changes to the legacy search

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-4725

image

### What?

I have added/removed/altered:

- [x] Added context to commodity descriptions that are simply 'other'
- [x] Added results from the commodities index to the commodities list
- [x] Added a bunch of coverage

### Why?

I am doing this because:

- This helps with some of the raised issues from the HMRC team and gives us some breathing room for strategic fixes

### DIFF
--- a/app/helpers/search_results_helper.rb
+++ b/app/helpers/search_results_helper.rb
@@ -1,4 +1,25 @@
 module SearchResultsHelper
+  def descriptions_with_other_handling(commodity)
+    ancestor_descriptions = commodity.ancestor_descriptions.reverse
+    return sanitize commodity.description.to_s unless commodity.description.to_s.match(/^other/i)
+
+    descriptions = []
+
+    ancestor_descriptions.each do |ancestor_description|
+      if ancestor_description.to_s.match(/^other/i)
+        descriptions.unshift(ancestor_description.to_s)
+      else
+        descriptions.unshift(ancestor_description.to_s)
+        break
+      end
+    end
+
+    descriptions[-1] = "<strong>#{descriptions.last}</strong>"
+    descriptions = descriptions.join(' > ')
+
+    sanitize descriptions
+  end
+
   def toggle_beta_search_inset_text
     if beta_search_enabled?
       switch_link = link_to('switch back to legacy search', toggle_beta_search_path)
--- a/app/helpers/search_results_helper.rb
+++ b/app/helpers/search_results_helper.rb
@@ -1,20 +1,18 @@
 module SearchResultsHelper
   def descriptions_with_other_handling(commodity)
-    ancestor_descriptions = commodity.ancestor_descriptions.reverse
     return sanitize commodity.description.to_s unless commodity.description.to_s.match(/^other/i)
+
+    ancestor_descriptions = commodity.ancestor_descriptions.reverse
 
     descriptions = []
 
     ancestor_descriptions.each do |ancestor_description|
-      if ancestor_description.to_s.match(/^other/i)
-        descriptions.unshift(ancestor_description.to_s)
-      else
-        descriptions.unshift(ancestor_description.to_s)
-        break
-      end
+      descriptions.unshift(ancestor_description.to_s)
+
+      break unless ancestor_description.to_s.match(/^other/i)
     end
 
-    descriptions[-1] = "<strong>#{descriptions.last}</strong>"
+    descriptions[-1] = tag.strong(descriptions.last)
     descriptions = descriptions.join(' > ')
 
     sanitize descriptions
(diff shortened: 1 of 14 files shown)


chrisbooth-tfm:
In the 'commodities' section of the results page this change includes non-declrable commodities/non-end line commodities, whereas the commodities section should only be full-on commodities, with no further digging down required.

e.g. 
https://dev.trade-tariff.service.gov.uk/search?q=knitted+t-shirt&resource_id=&input-autocomplete=knitted+t-shirt
there are some results there at subheading, where you still need to dig down. e.g. 61 09 90

willfish:
Talked through this in slack. So essentially:

1. We've included ancestor descriptions as searchable tokens
2. Higher up in the ancestor descriptions (e.g. at the heading level) is a positive match for "knitted"
3. This means that all commodities with a positive matching heading are returned
4. Because 'knitted' does not repeat at the commodity/subheading levels for the negated descriptions at those levels we score those documents lower
5. This ends up meaning the correct answer is at the top since its tokens repeat multiple times and opensearch boosts scores for repeating tokens

shjohnson:
Im not sure if anyone else wants to look through this but im happy with the changes and if chris is happy with the adjustments as mentioned above i think LGTM.

app/helpers/search_results_helper.rb
```diff
@@ -1,4 +1,25 @@
 module SearchResultsHelper
+  def descriptions_with_other_handling(commodity)
+    ancestor_descriptions = commodity.ancestor_descriptions.reverse
+    return sanitize commodity.description.to_s unless commodity.description.to_s.match(/^other/i)
+
```
alexdesi, line 4: this guard clause does not use ` ancestor_descriptions `, it could be move one line up:
```
    return sanitize commodity.description.to_s unless commodity.description.to_s.match(/^other/i)
    ancestor_descriptions = commodity.ancestor_descriptions.reverse
```

app/helpers/search_results_helper.rb
```diff
@@ -1,4 +1,25 @@
...
+    descriptions = []
+
+    ancestor_descriptions.each do |ancestor_description|
+      if ancestor_description.to_s.match(/^other/i)
+        descriptions.unshift(ancestor_description.to_s)
+      else
+        descriptions.unshift(ancestor_description.to_s)
+        break
+      end
+    end
```
alexdesi: both if/else branches exec 
` descriptions.unshift(ancestor_description.to_s)`
It can be simplified.

maybe something like:
```
 descriptions.unshift(ancestor_description.to_s)
 break unless ...
 ```

File: app/helpers/search_results_helper.rb [hunk 3]
```diff
@@ -1,4 +1,25 @@
...
+
+    descriptions[-1] = "<strong>#{descriptions.last}</strong>"
```
alexdesi: this is OK, I just notice we often use `tag` here in the Frondent to generate HTML tags

app/models/search/base_match.rb
```diff
@@ -18,7 +18,22 @@ def array_attr_reader(*args)
       def array_attr_writer(*names)
         names.each do |name|
           define_method("#{name}=") do |entry_data|
-            instance_variable_set("@#{name}", entry_data.map { |ed| name.to_s.singularize.capitalize.constantize.new(ed['_source'].key?('reference') ? ed['_source']['reference'] : ed['_source']) })
+            instance_variable_set(
+              "@#{name}",
+              entry_data.map do |ed|
+                klass = name.to_s.singularize.capitalize.constantize
```
alexdesi: I think this does not use `ed`. Should we put it outside the `map`?

app/models/search/goods_nomenclature_match.rb
```diff
@@ -12,7 +12,12 @@ class GoodsNomenclatureMatch < BaseMatch
     array_attr_writer :sections, :chapters, :headings, :commodities
 
     def commodities=(commodity_data)
-      @commodities = commodity_data.map { |cd| Commodity.new(cd['_source']) }
+      @commodities = commodity_data.map do |cd|
+        attributes = cd['_source']
+        attributes['score'] = cd['_score'] if cd['_score'].present?
+
+        Commodity.new(attributes)
+      end
```
alexdesi, line 20: I believe that:

`attributes['score'] = cd['_score'] if cd['_score'].present?`
is the same as
`attributes['score'] = cd['_score']`

alexdesi: if we want to remove the null cases, we could filter it out:
```
@commodities = commodity_data
                       .filter{|cd| cd['_score'].present?}
                       .map{...
```

willfish: That's an additional iteration that doesn't look better/add any benefits really. I'm happy just not assigning for now

alexdesi (approved):
:white_check_mark:

PR #1773, opened 2023-12-20
Add CSP policy and fix errors

shjohnson:
### Jira link

HOTT-4465

### What?

I have added/removed/altered:

- [ ] Added CSP policy
- [ ] Fixed inline content errors and css violations
- [ ] allowed *googletagmanager violations

### Notes
There is a JQMIGRATE info message that is annoying but as i understand is only an info message and think can only be resolved by migrating to a later version but not required for this PR.

### DIFF
--- a/config/initializers/content_security_policy.rb
+++ b/config/initializers/content_security_policy.rb
@@ -4,23 +4,23 @@
 # For further information see the following documentation
 # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy
 
-# Rails.application.configure do
-#   config.content_security_policy do |policy|
-#     policy.default_src :self, :https
-#     policy.font_src    :self, :https, :data
-#     policy.img_src     :self, :https, :data
-#     policy.object_src  :none
-#     policy.script_src  :self, :https
-#     policy.style_src   :self, :https
-#     # Specify URI for violation reports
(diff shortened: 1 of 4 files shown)


config/initializers/content_security_policy.rb
```diff
@@ -4,23 +4,23 @@
...
-#   # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy-Report-Only
-#   # config.content_security_policy_report_only = true
-# end
+Rails.application.configure do
+  config.content_security_policy do |policy|
+    policy.default_src :self
+    policy.font_src    :self, :data
+    policy.img_src     :self, :data
+    policy.object_src  :none
+    policy.script_src  :self, 'unsafe-eval', 'unsafe-inline', '*.googletagmanager.com'
```
jebw: Strongly feel we shouldn't do this - `unsafe-eval` and `unsafe-inline` pretty much defeat any benefit of CSP

jebw (changes requested):
some of those clauses are pretty self defeating

config/initializers/content_security_policy.rb
willfish: We actually don't need unsafe-line or unsafe-eval (none of our sources include these - I just grepped for these)

jebw (approved):
LGTM :raised_hands:

PR #1777, opened 2023-12-22
BAU: Removes unnecessary test context

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [x] Removed shared context and references to it

### Why?

I am doing this because:

- This was enabling configuration that was already in place from the test.rb file


alexdesi (approved):
:+1:

PR #1784, opened 2024-01-03
Fixes for HOTT-4678 Iceland-Norway ROO  Packaging defintions

HWallenberg:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-4678

### What?

I have 

- [x] Added accessor and placeholder for packaging retail

### Why?

I am doing this because:

-The ROO wizard wasn't splitting the different packaging on the components page with their specific articles and there is an extra .md file now


jebw (approved):
LGTM

PR #1788, opened 2024-01-05
HOTT-4810: Post migration cleanup

willfish:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-4809
https://transformuk.atlassian.net/browse/HOTT-4810

### What?

I have added/removed/altered:

- [x] Fixed broken redirecting to govuk site in the production environment
- [x] Removed references to the VCAP_APPLICATION env var
- [x] Removed tf variable that is no longer consumed
- [x] Removed dead root route which is overridden by another route

### Why?

I am doing this because:

- These are just some things that have been missed


terraform/main.tf
amberstarlight, line 1: Do we still need `GOVUK_APP_DOMAIN` in here? L65-66

willfish: Sadly we don't but a gem we use that uses another gem marks it as mandatory

terraform/variables.tf
```diff
@@ -45,8 +45,3 @@ variable "memory" {
   description = "Memory to allocate in MB. Powers of 2 only."
   type        = number
 }
-
-variable "beta_search_enabled" {
```
amberstarlight, line 49: We can get rid of the var in the corresponding `config_<environment>.tfvars` files now, too

amberstarlight (changes requested):

terraform/variables.tf
willfish: Done

amberstarlight (approved):

PR #1796, opened 2024-01-11
Fix Broken RoO Proofs ORD URL

HWallenberg:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-4516)

### What?

I have

- [x] Added method insert_ord_url in app/helpers/rules_of_origin_helper.rb to dynamically insert a URL where there is a bespoke placeholder (ie. {ord_url} in backend origin-declaration.mds under each regions RoO proofs section) 
- [x] Updated  app/views/rules_of_origin/_proofs.html.erb to use an ORD object to insert into the method.
- [x] Updated app/views/shared/_details.html.erb to consume the new method and render the correct URL.

### Why?

I am doing this because:

- The URL to an ORD in the RoO proofs section was never correct and has to be dynamic

### Have you? 

### Deployment risks 

- Possible backend files with updated placeholder may not deploy but this is very rare and won't cause an error.

Screenshot of origin reference document link at bottom of the screen
image

### DIFF
--- a/app/helpers/rules_of_origin_helper.rb
+++ b/app/helpers/rules_of_origin_helper.rb
@@ -58,7 +58,7 @@ module RulesOfOriginHelper
   end
 
   def replace_ord_url(content, ord)
-    content.gsub("{ord_url}", asset_path(ord.document_path))
+    content.gsub('{ord_url}', asset_path(ord.document_path))
   end
 
   def restrict_wrapping(content)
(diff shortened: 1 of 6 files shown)


jebw (dismissed):
LGTM

maverickvi (dismissed):

maverickvi (approved):

PR #1797, opened 2024-01-12
Update Sentry configuration to exclude specific exceptions.

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/HOTT-4857

### What?
Update Sentry configuration to exclude specific exceptions.

### Why?
During the last Rails update, these were erroneously deleted.


jebw (approved):
LGTM

PR #1799, opened 2024-01-12
HOTT-4244 Error message when the corresponding year exchange rates do not exist

HWallenberg:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-4244

### What?

I have altered:

- [x] Altered app/views/exchange_rates/show.html.erb with a conditional statement to check if there are any exchange rates which will display a message if not.
- [x] Update spec appropriately spec/views/exchange_rates/show.html.erb_spec.rb

### Why?

I am doing this because:

- User was able to enter an erroneous year into the address bar and a blank table was displayed instead of an appropriate message.

### Have you? (optional)

image


jebw (dismissed):

spec/views/exchange_rates/show.html.erb_spec.rb
```diff
@@ -7,23 +7,31 @@
...
+    it { is_expected.to have_css 'td', text: exchange_rate_collection.exchange_rates.first.currency_code.upcase }
+    it { is_expected.to have_css 'td', text: exchange_rate_collection.exchange_rates.first.rate }
+    it { is_expected.to have_css 'td', text: exchange_rate_collection.exchange_rates.first.formatted_validity_start_date }
+    it { is_expected.to have_css 'td', text: exchange_rate_collection.exchange_rates.first.formatted_validity_end_date }
+  end
+
+  context 'when there are no exchange rates' do
+    let(:exchange_rate_collection) { build(:exchange_rate_collection, exchange_rates: []) }
+
+    it { is_expected.to have_css 'p', text: "There are no #{exchange_rate_collection.type} exchange rates for the year specified." }
```
jebw, line 35: you could test that the table is not included?

```ruby
it { is_expected.not_to have_css 'table' }
```

jebw (dismissed):
LGTM - minor suggestion but perfectly happy without it as well

jebw (approved):
LGTM :raised_hands:

PR #1807, opened 2024-01-16
Changes the order of the Disallow directive.

alexdesi:
### What?
Disallow should come first, otherwise it might block the entire website from being indexed (and it looks like this is the case for OTT).

This PR allows: /exchange_rates, /tools, /news and /helps
And it blocks the indexing of all the other URLs.

### Jira link
https://transformuk.atlassian.net/browse/HOTT-4883

### Why?
To make some pages searchable (/exchange_rates, /tools, /news and /helps).

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,4 +1,7 @@
 # See http://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
 User-Agent: *
-Allow: /exchange_rates
 Disallow: /
+Allow: /exchange_rates
+Allow: /tools
+Allow: /news
+Allow: /help


public/robots.txt
```diff
@@ -1,4 +1,7 @@
 # See http://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
 User-Agent: *
-Allow: /exchange_rates
 Disallow: /
+Allow: /exchange_rates
+Allow: /tools
```
maverickvi, line 5: are these 3 links a requirement? can you link the ticket? I think the requirement is to only allow /exchange_rates?
Allow: /tools
Allow: /news
Allow: /help

maverickvi: I think it would be a good idea to get Google search console access or reregister the website on a new Google search console account before pushing to Robots.txt again

alexdesi: I asked Chris, and he'll confirm soon. 
~~No they are not, but I think we should include those pages:~~
~~my understanding is that we must not allow indexing of the goods, but it's OK  index the other pages like exchange_rates and the other pages that can be useful to the users.~~

~~For example, it makes sense that the user who searches for:~~
~~"Trade tariff news bulletin" to get as a result the page https://www.trade-tariff.service.gov.uk/news.~~

PR #1813, opened 2024-01-25
Help page for 999l changes only in development mode

shjohnson:
### Jira link

HOTT-4970

### What?

I have added/removed/altered:

- [ ] Added help page for 999l changes only for development as content needs review

Screenshot 2024-01-25 at 17 12 42

Screenshot 2024-01-25 at 10 17 12

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -1,6 +1,7 @@
 class PagesController < ApplicationController
   before_action :disable_search_form,
                 :disable_last_updated_footnote
+  before_action :disable_switch_service_banner, only: [:changes_999l]
 
   def glossary
     @glossary = Pages::Glossary.find(params[:id])
@@ -9,6 +10,8 @@ class PagesController < ApplicationController
   def cn2021_cn2022
     @chapters = Chapter.all
   end
+
+  def changes_999l; end
 
   def opensearch
     respond_to do |format|
--- a/app/views/pages/changes_999l.html.erb
+++ b/app/views/pages/changes_999l.html.erb
@@ -1,0 +1,470 @@
+<% content_for :top_breadcrumbs do %>
+<%= generate_breadcrumbs(
+  t('breadcrumb.measure_list_for_imports_post_999l'),
+  [
+    [t('breadcrumb.home'), home_path],
+    [t('breadcrumb.help'), help_path]
+  ]
+  )
+%>
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= page_header('Measure List for Imports Post 999L','Import Guidance') %>
+
+    <p class="govuk-body-m">
+      The document code 999L has been removed from all import measures from 1 February 2024 onwards. Replacement exemptions are appropriate for some measures and many of these are already available in CDS. We will update this document as changes are made around the removal of 999L.
+      This notice clarifies the options for traders across relevant measures.
+
+      The following measures have been updated in CDS already:
+      <div class="govuk-list">
+        <ul>
+          <li>350 Animal Health Certificate</li>
+          <li>351 Health and Safety Executive Import Licensing Firearms and Ammunition (firearms and nuclear licensing)</li>
+          <li>352 Attestation Document (Hops)</li>
+          <li>355 HMI Conformity Certificate</li>
+          <li>358 Home Office pre-cursor chemicals</li>
+          <li>359 Ammonium nitrates</li>
+          <li>360 Phytosanitary Certificates</li>
+          <li>362 Controlled drugs</li>
+          <li>363 Quarantine Release Certificate</li>
+          <li>705 Torture equipment</li>
+          <li>707 Ivory</li>
+          <li>710 CITES</li>
+          <li>712 IAS</li>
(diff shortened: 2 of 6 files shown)


app/controllers/pages_controller.rb
```diff
@@ -10,6 +10,10 @@ def cn2021_cn2022
     @chapters = Chapter.all
   end
 
+  def changes_999l
+    @disable_switch_service_banner = true
```
alexdesi: usually we put  disable switches on the before action:
` before_action :disable_switch_service_banner, only: [:changes_999l]`

app/views/pages/changes_999l.html.erb
```diff
@@ -0,0 +1,468 @@
...
..., 2933998015, 2933998016, 2933998018, 2933998019, 2933998020, 2933998021, 2933998022, 2933998023, 2933998024, 2933998027, 2933998030, 2933998032, 2933998033, 2933998034, 2933998036, 2933998037, 2933998038, 2933998040, 2933998041, 2933998042, 2933998045, 2933998046, 2933998047, 2933998048, 2933998053, 2933998054, 2933998055, 2933998056, 2933998058, 2933998059, 2933998061, 2933998063, 2933998067, 2933998068, 2933998071, 2933998072, 2933998073, 2933998075, 2933998078, 2933998081, 2933998082, 2934999010, 2934999012, 2934999015, 2934999016, 2934999018, 2934999020, 2934999021, 2934999022, 2934999023, 2934999024, 2934999025, 2934999026, 2934999027, 2934999028, 2934999030, 2934999035, 2934999037, 2934999039, 2934999041, 2934999042, 2934999043, 2934999044, 2934999049, 2934999056, 2934999059, 2934999060, 2934999061, 2934999062, 2934999063, 2934999064, 2934999066, 2934999067, 2934999068, 2934999071, 2934999074, 2934999075, 2934999077, 2934999078, 2934999079, 2934999080, 2934999083, 2934999084, 2934999085, 2934999086, 2934999087, 2934999088, 2937230000, 2939630000, 2939791000, 2939799010, 2939799020, 2939799030, 2939799040, 2939799050, 2939799060, 3004410000, 3004420000, 3004430000, 3502909000
```
alexdesi, line 200: these are data; should we hard code them in the view , or
maybe put them in YAML and read them via a controller? :thinking: 

not sure, it depends if  they never change

app/views/pages/help.html.erb
```diff
@@ -118,6 +118,13 @@
 
       <nav role="navigation" class="gem-c-related-navigation__nav-section" aria-labelledby="related-nav-related_items-40a88cdf" data-module="gem-toggle" data-gem-toggle-module-started="true">
         <ul class="gem-c-related-navigation__link-list" data-module="gem-track-click" data-gem-track-click-module-started="true">
+
+          <% unless Rails.env.production? %>
+            <li class="gem-c-related-navigation__link">
+              <%= link_to "Changes 999L", help_changes_999l_path %>
+            </li>
+          <% end %>
```
alexdesi: this kind of condition on the FE does not look nice, but I am not sure about an alternative.
Sure this should go only on prod?

alexdesi (commented):
It looks good, just minor comments.
Also, it'd be nice to have a test to check that the new page exists and maybe it's title.

app/views/pages/help.html.erb
shjohnson: Its an `unless` so it wont show on prod but actually this doesnt matter. I only had it to show on dev but actually we run rails envs in prod on dev so its pointless. it was literally a very small change that actually needs to be removed by end of day. it didnt work anyways 😂

app/views/pages/changes_999l.html.erb
shjohnson: Im not sure i think this is fine for now, its very satic content and will never know if they change or not as i guess policy. I think there is talk about presenting them in a different format so i think be wasted to mess around too much with them now. all these format changes were imagined by me and HMRC seem vaguely happy with it so will see what comes up.

app/controllers/pages_controller.rb
shjohnson: Yeah makes sense!

shjohnson:
Checking for content i think we were going against as such even if title but i can add routing spec

app/controllers/pages_controller.rb
```diff
@@ -10,6 +11,8 @@ def cn2021_cn2022
     @chapters = Chapter.all
   end
 
+  def changes_999l; end
```
alexdesi, line 14: :+1:

app/controllers/pages_controller.rb
```diff
@@ -1,6 +1,7 @@
 class PagesController < ApplicationController
   before_action :disable_search_form,
-                :disable_last_updated_footnote
+                :disable_last_updated_footnote,
+                :disable_switch_service_banner, only: [:changes_999l]
```
alexdesi: I think the behaviour we want is this:
```
 before_action :disable_search_form,

before_action :disable_switch_service_banner,  only: [:changes_999l]
```

alexdesi (dismissed):
:white_check_mark: :rocket:

alexdesi (approved):
:white_check_mark:

PR #1821, opened 2024-01-30
HOTT-4749: Bottom feedback banner with choice selection

maverickvi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-4749

### What?

I have added/removed/altered:

- [ ] Display feedback banner at bottom of all pages except on feedback page
- [ ] Remember users choice during feedback journey
- [ ] Remember users choice if any errors occur during form submission
- [ ] Email users choice
- [ ] Only allow valid choices

### Why?

I am doing this because:

- This improves feedback collection 

### Have you? (optional)

Screenshot 2024-01-30 at 16 19 58
Screenshot 2024-01-30 at 16 18 27
Screenshot 2024-01-30 at 16 19 03

### DIFF
--- a/app/webpacker/src/stylesheets/_feedback_useful_banner.scss
+++ b/app/webpacker/src/stylesheets/_feedback_useful_banner.scss
@@ -36,7 +36,7 @@
   }
 }
 
-@media (max-width: 640px) {
+@include govuk-media-query($until: tablet) {
   .feedback-useful-banner {
     p {
       padding-top: 9px;
--- a/app/webpacker/src/stylesheets/_feedback_useful_banner.scss
+++ b/app/webpacker/src/stylesheets/_feedback_useful_banner.scss
@@ -1,0 +1,45 @@
+.feedback-useful-banner {
+  border-top: 1px solid $govuk-border-colour;
+
+  .feedback-useful-container {
+    position: relative;
+    padding: 1em;
+    background-color: govuk-colour("light-grey");
+    height: 36px;
+    padding-top: 17px;
+  }
+
+  p {
+    font-weight: bold;
+    float: left;
+    padding-top: 7px;
+    padding-left: 11px;
+    padding-right: 15px;
+  }
+
+  .govuk-button {
+    border: 1px solid govuk-colour("black");
+    box-shadow: 0 2px 0 govuk-colour("black");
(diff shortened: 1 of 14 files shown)


amberstarlight (changes requested):
Couple of conflicts with main in this one to tidy up first but otherwise looks good to me.

maverickvi:
> Couple of conflicts with main in this one to tidy up first but otherwise looks good to me.

I just merged the other 2 ticket's like 20 mins ago so the conflicts are from those, I will sort these out, thanks

app/webpacker/src/stylesheets/_feedback_useful_banner.scss
```diff
@@ -0,0 +1,45 @@
...
+    box-shadow: 0 2px 0 govuk-colour("black");
+    min-width: 6em;
+  }
+}
+
+.govuk-footer {
+  border-top: 10px solid govuk-colour("blue");
+}
+
+@media (max-width: 900px) {
```
jebw, line 31: Is it possible to utilise the existing govuk break points rather than adding new ones? eg https://github.com/trade-tariff/trade-tariff-frontend/blob/main/app/webpacker/src/stylesheets/_tables.scss#L185

jebw (dismissed):
LGTM - one observation is adding yet more CSS break points

The GovUK Design System already defines these and we should utilise them rather than creating our own (same as we wouldn't define a heading to be a different size on just one page)

app/webpacker/src/stylesheets/_feedback_useful_banner.scss
maverickvi: I tried using Gov UK breakpoints but this is quite unique and the ones I tried didn't match my needs.

maverickvi:
@amberstarlight I have fixed the merge conflicts

app/webpacker/src/stylesheets/_feedback_useful_banner.scss
maverickvi: just updated 640 to use existing CSS breakpoint, we have to stick with 900px on line 31

amberstarlight (approved):
Looks good, agreed with J on reusing the `tablet` query.

PR #1842, opened 2024-02-28
HOTT-5171: Add GreenLanes api client models

rasikasri:
### Jira link

[HOTT-5171](https://transformuk.atlassian.net/browse/HOTT-5171)

### What?

I have added/removed/altered:

- [ ] Added a green lanes goods api models to the frontend

### Why?

I am doing this because:

- We need to fetch data from the green lanes goods nomenclature api and show it on an debug UI page

### Have you? (optional)

### DIFF
--- a/spec/models/green_lanes/category_assessment_spec.rb
+++ b/spec/models/green_lanes/category_assessment_spec.rb
@@ -1,0 +1,16 @@
+require 'spec_helper'
+
+RSpec.describe GreenLanes::CategoryAssessment do
+  subject(:category_assessment) { build(:category_assessment) }
+
+  describe '.relationships' do
+    let(:expected_relationships) do
+      %i[
+        geographical_area
+        excluded_geographical_areas
+      ]
+    end
+
+    it { expect(described_class.relationships).to match_array(expected_relationships) }
+  end
+end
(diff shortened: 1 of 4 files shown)


spec/models/green_lanes/category_assessment_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  describe '.relationships' do
+    let(:expected_relationships) do
+      %i[
+        geographical_area
+        excluded_geographical_areas
+      ]
+    end
+
+    it { expect(described_class.relationships).to match_array(expected_relationships) }
```
jebw: Rather than one 'test all the relationships' spec - this could be broken out into separate tests which would individually fail highlighting the origin of any future test failure, eg

```ruby
describe '.relationships' do
  subject { category_assessement.relationship }

  it { is_expected.to include :geographical_area }
  it { is_expected.to include :excluded_geographical_areas }
end
```

spec/models/green_lanes/goods_nomenclature_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+  subject(:goods_nomenclature) { build(:goods_nomenclature) }
+
+  describe '.relationships' do
+    let(:expected_relationships) do
+      %i[
+        applicable_category_assessments
+      ]
+    end
+
+    it { expect(described_class.relationships).to match_array(expected_relationships) }
```
jebw: The same approach could be used here

jebw (commented):
Looks good but could the specs be split to check the relationships separately

jebw (approved):
LGTM

PR #1844, opened 2024-03-04
HOTT-5148: Feedback form updates

maverickvi:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-5148

### What?

I have added/removed/altered:

- [ ] Added HMRC email
- [ ] Added red border when form has errors present

### Why?

I am doing this because:

-  Our feedback form needed to be updated

Screenshot 2024-03-04 at 15 51 35

### DIFF
--- a/app/webpacker/src/stylesheets/_feedback.scss
+++ b/app/webpacker/src/stylesheets/_feedback.scss
@@ -18,3 +18,7 @@
     color: red;
   }
 }
+
+#feedback-message-field-error { 
+  border-color: govuk-colour("red");
+}
--- a/app/views/feedback/new.html.erb
+++ b/app/views/feedback/new.html.erb
@@ -23,7 +23,7 @@
                             max_chars: 500 %>
 
       <p class="form-hint">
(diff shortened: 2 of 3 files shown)


app/webpacker/src/stylesheets/_feedback.scss
```diff
@@ -18,3 +18,7 @@
     color: red;
   }
 }
+
+#feedback-message-field-error {
```
jebw, line 22: Why do we need to manually style the error on the form, this isn't particularly scalable since we'd need to fix this for every form? The form builder be handling this already, if its not it probably means something else is up?

jebw: Looks like it _should_ just be working?

https://govuk-form-builder.netlify.app/introduction/error-handling/

jebw: Looks like an upstream bug - js is overwriting the class I think

jebw (approved):
LGTM

PR #1852, opened 2024-03-05
HOTT-5214: Add a green lanes GoodsNomenclature search and info pages

rasikasri:
### Jira link

[HOTT-5214](https://transformuk.atlassian.net/browse/HOTT-5214)

### What?

I have added/removed/altered:

- [ ] Add a page to allow looking up the greenlanes data for a goods nomenclature

### Why?

I am doing this because:

- Non technical users want to start querying the API

### Have you? (optional)


spec/requests/green_lanes/category_assessments_controller_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  let(:goods_nomenclature) { build :green_lanes_goods_nomenclature }
+
+  describe 'GET #create' do
+    before do
+      allow(GreenLanes::GoodsNomenclature).to receive(:find)
+                         .with(goods_nomenclature.goods_nomenclature_item_id)
+                         .and_return goods_nomenclature
+    end
+
+    let(:make_request) { post green_lanes_category_assessments_path, params: { green_lanes_category_assessment_search: { commodity_code: goods_nomenclature.goods_nomenclature_item_id } } }
```
jebw: Can you try to limit line length  - I tend to follow a rule of aiming for a soft limit of 80 chars, and a hard limit 100

This helps with readability of code

one approach would be to split the params into a separate `let` or another would be using a multiline `do`/`end` block

app/views/green_lanes/category_assessments/create.html.erb
```diff
@@ -0,0 +1,28 @@
+<%= page_header "Search results for &lsquo;#{h @category_assessments_search.commodity_code}&rsquo;".html_safe %>
```
jebw: `html_safe` is often best avoided - I can see the code sanitizes the comm code but in general its better to not add `html_safe` to codebases because it normalises usage and can lead to sanitization issues creeping in. 

One alternative is to use `sanitize`

```ruby
page_header sanitize "some code#{ @some.variable} sdfsldk"
```

Or you could just cut and past the the raw characters in (or use utf8 codes) and let ERB escape them

```ruby
page_header "Something ‘#{@some.variable}’"
```

Or do we really need the wrapping quote characters? I'd argue not

```ruby
page_header "Something #{@some.variable}"
```

app/views/green_lanes/category_assessments/_category.html.erb
```diff
@@ -0,0 +1,32 @@
+<% if @goods_nomenclature.applicable_category_assessments.any? %>
```
jebw, line 1: I think this needs an `else` - rather than showing a empty section, say that there are no applicable category assessements

app/controllers/green_lanes/category_assessments_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+
+    private
+
+    def ca_search_params
+      params.require(:green_lanes_category_assessment_search).permit(:commodity_code)
+    end
+
+    def check_allowed
+      raise TradeTariffFrontend::FeatureUnavailable unless TradeTariffFrontend.green_lane_allowed?
```
jebw, line 34: Good find :+1:

spec/requests/green_lanes/category_assessments_controller_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    it { is_expected.to have_http_status :ok }
+    it { is_expected.to have_attributes content_type: %r{text/html} }
+  end
+
+  describe 'GET #show' do
+    let(:make_request) { get green_lanes_category_assessments_path }
+
+    it { is_expected.to have_http_status :ok }
+    it { is_expected.to have_attributes content_type: %r{text/html} }
+  end
```
jebw, line 108: Can you and some a check that the exception is raised when the env var is negative - thanks

jebw (changes requested):
Looks really good and should be a great entry point for later work - couple of small changes requested, thanks

app/views/green_lanes/category_assessments/create.html.erb
rasikasri: Removed quote

app/views/green_lanes/category_assessments/_category.html.erb
rasikasri: Done

spec/requests/green_lanes/category_assessments_controller_spec.rb
rasikasri: Done

app/views/green_lanes/category_assessments/create.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= page_header "Search results for &lsquo;#{h @category_assessments_search.commodity_code}&rsquo;".html_safe %>
+<%= page_header "Search results for #{h @category_assessments_search.commodity_code}" %>
```
jebw, line 1: Not a problem but more FYI, the manual escaping of comm code (the `h`) can get dropped because since the string is no longer being marked `.html_safe` so the whole string will get escaped automatically

jebw (dismissed):

jebw (dismissed):

.env.development
```diff
@@ -14,7 +14,7 @@ DUTY_CALCULATOR=true
 DUTY_CALCULATOR_BASE_URL=http://localhost:3002/duty-calculator
 FRONTEND_HOST=http://localhost
 HTTPARTY_READ_TIMEOUT=661
-PERMUTATIONS=true
```
jebw: Why is this getting changed - this _should_ be unrelated to this work?

.env.development
```diff
@@ -14,7 +14,7 @@ DUTY_CALCULATOR=true
...
+PERMUTATIONS=false
 PORT=3001
 ROO_WIZARD=true
 SEARCH_BANNER=true
```
jebw: I believe we'd discussed disabling the search banner in local development which is maybe what you'd intended to change

jebw (changes requested):
Looks great - only needed change I think is the PERMUTATIONS env var change which makes local development behave differently to production

.env.development
rasikasri: I think this is a merging error

rasikasri: I fixed it now

jebw (approved):
LGTM

PR #1853, opened 2024-03-06
HOTT-5215: Add filtering the Goods Nomenclature page by geo area

rasikasri:
### Jira link

[HOTT-5215](https://transformuk.atlassian.net/browse/HOTT-5215)

### What?

I have added/removed/altered:

- [ ] Extend the green lanes goods nomenclature page to allow filtering by geographical area

### Why?

I am doing this because:

- User can choose country and feeding it into the api to filtering category assessment
-
-

### Have you? (optional)

Screenshot 2024-03-07 at 17 41 55

Screenshot 2024-03-07 at 17 42 34

### DIFF
--- a/app/views/green_lanes/category_assessments/show.html.erb
+++ b/app/views/green_lanes/category_assessments/show.html.erb
@@ -5,16 +5,17 @@
       <%= f.govuk_text_field :commodity_code, width: 'one-quarter',
                              label: { text: 'Commodity Code'} %>
 
-      <div class="country-picker-container">
-        <%= f.hidden_field :render_errors, value: false %>
-        <%= f.hidden_field :anchor, value: nil %>
-        <%= f.label :country, 'Select Country', class: 'govuk-label' %>
-        <%= f.select :country, GeographicalArea.country_options, { include_blank: 'All countries' }, class: 'custom-select', "aria-label": "Filter measures by country" %>
+      <%= f.govuk_collection_select :country,
+                                    TradingPartner.options,
+                                    :id,
+                                    :name,
+                                    options: { include_blank: 'All countries' },
+                                    label: { text: 'Select Origin' } %>
+
+      <div class="govuk-button-group">
+        <%= f.govuk_submit('Search') %>
       </div>
-      <div class="govuk-form-group govuk-!-margin-top-6"></div>
-      <div class="govuk-fieldset search-submit">
-        <%= f.govuk_submit("Search") %>
-      </div>
+
     </fieldset>
   </div>
 <% end %>
--- a/app/views/green_lanes/category_assessments/show.html.erb
+++ b/app/views/green_lanes/category_assessments/show.html.erb
@@ -5,7 +5,16 @@
       <%= f.govuk_text_field :commodity_code, width: 'one-quarter',
                              label: { text: 'Commodity Code'} %>
 
-      <%= f.govuk_submit("Search") %>
+      <div class="country-picker-container">
+        <%= f.hidden_field :render_errors, value: false %>
+        <%= f.hidden_field :anchor, value: nil %>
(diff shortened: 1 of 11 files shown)


app/views/green_lanes/category_assessments/show.html.erb
```diff
@@ -5,7 +5,16 @@
...
 
-      <%= f.govuk_submit("Search") %>
+      <div class="country-picker-container">
+        <%= f.hidden_field :render_errors, value: false %>
+        <%= f.hidden_field :anchor, value: nil %>
+        <%= f.label :country, 'Select Country', class: 'govuk-label' %>
+        <%= f.select :country, GeographicalArea.country_options, { include_blank: 'All countries' }, class: 'custom-select', "aria-label": "Filter measures by country" %>
+      </div>
+      <div class="govuk-form-group govuk-!-margin-top-6"></div>
+      <div class="govuk-fieldset search-submit">
```
jebw: Not sure what the search-submit is for _but_ if the formbuilder is used correctly we shouldn't need custom spacing

This might be because the earlier field isn't using the govuk formbuilders version of `select`

app/views/green_lanes/category_assessments/show.html.erb
```diff
@@ -5,7 +5,16 @@
       <%= f.govuk_text_field :commodity_code, width: 'one-quarter',
                              label: { text: 'Commodity Code'} %>
 
-      <%= f.govuk_submit("Search") %>
+      <div class="country-picker-container">
+        <%= f.hidden_field :render_errors, value: false %>
+        <%= f.hidden_field :anchor, value: nil %>
+        <%= f.label :country, 'Select Country', class: 'govuk-label' %>
+        <%= f.select :country, GeographicalArea.country_options, { include_blank: 'All countries' }, class: 'custom-select', "aria-label": "Filter measures by country" %>
```
jebw: Please use the govuk formbuilder helpers - means things like errors will show up correctly

https://govuk-form-builder.netlify.app/form-elements/select/

jebw: I'm not sure what `custom-select` does but it in general custom classes should be avoided wherever possible

The design system combined with the formbuilder in theory should provide all the classes we need - there's always exceptions but the ideal is to try and avoid adding custom css

app/views/green_lanes/category_assessments/show.html.erb
```diff
@@ -5,7 +5,16 @@
       <%= f.govuk_text_field :commodity_code, width: 'one-quarter',
                              label: { text: 'Commodity Code'} %>
 
-      <%= f.govuk_submit("Search") %>
+      <div class="country-picker-container">
+        <%= f.hidden_field :render_errors, value: false %>
+        <%= f.hidden_field :anchor, value: nil %>
+        <%= f.label :country, 'Select Country', class: 'govuk-label' %>
```
jebw: Might make more sense to call this origin, and have the attribute as 'Select Origin'

jebw (changes requested):
Looks great, can you add a screen grab to the PR and swap to using the formbuilder for the dropdown (it will generate the label as well), thanks :+1:

app/views/green_lanes/category_assessments/show.html.erb
rasikasri: Removed

app/views/green_lanes/category_assessments/show.html.erb
rasikasri: Using the form builder for select now

app/views/green_lanes/category_assessments/show.html.erb
rasikasri: Done

jebw (approved):
LGTM - much cleaner using the govuk form builder methods :raised_hands:

PR #1854, opened 2024-03-07
HOTT-5285 Added Subheading redirect to Subheadings controller to fix broken URLs

HWallenberg:
### Jira link

https://transformuk.atlassian.net/browse/HOTT-5285

### What?

I have added:

- [x] Added a redirect from 'Subheading' to 'Subheadings' Controller 

### Why?

I am doing this because:

- It appears there are many broken links in the RoO files where references to Subheadings use the Subheading URL for example ../subheading/2009610000-80 should be ../subheadings/2009610000-80.
- Rather than fix all the RoO PSR  .json files during their transition it would be better to redirect these URLs logically reducing the number of 404 (not found) errors instantly.


maverickvi:
I would suggest replacing the text in all RoO files so we can keep the routes file cleaner. I know your option is faster but the cleaner option is always better. Let's see what @jebw has to say about this

jebw (approved):
Approved but unless its urgent I'd much prefer the data was fixed in the backend

(Nicely implemented redirect though)

HWallenberg:
Closed PR as updating the RoO files is the cleaner option

PR #1863, opened 2024-03-20
GL-306: Iterations to debug UI

rasikasri:
### Jira link

[GL-306](https://transformuk.atlassian.net/browse/GL-306)

### What?

I have added/removed/altered:

- [x] Changed green lanes category assessments UI layout.

### Why?

I am doing this because:

- Added the category assessment search from into the result page as user can interact with it when viewing the results.
- Added measure type and regulation info as user can find out why this commodity is in this category
- Grouped category assessments into two groups and improvede presentation of them

### Have you? (optional)


app/models/green_lanes/category_assessment_search.rb
```diff
@@ -6,5 +6,5 @@ class GreenLanes::CategoryAssessmentSearch
   attr_accessor   :commodity_code,
                   :country
 
-  validates :commodity_code, presence: true, length: { minimum: 6, maximum: 10 }
+  validates :commodity_code, presence: true, length: { minimum: 10, maximum: 10, message: 'Please search declarable commodity.' }
```
jebw: That message is a bit confusing and GDS guidelines are to not say 'Please'  - how about 'Enter a declarable 10 digit commodity code' - then the user knows how many digits its expecting?

app/views/green_lanes/category_assessments/_categories.html.erb
```diff
@@ -0,0 +1,18 @@
+<% if @goods_nomenclature.filter_by_category('1').any?%>
+  <h2>Category 1 Assessments</h2>
+  <div class="govuk-accordion" data-module="govuk-accordion" id="accordion-default">
+    <% @goods_nomenclature.filter_by_category('1').each do |category| %>
+      <%= render 'green_lanes/category_assessments/category', category: category %>
```
jebw: Minor but for future reference theres a `collection:` option when rendering partials, so you can do

```ruby
render partial: 'green_lanes/category_assessments/category', 
       collection: @goods_nomenclature.filter_by_category('1')
```

config/routes.rb
```diff
@@ -107,7 +107,7 @@
 
   resolve('GreenLanes::CategoryAssessmentSearch') { [:category_assessments] }
 
-  scope constraints: ->(_req) { TradeTariffFrontend::ServiceChooser.xi? } do
+  scope constraints: ->(_req) { TradeTariffFrontend::ServiceChooser.uk? } do
```
jebw, line 110: :ok_hand:

jebw (changes requested):
LGTM - only query is around the error message which doesn't tell the user why their 6 or 8 digit code got rejected

app/views/green_lanes/category_assessments/_categories.html.erb
rasikasri: Done

app/models/green_lanes/category_assessment_search.rb
rasikasri: Done

app/models/green_lanes/goods_nomenclature.rb
```diff
@@ -21,6 +21,8 @@ def filter_by_category(category)
     if applicable_category_assessments.any?
       grouped = applicable_category_assessments.group_by(&:category)
       grouped[category]
+    else
+      []
```
jebw: a simpler approach to achieve the same thing would be as follows?

```ruby
grouped = applicable_category_assessments.group_by(&:category)
grouped[category] || []
```

app/models/trading_partner.rb
```diff
@@ -15,7 +15,7 @@ def validate_country
   end
 
   def self.options
-    GeographicalArea.all.map do |geographical_area|
+    GeographicalArea.all.sort_by(&:long_description).map do |geographical_area|
```
jebw: This affects other parts of the site? Has this been cleared with UCD/Chris to change the ordering across the whole site?

rasikasri: Moved this to CategoryAssessmentSearch

app/models/green_lanes/goods_nomenclature.rb
rasikasri: Done

app/models/green_lanes/category_assessment_search.rb
```diff
@@ -8,4 +8,13 @@ class GreenLanes::CategoryAssessmentSearch
...
                              length: { minimum: 10, maximum: 10, message: 'Enter a declarable 10 digit commodity code.' }
+
+  def self.country_options
+    GeographicalArea.all.sort_by(&:long_description).map do |geographical_area|
+      OpenStruct.new(
+        name: geographical_area.long_description,
+        id: geographical_area.id,
+      )
+    end
+  end
```
jebw: This can be simplified to

```ruby
def self.country_options
  GeographicalArea.all.sort_by(&:long_description)
end

File: app/views/green_lanes/category_assessments/_search_form.html.erb [hunk 7]
```diff
@@ -8,7 +8,7 @@
                              label: { text: 'Commodity Code'} %>
 
       <%= f.govuk_collection_select :country,
-                                    TradingPartner.options,
+                                    GreenLanes::CategoryAssessmentSearch.country_options,
                                     :id,
                                     :name,
```
jebw: this should change to `:long_description` which avoids then needing to map all the data into an openstruct

jebw (changes requested):
Hey, looks good but there's no need to map the country data into openstructs - the formbuilder will work with whichever attributes you tell it to work with - its only using `:name` and the moment because you're telling it to use `:name`

see https://govuk-form-builder.netlify.app/form-elements/select/#input-erb-select-field-with-a-label-and-hint

app/models/green_lanes/category_assessment_search.rb
rasikasri: Done

app/views/green_lanes/category_assessments/_search_form.html.erb
rasikasri: Done

jebw (approved):
LGTM :raised_hands:

PR #1879, opened 2024-04-18
OTT-29 Updated trading_partners select country page

HWallenberg:
### Jira link

https://transformuk.atlassian.net/browse/OTT-29

### What?

I have altered:

- [x] View for 'Select a country' page (AKA trading_partners view)

### Why?

I am doing this because:

- UX Change request that hint text is tariff site dependant and submit text should change

BEFORE 
image

AFTER (UK)
image

AFTER (XI)
image


spec/views/trading_partners/show.html.erb_spec.rb
```diff
@@ -8,4 +8,22 @@
   describe 'header' do
     it { is_expected.to have_css 'header span', text: /UK Integrated/ }
   end
+
+  describe 'Tariff specific change xi' do
```
willfish:
```suggestion
  context 'when on the xi service' do
```

spec/views/trading_partners/show.html.erb_spec.rb
```diff
@@ -8,4 +8,22 @@
...
+    before do
+      allow(TradeTariffFrontend::ServiceChooser).to \
+        receive(:service_choice).and_return 'xi'
+    end
+
+    it { is_expected.to have_css 'form', text: / View EU / }
+  end
+
+  describe 'Tariff specific change uk' do
```
willfish:
```suggestion
  context 'when on the uk service' do
```

spec/views/trading_partners/show.html.erb_spec.rb
```diff
@@ -8,4 +8,22 @@
...
+    before do
```
willfish: Actually you can kill these contexts and use the shared one:

```ruby
include _context 'with XI service' { it { is_expected.to have_css 'form', text: /View EU/ } }
include _context 'with UK service' { it { is_expected.to have_css 'form', text: /View UK / } }
```

willfish (approved):

PR #1887, opened 2024-04-23
GL-378 Added deserialization of new API format

jebw:
### Jira link

GL-378

### What?

I have added/removed/altered:

- [x] Added ApiEntity model for Theme
- [x] Added relationships for measure types and regulations
- [x] Show measure types and regulations on debug ui
- [x] Also replaced problematic usage of NotImplementedError because it doesn't inherit from StandardError so doesn't get rescued
- [x] Changed UI to show both Cat 1 and Cat 2 assessments at the same time

### Why?

I am doing this because:

- So we can show results from the latest API format

- Low, debug screen for internal usage for now

### DIFF
--- a/spec/models/green_lanes/goods_nomenclature_spec.rb
+++ b/spec/models/green_lanes/goods_nomenclature_spec.rb
@@ -1,7 +1,7 @@
 require 'spec_helper'
 
 RSpec.describe GreenLanes::GoodsNomenclature do
-  subject(:goods_nomenclature) { build(:goods_nomenclature) }
+  subject(:goods_nomenclature) { build(:green_lanes_goods_nomenclature) }
 
   describe '.relationships' do
     subject { described_class.relationships }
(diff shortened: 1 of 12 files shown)


rasikasri (dismissed):

spec/models/green_lanes/goods_nomenclature_spec.rb
```diff
@@ -14,4 +14,46 @@
...
+    end
+
+    context 'with only category 2' do
+      let(:assessments) { attributes_for_pair :category_assessment, category: 2 }
+
+      it { is_expected.to have_attributes length: 2 }
+      it { is_expected.to all have_attributes category: 2 }
+    end
+
+    context 'with mix of category 1 and category 2' do
```
alexdesi: context "with both categories: 1 and 2"

spec/models/green_lanes/goods_nomenclature_spec.rb
```diff
@@ -14,4 +14,46 @@
...
+          attributes_for(:category_assessment, category: 2),
+          attributes_for(:category_assessment, category: 1),
+        ]
+      end
+
+      it { is_expected.to have_attributes length: 1 }
+      it { is_expected.to all have_attributes category: 1 }
+    end
+
+    context 'with no assessments' do
```
alexdesi, line 52: should we call this categories as well? for consistencies with the other tests?
    
    `context 'with no assessments' do`
    
    or should we amend the other test descriptions?

alexdesi (dismissed):
It looks good to me :white_check_mark: 

spec/models/green_lanes/goods_nomenclature_spec.rb
jebw: Good call - amended the context labels for consistency

alexdesi (approved):
:white_check_mark:

PR #1894, opened 2024-05-01
OTT-14 Updated Javascript following changes to _country_picker.html.erb html GDS Hierarchy

HWallenberg:
### Jira link

https://transformuk.atlassian.net/browse/OTT-14

### What?

I have altered:

- [x] js-country-picker-select as it wasn't maintaining the current tab (ie. Origin) due to the change in HTML hierarchy in _country_picker.html.erb

### Why?

I am doing this because:

- HTML Hierarchy had to change to meet GDS guidelines thus causing the js to maintain the tab.
- Solution was to select the closest anchor field that matched outside of the govuk-fieldset container.


willfish (approved):
We should really add a test for this lot when we move over to stimulus.js

PR #1897, opened 2024-05-01
GL-385: Create the first two steps of the Check Moving Requirement user journey

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/GL-385
https://transformuk.atlassian.net/browse/GL-388

Create the first two steps of the Check Moving Requirement feature

- Step1 is the /check_moving_requirements/start page
- Step2 is the /check_moving_requirements/edit page

### What?
These pages are part of the user journey to check for the requirements for moving goods from GB to NI

### Why?
Part of the GL job

### Have you? (optional)

- [] Reviewed view changes with stakeholders
WIP

WIP

TODO: need to put this feature behind feature flag


alexdesi:
page /check_movement_requirements/start:
Screenshot from 2024-05-03 16-16-44

page /check_movement_requirements/edit:
Screenshot from 2024-05-03 16-17-17

page /check_movement_requirements/result
Screenshot from 2024-05-03 16-17-03

app/views/check_moving_requirements/edit.html.erb
```diff
@@ -0,0 +1,65 @@
...
+
+      <p>
+        Commodity codes are internationally recognised reference numbers that describe specific goods.
+        Search the Online Trade Tariff if you're unsure of your commodity code.
+      </p>
+
+      <%= f.govuk_text_field :commodity_code, width: 'one-quarter',
+                  label: { text: 'Commodity Code'} %>
+
+      <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
```
jebw: does this need to self close? not quite sure what does and does not anymore post html5

app/views/check_moving_requirements/edit.html.erb
```diff
@@ -0,0 +1,65 @@
...
+          method: :put,
+          url: update_check_moving_requirements_path do |f| %>
+
+      <p>
+        Commodity codes are internationally recognised reference numbers that describe specific goods.
+        Search the Online Trade Tariff if you're unsure of your commodity code.
+      </p>
+
+      <%= f.govuk_text_field :commodity_code, width: 'one-quarter',
+                  label: { text: 'Commodity Code'} %>
```
jebw: indentation looks a bit off?

app/views/check_moving_requirements/edit.html.erb
```diff
@@ -0,0 +1,65 @@
...
+                              :long_description,
+                              options: { include_blank: '' },
+                              label: { text: 'Select Origin' } %>
+
+      <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
+
+      <%= f.govuk_date_field :moving_date,
+        maxlength_enabled: true,
+        legend: { text: 'When is the movement of goods taking place?', size: 's', tag: 'h1' },
+        hint: { text: "Arrangements can change over time. If you are unsure of the date of movement and want to check arrangements currently in place, enter today's date." } %>
```
jebw: Indentation?

app/controllers/check_moving_requirements_controller.rb
```diff
@@ -0,0 +1,33 @@
+class CheckMovingRequirementsController < ApplicationController
```
jebw: Can this be namespaced under `GreenLanes`, thanks

.env.development
```diff
@@ -10,6 +10,7 @@ BASIC_USERNAME=test
 BETA_SEARCH_HEADING_STATISTICS_THRESHOLD=3
 BETA_SEARCH_FACET_FILTER_DISPLAY_PERCENTAGE_THRESHOLD=5
 BETA_SEARCH_SWITCHING_ENABLED=true
+CHECK_MOVING_REQUIREMENTS_ENABLED=false
```
jebw: GREEN_LANES_ENABLED

config/routes.rb
```diff
@@ -215,6 +215,16 @@
...
+  get '/check_moving_requirements/start', to: 'check_moving_requirements#start', as: :start_check_moving_requirements
+  get '/check_moving_requirements/edit', to: 'check_moving_requirements#edit', as: :edit_check_moving_requirements
+  put '/check_moving_requirements/update', to: 'check_moving_requirements#update', as: :update_check_moving_requirements
+  get '/check_moving_requirements/result', to: 'check_moving_requirements#result', as: :result_check_moving_requirements
+
+  # This solution does not work because edit action could have an empy commodity_code in the URL
+  # resources :check_moving_requirements, only: %i[show update edit] do
+  #   get 'start', on: :collection
+  # end
+
```
jebw: I know I'd originally suggested a singular resource but it feels like this set of screens maps better to a plural resource?

```ruby
  scope module: :green_lanes do
    resources :check_moving_goods_requirements, only: %i[index edit update destroy]
  end

**Review by jebw — CHANGES_REQUESTED** [review]
Hey there, looks good - there's some changes which I think would benefit it

1. Namespacing the code (models/conrtollers/views) - even if the routes need to be at the top level, we can still keep the code a bit more organised
2. I think its possible to use `resources ...` in the `config/routes.rb` to clean that up a little
3. It might make sense to have a generic green lanes feature flag I think since we'll want to use the same for the results screens _but_ if you're envisioning those just using this feature flag then that should be fine

File: app/controllers/check_moving_requirements_controller.rb (hunk 4 above)
Comment (alexdesi) [reply app/controllers/check_moving_requirements_controller.rb]: I agree, done!

File: .env.development (hunk 5 above)
Comment (alexdesi) [reply .env.development]: Done, also I reused/renamed the  ALLOW_GREEN_LANES (added by @rasikasri )

File: config/routes.rb (hunk 6 above)
Comment (alexdesi) [reply config/routes.rb]: I adopted a similar solution, under the existing
`    namespace :green_lanes do`

File: app/forms/check_moving_requirements_form.rb [hunk 7]
```diff
@@ -0,0 +1,13 @@
+class CheckMovingRequirementsForm
```
jebw: Namespacing

app/views/check_moving_requirements/edit.html.erb
alexdesi: yes, <hr> is the correct way in html5, XHTML is more restrict and it need to be close

jebw (dismissed):
LGTM

rasikasri (approved):

PR #1898, opened 2024-05-02
OTT-254-Removed govuk-width-container div from Country Picker

HWallenberg:
### Jira link

https://transformuk.atlassian.net/browse/OTT-254

### What?

I have removed

- [x] govuk-width-container div from Country Picker

### Why?

I am doing this because:

- It was causing a padding layout issue in mobile (small screen) mode.

### Have you? (optional)

BEFORE
image

AFTER
image


jebw (dismissed):
Looks good but indentations looks off? Looks like this is an original indentation problem thats just stepped in a level?

Good debugging on finding what was up

HWallenberg:
@jebw I've fixed indentation please re-review, thanks.

jebw (approved):
LGTM

PR #1911, opened 2024-05-16
Update chrome-version

alexdesi:
Update Chrome version to 115

Why?
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -190,7 +190,7 @@ jobs:
       - checkout
       - run: sudo apt-get update
       - browser-tools/install-chrome:
-          chrome-version: 114.0.5735.90 # TODO: remove -> https://github.com/CircleCI-Public/browser-tools-orb/issues/75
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -190,7 +190,7 @@ jobs:
       - checkout
       - run: sudo apt-get update
       - browser-tools/install-chrome:
```
amberstarlight, line 192:
```suggestion
      - browser-tools/install-chrome
```

.circleci/config.yml
```diff
@@ -190,7 +190,7 @@ jobs:
...
-          chrome-version: 114.0.5735.90 # TODO: remove -> https://github.com/CircleCI-Public/browser-tools-orb/issues/75
+          chrome-version: 124.0.6367.154 # TODO: remove -> https://github.com/CircleCI-Public/browser-tools-orb/issues/75
```
amberstarlight, line 193:
```suggestion
```

amberstarlight (commented):
We could stop specifying the version here - Google keeps purging older versions from their `dl` subdomain so this will just pop up again later I think

alexdesi:
This has been fixed.

PR #1915, opened 2024-05-21
GL-502: Add beta service

amberstarlight:
### Jira link

[GL-502](https://transformuk.atlassian.net/browse/GL-502)

### What?

I have added/removed/altered:

- Added beta frontend service.

### Why?

I am doing this because:

- The design team wants a stable beta service which can be used to test feature flags.


willfish (approved):
I think the devs can work out the feature flags as they go. :sheep: :it:

PR #1925, opened 2024-06-06
OTT-287 Fix Trading Partners to persist variety of commodity IDs over multiple tabs

HWallenberg:
### Jira link

https://transformuk.atlassian.net/browse/OTT-287

### What?

I have altered:

- [x] Trading Partners Controller and View to persist commodity ID on submit.
- [x] Trading Partners 'Reset all countries' link to use request.referer for commodity id
- [x] Fix syntax error in commodities.js

### Why?

I am doing this because:

- Request.Referer wasn't persisting with the current tab's commodity id on submit to the trading partners controller
- 'Reset all countries' was defaulting to the session commodity id and not the 'tabs' (ie the one from request.referer)
- Javascript kept on breaking in debug so syntax fix applied

Tested locally but QA Required to test over multiple tabs as per the Jira.

### DIFF
--- a/app/webpacker/src/javascripts/commodities.js
+++ b/app/webpacker/src/javascripts/commodities.js
@@ -287,7 +287,7 @@
         });
 
         $('form').on('click', 'button[type=submit]', function(e) {
-          this.closest('form').trigger('submit');
+          $(this).closest('form').trigger('submit');
         });
 
         this.responsivePlaceholder.initialize();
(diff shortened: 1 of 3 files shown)


app/webpacker/src/javascripts/commodities.js
```diff
@@ -287,7 +287,7 @@
         });
 
         $('form').on('click', 'button[type=submit]', function(e) {
-          this.closest('form').trigger('submit');
+          $(this).closest('form').trigger('submit');
```
willfish, line 290: jquery :scream_cat:

willfish (approved):

PR #1929, opened 2024-06-18
OTT-285: Enables stateless linkage back to goods nomenclature

willfish:
### Jira link

OTT-285

### What?

I have added/removed/altered:

- [x] Removes dead references to hard coded form submitted referer
- [x] Removes references to the goods nomenclature code in the session
- [x] Move to having goods_nomenclature_code everywhere we need it in the URL bar
- [x] Move to having goods_nomenclature_code everywhere we need it in form params

### Why?

We were half-in on two implementations for rendering back links and redirecting to a goods nomenclature across different pages.

- *One stateful* - it dependend on the session having the goods nomenclature code set and passed around via each request to the backend
- *One stateless* - passing goods nomenclature code values by submitting a hidden field in the trading partners form

This change consolidates these two approaches and removes any redundant references to the session variant.

The stateful variant meant that users could only open a single tab at a given time since the session is shared by all tabs.


alexdesi (approved):
It looks good :white_check_mark:

PR #1933, opened 2024-06-26
Gl596, GL-629 - Result pages Cat1 and Cat 2

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/GL-596
https://transformuk.atlassian.net/browse/GL-629

### What?
- I have added two new pages for the result when the result is only Cat1 and only Cat2
- Refactor the DetermineCategory class
 
### Why?
I am doing this because:
The user needs to know what are the categories of their goods before moving them to Northern Ireland

Very low risk.
Those pages are under Feature Flag, wich is disabled on prod.

### DIFF
--- a/app/controllers/green_lanes/moving_requirements_controller.rb
+++ b/app/controllers/green_lanes/moving_requirements_controller.rb
@@ -38,23 +38,24 @@ module GreenLanes
     end
 
     def result
-      @goods_nomenclature = GreenLanes::GoodsNomenclature.find(
+      goods_nomenclature = GreenLanes::GoodsNomenclature.find(
         moving_requirements_params[:commodity_code],
         { filter: { geographical_area_id: moving_requirements_params[:country_of_origin] } },
         { authorization: TradeTariffFrontend.green_lanes_api_token },
       )
 
-      @commodity_code = @goods_nomenclature.goods_nomenclature_item_id
-      @country_of_origin = moving_requirements_params[:country_of_origin]
+      @commodity_code = goods_nomenclature.goods_nomenclature_item_id
+      @country_of_origin = moving_requirements_params[:country_of_origin] || GeographicalArea::ERGA_OMNES
       @country_description = GeographicalArea.find(@country_of_origin).description
       @moving_date = moving_requirements_params[:moving_date]
+      @determine_categories = GreenLanes::DetermineCategory.new(goods_nomenclature)
 
-      categories = GreenLanes::DetermineCategory.new(@goods_nomenclature).call
+      @categories = @determine_categories.call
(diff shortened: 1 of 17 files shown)


app/services/green_lanes/determine_category.rb
```diff
@@ -25,6 +25,14 @@ def call
       end
     end
```
rasikasri, line 30: These methods are repeating same logic over and over and contradicting DRY prinsiple. Instead you can write two parameterised methods, something like
`def category_assessments(category)
    goods_nomenclature.applicable_category_assessments.select { |ca| ca.theme.category == category }
  end`

`def has_exemptions(category)
    category_assessments(category).select {|ca| ca.exemptions.any?}
  end`
  
  These two methods are enough to detemine the result in categories emthod

app/controllers/green_lanes/moving_requirements_controller.rb
```diff
@@ -38,23 +38,25 @@ def update
...
         { authorization: TradeTariffFrontend.green_lanes_api_token },
       )
 
-      @commodity_code = @goods_nomenclature.goods_nomenclature_item_id
-      @country_of_origin = moving_requirements_params[:country_of_origin]
+      @commodity_code = goods_nomenclature.goods_nomenclature_item_id
+      @country_of_origin = moving_requirements_params[:country_of_origin] || GeographicalArea::ERGA_OMNES
       @country_description = GeographicalArea.find(@country_of_origin).description
       @moving_date = moving_requirements_params[:moving_date]
+      @determine_categories = GreenLanes::DetermineCategory.new(goods_nomenclature)
```
rasikasri, line 51: Better to follow naming convention for the service class name, [DetermineCategory] is more of a method name, You can rename it to [CategoryDetermineService] or something similar

app/services/green_lanes/determine_category.rb
alexdesi: Good point, 
I removed the duplication without undermining readability.
It should be good now.

app/controllers/green_lanes/moving_requirements_controller.rb
alexdesi: No worries about this; it will be renamed on the next ticket because it will include the questions too.

rasikasri (approved):

app/controllers/green_lanes/moving_requirements_controller.rb
alexdesi: Anyway, this is not strictly a service. It has more than one method, and there is no method named `call`. I'll move it to a different place.

PR #1934, opened 2024-06-26
OTT-288  Updates to Country Selector on Commodity Code page

HWallenberg:
### Jira link

https://transformuk.atlassian.net/browse/OTT-285
https://transformuk.atlassian.net/browse/OTT-288

### What?

I have added/removed/altered:

- [x] Updated js-country-picker-select javascript function for described UX Changes
- [x] Updated GeographicalArea>CountryOptions to sort list of countries and relevant spec
- [x]  Updated the .autocomplete__arrow scss class

### Why?

I am doing this because:

- The UX Changes were to add 'All countries' as an available option and to sort the remaining options into alphabetical order
- The country selector was triggering an onConfirm event onBlur which was causing an error if the user clicked away from the select options without making a selection
- The pointer was missing the select element when hovering over the custom arrow

### Have you? (optional)

### DIFF
--- a/app/webpacker/src/javascripts/commodities.js
+++ b/app/webpacker/src/javascripts/commodities.js
@@ -538,21 +538,23 @@
 
         $('.js-country-picker-select').each(function() {
           (function(element) {
+            let previousValue = element.find('option:selected').text();
             accessibleAutocomplete.enhanceSelectElement({
               selectElement: element[0],
               minLength: 2,
-              autoselect: true,
+              autoselect: false,
               showAllValues: true,
-              confirmOnBlur: true,
+              confirmOnBlur: false,
               alwaysDisplayArrow: true,
               displayMenu: 'overlay',
+              placeholder: previousValue, // in case the user clicks away without selecting anything
               dropdownArrow: function() {
                 return '<span class=\'autocomplete__arrow\'></span>';
               },
               onConfirm: function(confirmed) {
                 const commodityCode = $('.commodity-header').data('comm-code');
                 // Handle the "All countries" case
-                if (confirmed === ' ' || confirmed === 'All countries') {
+                if (confirmed === 'All countries') {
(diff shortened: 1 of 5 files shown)


app/webpacker/src/javascripts/commodities.js
```diff
@@ -538,26 +538,36 @@
...
                 return '<span class=\'autocomplete__arrow\'></span>';
               },
+              defaultValue: 'All countries',
               onConfirm: function(confirmed) {
-                const code = /\((\w\w)\)/.test(confirmed) ? /\((\w\w)\)/.exec(confirmed)[1] : null;
-                element.val(code);
-                const anchorInput = element.closest('.govuk-fieldset').find('input[name$="[anchor]"]');
-                anchorInput.val(window.location.hash.substring(1)); // maintain the tab
-                element.parents('form:first').trigger('submit');
+                const commodityCode = $('.commodity-header').data('comm-code');
```
willfish, line 555: I know this feels frivolous but I'm wondering whether we can't extract all of this out into its own stimulus controller and write some unit tests?

I've tried it in the past and got waylaid but this file is just a bit horrible and is actually doing quite a lot of stuff

willfish (dismissed):

app/webpacker/src/javascripts/commodities.js
```diff
@@ -538,26 +538,36 @@
...
               showAllValues: true,
-              confirmOnBlur: true,
+              confirmOnBlur: false,
               alwaysDisplayArrow: true,
               displayMenu: 'overlay',
+              placeholder: previousValue, // in case the user clicks away without selecting anything
               dropdownArrow: function() {
                 return '<span class=\'autocomplete__arrow\'></span>';
               },
+              defaultValue: 'All countries',
```
willfish: I think we can remove this chipolata?

willfish (approved):

willfish (approved):

PR #1954, opened 2024-07-24
Hooking up the check your answers and the category result pages

shjohnson:
### Jira link

https://transformuk.atlassian.net/browse/GL-673
https://transformuk.atlassian.net/browse/GL-674
https://transformuk.atlassian.net/browse/GL-675
https://transformuk.atlassian.net/browse/GL-638

### What?

I have added/removed/altered:

- [ ] The three category results pages
- [ ] The Check Your Answers page


app/controllers/green_lanes/check_your_answers_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    def set_instance_variables
+      @commodity_code = results_params[:commodity_code]
+      @country_of_origin = results_params[:country_of_origin] || GeographicalArea::ERGA_OMNES
+      @country_description = GeographicalArea.find(@country_of_origin).description
+      @moving_date = results_params[:moving_date]
+    end
+
+    def determine_category
```
alexdesi: should we call this method
`determined_category` or even just `category`?

app/controllers/green_lanes/applicable_exemptions_controller.rb
```diff
@@ -135,6 +149,7 @@ def handle_next_page(next_page)
         commodity_code: params[:commodity_code],
         country_of_origin: params[:country_of_origin],
         moving_date: params[:moving_date],
+        category: params[:category],
```
willfish, line 152: Does this duplicate the category above?

app/controllers/green_lanes/results_controller.rb
```diff
@@ -2,66 +2,67 @@ module GreenLanes
...
-      case next_page
-      when :result_cat_1
-        render 'result_cat_1'
-      when :result_cat_2
-        render 'result_cat_2'
-      when :result_cat_3
-        render 'result_cat_3'
-      end
+      @category = results_params[:category]
+      @ans = JSON.parse(results_params[:ans])
```
willfish: I don't think you need `JSON.parse`, here?

willfish: It should just be url-encoded and parseable

app/controllers/green_lanes/results_controller.rb
```diff
@@ -2,66 +2,67 @@ module GreenLanes
...
+    private
+
+    def set_up
+      check_green_lanes_enabled
+      disable_switch_service_banner
+      disable_search_form
+      goods_nomenclature
+      set_instance_variables
+      determine_assessments
+    end
```
alexdesi: we are mixing many concern here, it may be difficult to maintain.

- We could keep:
      check_green_lanes_enabled
      disable_switch_service_banner
      disable_search_form
      
 in the before filter.
 
- we probably don't need to force the memorization on `goods_nomenclature`

- `determine_assessments` is only called in the initialize, we can just call the assignments in it

app/controllers/green_lanes/results_controller.rb
```diff
@@ -2,66 +2,67 @@ module GreenLanes
...
+        cat_2_exemptions_apply:,
       )
     end
 
     def cat_1_exemptions_apply
-      case results_params[:c1ex]
-      when 'true' then true
-      when 'false' then false
-      end
+      results_params[:c1ex] == 'true'
```
willfish: I'm wondering what happens, here, if there are no cat_1_exemptions questions and we pass the answer as false?

```ruby
{
  c1ex: nil,
  c2ex: nil,
}

cat_1_exemptions_apply # false
cat_2_exemptions_apply # false
```

app/controllers/green_lanes/check_your_answers_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+      @c2ex = params[:c2ex]
+    end
+
+    private
+
+    def goods_nomenclature
+      @goods_nomenclature ||= GreenLanes::FetchGoodsNomenclature.new(results_params).call
+    end
+
+    def set_instance_variables
```
alexdesi: It is probably clearer if we put all the assignments inside the `new` action.

app/controllers/green_lanes/results_controller.rb
willfish: So my understanding of the service we call to determine the category/result is that true, false and nil essentially have different meanings

- `true` - an affirmative answer was given to the question
- `false` - a negative answers was given
- `nil` - there was no question to ask

app/helpers/green_lanes_helper.rb
```diff
@@ -22,6 +22,26 @@ def all_exemption_questions_replied?(questions_count)
     questions_count == questions_replied_count
   end
 
+  def render_exemptions_or_no_card(category, exemptions_met, no_exemptions)
+    if no_exemptions
```
willfish: You could just condense this chipolata
```suggestion
    if no_exemptions || exemptions_met.empty?
```

alexdesi (commented):
`set_instance_variables`
or
`setup`
makes the code a bit harder to follow in my opinion.

... I'am writing some other comments

app/controllers/green_lanes/check_your_answers_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+      @country_of_origin = results_params[:country_of_origin] || GeographicalArea::ERGA_OMNES
+      @country_description = GeographicalArea.find(@country_of_origin).description
+      @moving_date = results_params[:moving_date]
+    end
+
+    def determine_category
+      @determine_category ||= DetermineCategory.new(@goods_nomenclature)
+    end
+
+    def results_params
```
alexdesi: I was trying to think of a better name,
it is called result_params,
but the params do not mention result at all :thinking:

app/controllers/green_lanes/results_controller.rb
```diff
@@ -2,66 +2,67 @@ module GreenLanes
...
+    def results_params
+      params.permit(:commodity_code, :country_of_origin, :moving_date, :c1ex, :c2ex, :category, :ans)
+    end
+
+    def goods_nomenclature
+      @goods_nomenclature ||= GreenLanes::FetchGoodsNomenclature.new(results_params).call
+    end
 
-      @commodity_code = goods_nomenclature.goods_nomenclature_item_id
+    def set_instance_variables
```
alexdesi: not much in favour of this pattern.
`set_instance_variables` is called in just one place.

app/services/green_lanes/determine_assessments.rb
```diff
@@ -0,0 +1,41 @@
+module GreenLanes
+  class DetermineAssessments
```
willfish: This doesn't have the api of a service object and it feels more like a presenter. Given its just a presenter, I'm wondering if we can avoid the side effects of calling determine_assessments in the constructor and just pull the answers to these specific questions on read?

app/controllers/green_lanes/results_controller.rb
willfish:
```ruby
class CategoryAssessmentsPresenter
  def initialize(goods_nomenclature, ans)
    @goods_nomenclature = goods_nomenclature
    @ans = ans
  end

  def category_1_assessments
    assessments.category_1_assessments
  end 

  def assessments
    @assessments ||= GreenLanes::DetermineAssessments.new(@goods_nomenclature, @ans)
  end
end
```

app/services/green_lanes/determine_assessments.rb
willfish:
```ruby
class CategoryAssessmentsPresenter
  def initialize(goods_nomenclature, ans)
    @goods_nomenclature = goods_nomenclature
    @ans = ans
  end

  def category_1_assessments
    assessments.category_1_assessments
  end 

  def assessments
    @assessments ||= GreenLanes::DetermineAssessments.new(@goods_nomenclature, @ans)
  end
end
```

willfish: Or better yet, just present the result of the assessments

```ruby
class CategoryAssessmentsPresenter
  def initialize(assessments)
    @assessments = assessments
  end

  def category_1_assessments
    assessments.category_1_assessments
  end

  private

  attr_reader :assessments
end
```

config/routes.rb
```diff
@@ -112,9 +112,9 @@
       resource :category_assessments, only: %i[create show]
 
       resource :start, only: %i[new]
-      resource :moving_requirements, only: %i[new create]
+      resource :moving_requirements, only: %i[new create update]
```
willfish:
```suggestion
      resource :moving_requirements, only: %i[new create]
```

willfish: So my understanding here is we have a rendered form (new) and a submitted (post) to a form create.

app/controllers/green_lanes/results_controller.rb
shjohnson: Yeah i did see this but then its caught further down the line as they wont have the cat 2 answers at all in the params?

the {1:{...},2:{...}}

app/controllers/green_lanes/results_controller.rb
shjohnson: This was a quick monkey patch as was getting annoyed with the params, i think we can fix this going forward and try just focus on getting the logic correct

app/controllers/green_lanes/applicable_exemptions_controller.rb
shjohnson: yes maybe this might have been a bad rebase 👍

app/controllers/green_lanes/results_controller.rb
```diff
@@ -2,66 +2,67 @@ module GreenLanes
   class ResultsController < ApplicationController
     include GreenLanesHelper
 
-    before_action :check_green_lanes_enabled,
-                  :disable_switch_service_banner,
-                  :disable_search_form
+    before_action :set_up, only: :show
```
willfish:
```suggestion
  before_action :check_green_lanes_enabled,
    :disable_switch_service_banner,
    :disable_search_form,
    :goods_nomenclature,
    :set_instance_variables,
    :determine_assessments,
    only: :show
```

willfish: I'm not sure the set_up method adds anything for readability, here?

app/controllers/green_lanes/check_your_answers_controller.rb
shjohnson: yeah thats fair

app/controllers/green_lanes/check_your_answers_controller.rb
shjohnson: there was quite a lot of refactoring i was doing and yeah can change back it just started to look a bit messy in my eyes

app/controllers/green_lanes/check_your_answers_controller.rb
shjohnson: erm i think this is fine as they are going to the results controller? but it has the params to build the result? Answers?

app/controllers/green_lanes/results_controller.rb
```diff
@@ -2,66 +2,67 @@ module GreenLanes
...
-      when :result_cat_3
-        render 'result_cat_3'
-      end
+      @category = results_params[:category]
+      @ans = JSON.parse(results_params[:ans])
     end
 
-    private
+    def determine_assessments
+      assessments = GreenLanes::DetermineAssessments.new(@goods_nomenclature, @ans)
```
willfish: We've got a lot of dependency coupling to the presenter, here, in the controller. I'm wondering whether its cleaner to just memoize the presenter and have the view ask questions of it?

app/controllers/green_lanes/results_controller.rb
shjohnson: So this again was part of a slight refactor happy to discuss. actually it would be better to move some of these into the green lane helper?

app/controllers/green_lanes/results_controller.rb
shjohnson: happy to discuss

app/views/green_lanes/check_your_answers/new.html.erb
```diff
@@ -0,0 +1,25 @@
...
+
+    <%= render 'green_lanes/shared/summary_card', title: 'About your goods', edit_path: 'EDIT GOOD DETAILS PAGE' %>
+      
+    <h2 class="govuk-heading-m">Category exemptions</h2>
+    
+    <%= render 'category_exemptions', title: 'Category 1 exemptions', category: '1', assessments: @category_one_assessments, edit_path: 'EDIT PATH HERE CAT 1 questions' %>
+    <%= render 'category_exemptions', title: 'Category 2 exemptions', category: '2', assessments: @category_two_assessments, edit_path: 'EDIT PATH HERE CAT 2 questions' %>
+
+    <div class="govuk-button-group">
+      <%= form_with url: green_lanes_result_path(params[:category]), method: :get do %>
```
willfish: Ah, I see. You're submitting the answers to a show and having a json body in the request object here that are basically url encoded form params.

So if we just make the method and verb a create/POST then I think we're not going to have to do JSON.parse in the result step

config/routes.rb
```diff
@@ -112,9 +112,9 @@
...
       resource :applicable_exemptions, only: %i[new create]
-      # resource :check_your_answers, only: %i[new create]
+      resource :check_your_answers, only: %i[new create]
       resources :results, param: :category, only: %i[show]
```
willfish:
```suggestion
      resources :results, param: :category, only: %i[create]
```

willfish: I guess we can just make this a post and simplify the parsing of the hidden form fields and such

app/controllers/green_lanes/results_controller.rb
willfish: Yeah, I think it just gets sorted with a migration to a POST create

app/controllers/green_lanes/results_controller.rb
willfish: Maybe we can just pass the raw value to the service, then

app/controllers/green_lanes/check_your_answers_controller.rb
willfish: I guess he's thinking this because the idiom in controllers is that receiving params methods are typically called `<model_name>_params` and it makes you think you're in the results controller. I don't mind too much personally.

app/controllers/green_lanes/results_controller.rb
willfish: Resolving as this was me getting confused by what was happening

app/helpers/green_lanes_helper.rb
```diff
@@ -22,6 +22,33 @@ def all_exemption_questions_replied?(questions_count)
...
+      render('category_assessments_card', category:)
+    elsif !all_exemptions_met
+      render('category_assessments_card', category:)
+    else
+      render 'exemptions_card', category_number: category
+    end
+  end
+
+  def exemption_status(ans, category, category_assessment_id, exemption_code)
```
alexdesi: Can we call this `exemption_met`?

app/views/green_lanes/check_your_answers/_category_exemptions.html.erb
```diff
@@ -0,0 +1,29 @@
...
+    </div>
+    <div class="govuk-summary-card__content">
+      <dl class="govuk-summary-list">
+        <% assessments.each do |category_assessment| %>
+          <% category_assessment.exemptions.each do |exemption| %>
+            <div class="govuk-summary-list__row">
+              <dt class="govuk-summary-list__key"><%= exemption.code %></dt>
+              <dd class="govuk-summary-list__value"><%= exemption.formatted_description %></dd>
+              <dd class="govuk-summary-list__actions">
+                <% if !@ans["#{category}"] || @ans["#{category}"]["#{category_assessment['category_assessment_id']}"] == ['none'] %>
```
alexdesi: This condition is a bit too complex, should we move into a helper? 
`!@ans["#{category}"] || @ans["#{category}"]["#{category_assessment['category_assessment_id']}"] == ['none']`

app/views/green_lanes/results/_category_exemption_logic.html.erb
```diff
@@ -0,0 +1,36 @@
...
+
+3, no exemptions
+3, cat 1 met, cat 2 met
+
+
+
+
+cat 2
+
+multiple cat 1's %>
```
alexdesi: should we remove these comments?

PR #1955, opened 2024-07-24
GL-728: Adds handling of redirects to result pages

alexdesi:
### Jira link

https://transformuk.atlassian.net/browse/GL-728

### What?

- [x] Updated the DetermineNextPage service to reflect all of the conditions around applicable exemptions questions being answered
- [x] Updated the return signature of the DetermineNextPage service to just return the URL we should be on
- [x] Changed the DetermineCategories service return signature to include numerical results for the category that applies

### Why?

- This gets us going through the journey to get to the results pages
- This should unblock Sam


app/services/green_lanes/determine_next_page.rb
```diff
@@ -47,3 +49,7 @@ def question_unanswered?(cat_exemptions_apply)
     end
   end
 end
+
+# result.size > 1 -> question page
```
willfish: Am guessing you don't want this :smile:

config/routes.rb
```diff
@@ -115,7 +115,7 @@
       resource :moving_requirements, only: %i[new create]
       resource :applicable_exemptions, only: %i[new create]
       # resource :check_your_answers, only: %i[new create]
-      resource :results, only: %i[show]
+      resources :results, param: :category, only: %i[show]
```
willfish, line 118: Is this a query/positional param constraint? Not sure we need constraints for routes rendered/managed by us?

spec/requests/green_lanes/applicable_exemptions_controller_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+        },
+        category:,
+        commodity_code:,
+        country_of_origin:,
+        moving_date:,
+      }
+    end
+
+    it { is_expected.to have_http_status :redirect }
+    it { is_expected.to redirect_to('/green_lanes/results/1?commodity_code=4114109000&country_of_origin=UA&moving_date=2024-05-29') }
```
willfish: :raised_hands:

app/services/green_lanes/determine_next_page.rb
alexdesi: indeed :smile:

config/routes.rb
alexdesi: This allows us to use:

'/green_lanes/results/1' in the routes
and get the params['category'] == '1'  

for example
---
Not sure about your point, let's chat about it

PR #1959, opened 2024-07-31
GL-842: UR2 Iterations: Start page

alexdesi:
### What?

This is the final version of this screen, which we would like to be operational for the Beta UR.

### Jira link
https://transformuk.atlassian.net/browse/GL-842


shjohnson (approved):
🍏

PR #1961, opened 2024-07-31
GL moving requirements flow fixes

shjohnson:
This PR is to fix some flow issues after last few PR's going in


app/controllers/green_lanes/applicable_exemptions_controller.rb
```diff
@@ -38,22 +33,12 @@ def determine_category
...
-        { authorization: TradeTariffFrontend.green_lanes_api_token },
-      )
+      @goods_nomenclature ||= FetchGoodsNomenclature.new(goods_nomenclature_params).call
     end
 
     # Form handling methods
-    def exemptions_form
-      ApplicableExemptionsForm.new(exemptions_params)
+    def set_exemptions_form
+      @exemptions_form = ApplicableExemptionsForm.new(exemptions_params)
```
alexdesi: keeping the assignment in the action  makes it clear what are the 
instance vars that are used in the views.

This way (before filter, calling a method assigning the instance var) is a bit obscure.

app/helpers/green_lanes_helper.rb
```diff
@@ -31,15 +31,41 @@ def render_exemptions_or_no_card(category, assessments)
...
+    end
+  end
+
   private
 
+  def render_category_3_exemptions(assessments)
+    safe_join([render_exemptions_or_no_card(1, assessments), render_exemptions_or_no_card(2, assessments)])
+  end
+
+  def render_other_category_exemptions(assessments)
```
alexdesi: this is called only my `render_exemptions` if I am not wrong

Should we keep the logic inside render_exemptions?
It'd be easier to test (and we must add tests for it)

app/controllers/green_lanes/results_controller.rb
```diff
@@ -12,20 +12,24 @@ def create
...
+      params[:ans] = parsed_ans if parsed_ans
+
       params.permit(
         :commodity_code,
         :country_of_origin,
         :moving_date,
         :c1ex,
         :c2ex,
-        ans: {},
+        ans: permit_dynamic_ans(parsed_ans),
```
alexdesi: we started using `answers` instead of `ans` to make it clear that it is plural.
(if you are sure it is singular you can use `ans`)

app/controllers/green_lanes/results_controller.rb
```diff
@@ -52,5 +56,14 @@ def c1ex
     def c2ex
       results_params[:c2ex]
     end
+
+    def permit_dynamic_ans(params)
+      if params.is_a?(Hash)
+        params.each_key do |key|
+          params[key] = params[key].is_a?(Hash) ? permit_dynamic_ans(params[key]) : params[key]
```
alexdesi: is this recursive?
this looks complex, we should have at least a test for it.

alexdesi (commented):
Added a couple of comments,
I am a bit worried about the growing number of helper methods, with not tests

spec/services/green_lanes/fetch_goods_nomenclature_spec.rb
```diff
@@ -19,21 +19,27 @@
...
           },
-          authorization: 'test_token',
-        },
-      )
-      expect(result).to eq(expected_response)
+          authorization: TradeTariffFrontend.green_lanes_api_token,
+        )
+      end
+
+      it 'returns the expected response' do
```
alexdesi: I am aware that you did not start this pattern, 
return the `returns the expected response` is not such helpful description :smile_cat:

app/helpers/green_lanes_helper.rb
shjohnson: we can add tests for this

app/controllers/green_lanes/applicable_exemptions_controller.rb
shjohnson: il revert back

app/helpers/green_lanes_helper.rb
shjohnson: I moved out to make easy of breaking it down aswell

spec/services/green_lanes/fetch_goods_nomenclature_spec.rb
```diff
@@ -19,21 +19,27 @@
...
-      result = service.call
+    subject(:fetch_goods) { described_class.new(params).call }
 
-      expect(GreenLanes::GoodsNomenclature).to have_received(:find).with(
-        params[:commodity_code],
-        {
-          filter: {
-            geographical_area_id: params[:country_of_origin],
-            moving_date: params[:moving_date],
+    context 'with valid parameters' do
```
willfish: I can't see we need this context

app/controllers/green_lanes/results_controller.rb
willfish: So I think this is effectively `permit!`

alexdesi (approved):
:white_check_mark:

app/controllers/green_lanes/results_controller.rb
```diff
@@ -25,7 +25,7 @@ def results_params
         :moving_date,
         :c1ex,
         :c2ex,
-        ans: {},
+        :ans,
```
shjohnson, line 28: 👏

PR #1964, opened 2024-08-05
Gl 753 spimm elgibility

shjohnson:
### Jira link

https://transformuk.atlassian.net/browse/GL-753

Adds in the SPIMM eligibility questions to be right after the start page

### DIFF
--- a/app/helpers/green_lanes_helper.rb
+++ b/app/helpers/green_lanes_helper.rb
@@ -54,6 +54,14 @@ module GreenLanesHelper
     end
   end
 
+  def yes_no_options
+    [OpenStruct.new(id: 'yes', name: 'Yes'), OpenStruct.new(id: 'no', name: 'No')]
+  end
+
+  def yes_no_not_sure_options
+    [OpenStruct.new(id: 'yes', name: 'Yes'), OpenStruct.new(id: 'no', name: 'No'), OpenStruct.new(id: 'not_sure', name: 'Not sure')]
+  end
+
   private
 
   def render_all_exemptions(assessments)
--- a/app/services/green_lanes/eligibility_result.rb
+++ b/app/services/green_lanes/eligibility_result.rb
@@ -4,7 +4,8 @@ module GreenLanes
     NOT_YET_ELIGIBLE = 'not_yet_eligible'.freeze
     NOT_ELIGIBLE = 'not_eligible'.freeze
 
-    POSITIVE_RESPONSES = %w[yes not_sure].freeze
+    NOT_SURE = 'not_sure'.freeze
+    YES = 'yes'.freeze
 
     def initialize(params)
       @moving_goods_gb_to_ni = normalize_answer(params[:moving_goods_gb_to_ni])
@@ -23,23 +24,35 @@ module GreenLanes
     private
 
     def not_eligible?
-      !@moving_goods_gb_to_ni || !@free_circulation_in_uk || !@end_consumers_in_uk
+      [@moving_goods_gb_to_ni, @free_circulation_in_uk, @end_consumers_in_uk].any?(&:!)
     end
 
     def not_yet_eligible?
-      all_positive_conditions_met? && !@ukims
+      all_conditions_met_except_ukims? || end_consumers_not_sure?
     end
 
     def may_be_eligible?
-      all_positive_conditions_met? && @ukims
+      all_positive_conditions_met?
     end
 
     def all_positive_conditions_met?
-      @moving_goods_gb_to_ni && @free_circulation_in_uk && @end_consumers_in_uk
+      [@moving_goods_gb_to_ni, @free_circulation_in_uk, @end_consumers_in_uk == true, @ukims].all?
+    end
+
+    def all_conditions_met_except_ukims?
+      [@moving_goods_gb_to_ni, @free_circulation_in_uk, @end_consumers_in_uk].all? && !@ukims
+    end
+
+    def end_consumers_not_sure?
+      @moving_goods_gb_to_ni && @free_circulation_in_uk && @end_consumers_in_uk == NOT_SURE
     end
 
     def normalize_answer(response)
-      POSITIVE_RESPONSES.include?(response.downcase)
+      case response.downcase
+      when YES then true
+      when NOT_SURE then NOT_SURE
+      else false
+      end
     end
   end
 end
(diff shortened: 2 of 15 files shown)


app/helpers/green_lanes_helper.rb
```diff
@@ -54,6 +54,14 @@ def render_exemptions(assessments, category)
     end
   end
 
+  def yes_no_options
+    [OpenStruct.new(id: 'yes', name: 'Yes'), OpenStruct.new(id: 'no', name: 'No')]
```
alexdesi, line 58: nice one :rocket: 

I was just thinking that Struct may be enough, we don't need to expand them with new keys

app/helpers/green_lanes_helper.rb
```diff
@@ -54,6 +54,14 @@ def render_exemptions(assessments, category)
...
+  end
+
+  def yes_no_not_sure_options
+    [OpenStruct.new(id: 'yes', name: 'Yes'), OpenStruct.new(id: 'no', name: 'No'), OpenStruct.new(id: 'not_sure', name: 'Not sure')]
```
alexdesi, line 62: same here
```
 [
  Struct.new(id: 'yes', name: 'Yes'), 
  Struct.new(id: 'no', name: 'No'),
  Struct.new(id: 'not_sure', name: 'Not sure')
]

```

app/services/green_lanes/eligibility_result.rb
```diff
@@ -0,0 +1,58 @@
+module GreenLanes
+  class EligibilityResult
+    YOU_MAY_BE_ELIGIBLE = 'you_may_be_eligible'.freeze
+    NOT_YET_ELIGIBLE = 'not_yet_eligible'.freeze
+    NOT_ELIGIBLE = 'not_eligible'.freeze
+
+    NOT_SURE = 'not_sure'.freeze
+    YES = 'yes'.freeze
```
alexdesi, line 8: is `NO` missing?

spec/forms/green_lanes/eligibility_form_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+          free_circulation_in_uk: 'yes',
+          end_consumers_in_uk: 'yes',
+          ukims: 'yes',
+        }
+      end
+
+      it { expect(form.errors).to be_empty }
+    end
+
+    context 'when an commodity_code is missing' do
```
alexdesi: `when a**n** commodity_code is missing`
or it should be
`when the commodity_code is missing`

spec/requests/green_lanes/eligibilities_controller_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      green_lanes_eligibility_form: {
+        moving_goods_gb_to_ni: 'yes',
+        free_circulation_in_uk: 'yes',
+        end_consumers_in_uk: 'yes',
+        ukims: 'yes',
+      },
+    }
+  end
+
+  let(:invalid_params) do
```
alexdesi: we can move this inside the context
"with invalid params"

spec/requests/green_lanes/eligibilities_controller_spec.rb
```diff
@@ -0,0 +1,65 @@
+require 'spec_helper'
+
+RSpec.describe GreenLanes::EligibilitiesController, type: :request do
+  subject { make_request && response }
+
+  before do
+    allow(TradeTariffFrontend).to receive(:green_lanes_enabled?).and_return true
+  end
+
+  let(:valid_params) do
```
alexdesi: maybe we should move this inside the appropriate context

app/services/green_lanes/eligibility_result.rb
```diff
@@ -0,0 +1,58 @@
...
+      return NOT_YET_ELIGIBLE if not_yet_eligible?
+
+      NOT_ELIGIBLE
+    end
+
+    private
+
+    def not_eligible?
+      [@moving_goods_gb_to_ni, @free_circulation_in_uk, @end_consumers_in_uk].any?(&:!)
+    end
```
alexdesi, line 28: I'd suggest to use straigh logic:

```
def eligible?
  @moving_goods_gb_to_ni && @free_circulation_in_uk && @end_consumers_in_uk
end
```
and using `!eligible?` if we need the not_eligible

alexdesi (commented):
It is very good,  I've only added minor comments

app/helpers/green_lanes_helper.rb
shjohnson: Yeah possibly ive not tested to be honest, i cant remember why i used this pattern in the past

app/services/green_lanes/eligibility_result.rb
shjohnson: i could add no your right and might make more sense for reading 👍

app/services/green_lanes/eligibility_result.rb
shjohnson: I know weve discussed this in private so happy to change to 

`!@moving_goods_gb_to_ni || !@free_circulation_in_uk || !@end_consumers_in_uk` if preferred

spec/forms/green_lanes/eligibility_form_spec.rb
shjohnson: good catch and yes the bottom i think i prefer

spec/requests/green_lanes/eligibilities_controller_spec.rb
shjohnson: yeah i can have a look at this i think this is here how i first had the tests set up

app/services/green_lanes/eligibility_result.rb
shjohnson: actually its not used so dont need this here to be honest..

app/helpers/green_lanes_helper.rb
willfish: I think these relate more to the form than the green lanes helper?

app/services/green_lanes/eligibility_result.rb
```diff
@@ -0,0 +1,58 @@
...
+    def call
+      return YOU_MAY_BE_ELIGIBLE if may_be_eligible?
+      return NOT_YET_ELIGIBLE if not_yet_eligible?
+
+      NOT_ELIGIBLE
+    end
+
+    private
+
+    def not_eligible?
```
willfish, line 26: This isn't actually used anywhere

willfish (approved):

PR #1965, opened 2024-08-05
GL-854: fix from QA testing.

alexdesi:
### What?

Fix from QA testing.

I have added/removed/altered:

- [ ] Added foo in bar
- [ ] Removed baz

### Why?
QA spotted some diff from the Jira spec that must be fixed.


willfish (approved):

shjohnson (approved):
🟢

PR #1968, opened 2024-08-06
Gl 862 

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/GL-862

### What?
    - Various UR2 design improvement
    - Add the classification to the "Your goods" section in the result page.
    - Refactored logic into the classifiable concern.

### Note
this is part of the GL-862,
more changes on it are coming.

### DIFF
--- a/app/models/concerns/classifiable.rb
+++ b/app/models/concerns/classifiable.rb
@@ -1,0 +1,34 @@
+module Classifiable
+  extend ActiveSupport::Concern
+
+  def descriptions_with_other_handling
+    return [] unless to_s.match(/^other$/i)
+
+    all_other = true
+    descriptions = []
+
+    ancestors.reverse.each do |ancestor|
+      if ancestor.is_other?
+        descriptions.unshift(ancestor.to_s)
+      else
+        descriptions.unshift(ancestor.to_s)
+        all_other = false
+        break
+      end
+    end
+
+    if all_other
+      descriptions.unshift(heading.to_s)
+    end
+
+    descriptions
+  end
+
+  def is_other?
+    description.match(/^other$/i)
+  end
+
+  def to_s
+    formatted_description || description
+  end
+end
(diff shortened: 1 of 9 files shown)


app/models/concerns/classifiable.rb
```diff
@@ -0,0 +1,34 @@
+module Classifiable
```
shjohnson, line 1: Is this tested anywhere like in another class as side affect?

app/views/green_lanes/results/_result_category_1.html.erb
```diff
@@ -1,36 +1,54 @@
 <div class="govuk-panel govuk-panel--confirmation">
   <h1 class="govuk-panel__title">
-    Your goods are Category 1
```
shjohnson, line 3: Should this be changed?

app/views/green_lanes/shared/_summary_card.html.erb
```diff
@@ -11,6 +11,12 @@
           <dt class="govuk-summary-list__key">Commodity</dt>
           <dd class="govuk-summary-list__value"><%= @commodity_code %></dd>
         </div>
+        <div class="govuk-summary-list__row">
```
shjohnson, line 14: 👏

shjohnson (commented):
Happy to approve but there is one comment about testing on the classification class

app/models/concerns/classifiable.rb
alexdesi: Yes,  good point. This is tested in:
`spec/models/commodity_spec.rb`

willfish: And as it gets rendered in the views

app/models/green_lanes/goods_nomenclature.rb
```diff
@@ -13,9 +14,7 @@ class GreenLanes::GoodsNomenclature
 
   has_many :applicable_category_assessments, class_name: 'GreenLanes::CategoryAssessment'
 
-  def all(_opts = {})
```
willfish, line 16: This makes me so happy

willfish (approved):

PR #1972, opened 2024-08-08
Cleanup branch

shjohnson:
As the name suggests this is a cleanup branch!

- Removes the edit text from buttons which will get added back in a future branch maybe next week!
- Altered and created logic for the back links in the CYA pages and the exemption question pages
- Remove method which was left over from another branch
- Changed the CYA page to be a show action as it makes more sense


app/controllers/green_lanes/applicable_exemptions_controller.rb
```diff
@@ -7,24 +7,58 @@ class ApplicableExemptionsController < ApplicationController
...
+      if category == 2
+        new_green_lanes_applicable_exemptions_path(
+          category: 1,
+          commodity_code: permitted_params[:commodity_code],
+          country_of_origin: permitted_params[:country_of_origin],
+          moving_date: permitted_params[:moving_date],
+          ans: parsed_ans(permitted_params[:ans]),
+          c1ex: permitted_params[:c1ex],
+        )
+      else
```
alexdesi, line 43: to understand: what is this case?
I guess  category == 1,
maybe we can add a comment

app/controllers/green_lanes/applicable_exemptions_controller.rb
```diff
@@ -7,24 +7,58 @@ class ApplicableExemptionsController < ApplicationController
...
+      end
+    end
+
+    def parsed_ans(ans_param)
+      return ans_param unless ans_param.is_a?(String)
+
+      begin
+        JSON.parse(ans_param)
+      rescue JSON::ParserError
+        ans_param
```
alexdesi: so if you get an error you return the `parsed_ans` unparsed?
why is that?

app/controllers/green_lanes/applicable_exemptions_controller.rb
```diff
@@ -7,24 +7,58 @@ class ApplicableExemptionsController < ApplicationController
...
+        new_green_lanes_moving_requirements_path(
+          commodity_code: params[:commodity_code],
+          country_of_origin: params[:country_of_origin],
+          moving_date: params[:moving_date],
+        )
+      end
+    end
+
+    def parsed_ans(ans_param)
+      return ans_param unless ans_param.is_a?(String)
```
alexdesi, line 53: Why ans_ param could not be a string?

app/controllers/green_lanes/check_your_answers_controller.rb
```diff
@@ -15,10 +15,42 @@ def new
...
       @c2ex = check_your_answers_params[:c2ex]
+      @back_link_path = determine_back_link_path(check_your_answers_params)
     end
 
     private
 
+    def determine_back_link_path(permitted_params)
+      cat_2_questions_exist = !permitted_params[:ans]['2'].nil?
+
+      if all_exemptions_met?(1, @category_one_assessments, @answers) && cat_2_questions_exist
```
alexdesi: why does the backlink need to know if all exemptions are met?

alexdesi (dismissed):
good improvements
a little concern is on the `determine_back_link_path`
it's not clear to me  why we are using those condition in the ifs.

It would not be easier just to have different backlink methods?

app/controllers/green_lanes/applicable_exemptions_controller.rb
shjohnson: So this is because its used for the cat 1 and 2 questions. the bottom case is if you are in the cat1 questions and click back so you go back to the good details page.

the top one is if you are on cat 2 questions then you go back to the cat 1 questions

app/controllers/green_lanes/applicable_exemptions_controller.rb
shjohnson: Yeah possibly i can skip this part because its virtually impossible to happen will remove

app/controllers/green_lanes/check_your_answers_controller.rb
shjohnson: Yeah thinking about this now maybe they could be removed im trying to remember why i put them there. i guess if cat 2 questions exist then fine can go to them

alexdesi (approved):
:white_check_mark:

PR #1976, opened 2024-08-08
Add protection when green lanes not enabled

shjohnson:
### Jira link

Add the protection for if the green lanes is not enabled production


alexdesi (approved):
:white_check_mark: :rocket:

PR #1978, opened 2024-08-12
Fix some broken logic with at least one exemption not showing for a CA

shjohnson:
https://transformuk.atlassian.net/browse/GL-923

This PR handles some of the broken logic when if a commodity returns at least one Ca without an exemption the user should be skipped to the CYA page and not the exemption questions page.

Also fixes an error where no answers are provided in the exemptions questionaire

### DIFF
--- a/app/controllers/green_lanes/check_your_answers_controller.rb
+++ b/app/controllers/green_lanes/check_your_answers_controller.rb
@@ -21,34 +21,27 @@ module GreenLanes
     private
 
     def determine_back_link_path(permitted_params)
-      cat_2_questions_exist = !permitted_params[:ans]['2'].nil?
-      cat_1_questions_dont_exist = permitted_params[:ans]['1'].nil?
+      ans = permitted_params[:ans]
+      category = if ans.nil?
+                   1
+                 elsif ans['2'].present?
+                   2
+                 else
+                   1
+                 end
 
-      if cat_2_questions_exist
-        new_green_lanes_applicable_exemptions_path(
-          category: 2,
-          commodity_code: permitted_params[:commodity_code],
-          country_of_origin: permitted_params[:country_of_origin],
-          moving_date: permitted_params[:moving_date],
-          ans: permitted_params[:ans],
-          c1ex: permitted_params[:c1ex],
-          c2ex: permitted_params[:c2ex],
-        )
-      elsif cat_1_questions_dont_exist
-        new_green_lanes_moving_requirements_path(
-          commodity_code: permitted_params[:commodity_code],
-          country_of_origin: permitted_params[:country_of_origin],
-          moving_date: permitted_params[:moving_date],
-        )
+      base_params = {
+        commodity_code: permitted_params[:commodity_code],
+        country_of_origin: permitted_params[:country_of_origin],
+        moving_date: permitted_params[:moving_date],
+      }
+
+      if category == 2
+        new_green_lanes_applicable_exemptions_path(base_params.merge(category:, ans:, c1ex: permitted_params[:c1ex], c2ex: permitted_params[:c2ex]))
+      elsif ans.nil? || ans['1'].nil?
+        new_green_lanes_moving_requirements_path(base_params)
       else
-        new_green_lanes_applicable_exemptions_path(
-          category: 1,
-          commodity_code: permitted_params[:commodity_code],
-          country_of_origin: permitted_params[:country_of_origin],
-          moving_date: permitted_params[:moving_date],
-          ans: permitted_params[:ans],
-          c1ex: permitted_params[:c1ex],
-        )
+        new_green_lanes_applicable_exemptions_path(base_params.merge(category:, ans:, c1ex: permitted_params[:c1ex]))
       end
     end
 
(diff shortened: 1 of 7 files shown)


app/controllers/green_lanes/results_controller.rb
```diff
@@ -29,6 +30,12 @@ def results_params
       )
     end
 
+    def cas_without_exemptions
```
willfish, line 33: I guess cas is category assessments :thinking:

shjohnson: Yeah sorry i can change that if needed

app/views/green_lanes/results/_result_category_3.html.erb
```diff
@@ -64,4 +64,6 @@
 
 <%= render 'green_lanes/shared/about_your_goods_card', title: 'Your movement of goods', edit_path: 'EDIT GOOD DETAILS PAGE' %>
 
-<%= render_exemptions(@assessments, @category) %>
+<% unless @answers.nil? %>
```
shjohnson, line 67: If the user is cat 3 and hasnt been shown any answers then we dont show the below sections

willfish (approved):
No testss :scream:

app/views/green_lanes/results/_category_assessments_card.erb
```diff
@@ -7,25 +7,24 @@
 
   <div class="govuk-summary-card__content">
     <dl class="govuk-summary-list">
-      <% @assessments.public_send("cat_#{category}_exemptions_not_met").each do |category_assessment| %>
+      <% assessments = @cas_without_exemptions.present? ? @cas_without_exemptions : @assessments.public_send("cat_#{category}_exemptions_not_met") %>
```
shjohnson, line 10: If there are any category assessments without exemptions then we show them and say `no exemptions available`

app/controllers/green_lanes/results_controller.rb
willfish: No strong view. :raised_hands:

app/views/green_lanes/check_your_answers/show.html.erb
```diff
@@ -11,14 +11,16 @@
                edit_path: 'EDIT GOOD DETAILS PAGE'
     %>
 
-    <h2 class="govuk-heading-m">Category exemptions</h2>
+    <% unless @category_one_assessments.empty? || @answers.nil? %>
```
shjohnson, line 14: Again if the user has not answered anything then we dont show this section

app/services/green_lanes/determine_next_page.rb
```diff
@@ -31,6 +31,7 @@ def single_category?
     end
 
     def handle_cat1_cat2(cat_1_exemptions_apply)
+      return check_your_answers if @determine_category.cat1_without_exemptions.present?
```
shjohnson, line 34: If any cat 1 assessments dont have exemptions then they have to go straight to the CYA page as they cant be exemption from them

app/presenters/green_lanes/assessments_presenter.rb
```diff
@@ -27,10 +27,14 @@ def cat_2_exemptions
     end
 
     def cat_1_exemptions_met
+      return [] if @answers.nil?
```
shjohnson, line 30: just a guard clause if no answers have been given

app/controllers/green_lanes/results_controller.rb
```diff
@@ -29,6 +30,12 @@ def results_params
...
+      return [] if @category == '3'
```
shjohnson, line 34: another guard clause for when the user goes straight to a cat 3 result

app/controllers/green_lanes/check_your_answers_controller.rb
```diff
@@ -21,34 +21,27 @@ def show
     private
 
     def determine_back_link_path(permitted_params)
-      cat_2_questions_exist = !permitted_params[:ans]['2'].nil?
-      cat_1_questions_dont_exist = permitted_params[:ans]['1'].nil?
+      ans = permitted_params[:ans]
+      category = if ans.nil?
```
alexdesi, line 25: `dig` simplifies the retrieval of nested values, it seems a good use here:

```
if  ans&.dig('2').present? 
  2
else
  1
end
```

app/views/green_lanes/check_your_answers/show.html.erb
alexdesi, line 14: two things
- We've been named it using numbers until now, to be consistent let's call it:
`@category_1_assessments`

-  it looks the logic is same as this one (that should be more readable):

`if  @category_one_assessments.present? && @answers.present?`

app/views/green_lanes/results/_result_category_3.html.erb
alexdesi, line 67: I guess we can do the opposite:

```
<% if @anwers.present? %>
....

````

app/views/green_lanes/check_your_answers/show.html.erb
shjohnson: I can change this but it was used anyways @category_one_assessments.

Im not sure that logic would work as there could be cat 1 assessments but the user hasnt been shown them because at least one of them has no exemptions hence adding the or clause

app/controllers/green_lanes/check_your_answers_controller.rb
shjohnson: ok will try

app/views/green_lanes/check_your_answers/show.html.erb
alexdesi: yeah, I think I made a mistake in my  logic

PR #1989, opened 2024-08-13
Fix multiple exemptions

shjohnson:
### Jira link

https://transformuk.atlassian.net/browse/GL-957

### DIFF
--- a/app/forms/green_lanes/applicable_exemptions_form.rb
+++ b/app/forms/green_lanes/applicable_exemptions_form.rb
@@ -5,7 +5,7 @@ module GreenLanes
     include ActiveRecord::AttributeAssignment
 
     NO_ANSWER_WAS_GIVEN = [''].freeze
-    NONE_ANSWER_WAS_GIVEN = ['', 'none'].freeze
+    NONE_ANSWER_WAS_GIVEN = 'none'.freeze
 
     attr_reader :answers, :category_assessments
 
@@ -38,7 +38,7 @@ module GreenLanes
     end
 
     def exempt?
-      answers.values.none? { |value| value == NONE_ANSWER_WAS_GIVEN }
+      !answers.values.flatten.include?(NONE_ANSWER_WAS_GIVEN)
     end
 
     def presented_answers
--- a/app/helpers/green_lanes_helper.rb
+++ b/app/helpers/green_lanes_helper.rb
@@ -7,15 +7,19 @@ module GreenLanesHelper
     exemption_checkbox_checked?(category_assessment_id, 'none')
   end
 
-  def render_exemptions_or_no_card(category, assessments)
+  def render_exemptions_or_no_card(category, assessments, result)
     no_exemptions = assessments.public_send("no_cat#{category}_exemptions")
     exemptions_met = assessments.public_send("cat_#{category}_exemptions_met")
     total_exemptions = assessments.public_send("cat_#{category}_exemptions")
 
     all_exemptions_met = total_exemptions.count == exemptions_met.count
 
-    template = no_exemptions || !all_exemptions_met ? 'category_assessments_card' : 'exemptions_card'
-    render(template, category:)
+    if result == '3'
+      render('exemptions_card', category:)
+    else
+      template = no_exemptions || !all_exemptions_met ? 'category_assessments_card' : 'exemptions_card'
+      render(template, category:)
+    end
   end
 
   def exemption_met?(exemption_code, category, category_assessment, answers)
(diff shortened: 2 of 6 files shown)


app/forms/green_lanes/applicable_exemptions_form.rb
```diff
@@ -38,7 +38,7 @@ def initialize(attributes)
     end
 
     def exempt?
-      answers.values.none? { |value| value == NONE_ANSWER_WAS_GIVEN }
+      !answers.values.flatten.include?(NONE_ANSWER_WAS_GIVEN)
```
shjohnson, line 41: Changed this because sometimes if multiple answers we were getting `['','','none']` so if it contains none it counts as a none answer

app/helpers/green_lanes_helper.rb
```diff
@@ -1,59 +1,52 @@
...
-    cas_checked = category_assessments_checked(category_assessment_id)
-
-    return if cas_checked.nil?
-
-    cas_checked.include?('none')
+    exemption_checkbox_checked?(category_assessment_id, 'none')
   end
 
-  def render_exemptions_or_no_card(category, assessments)
+  def render_exemptions_or_no_card(category, assessments, result)
```
shjohnson, line 10: passing the result in mainly for cat 3;

There are some cc's which have no cat 1's and they still could be standard category

app/helpers/green_lanes_helper.rb
```diff
@@ -1,59 +1,52 @@
...
+      render('exemptions_card', category:)
     else
-      render 'exemptions_card', category:
+      template = no_exemptions || !all_exemptions_met ? 'category_assessments_card' : 'exemptions_card'
+      render(template, category:)
     end
   end
 
-  def exemption_met?(category, category_assessment, answers)
+  def exemption_met?(exemption_code, category, category_assessment, answers)
```
shjohnson, line 25: We need to be able to tell for the actual exemption if its passed not the whole assessment

app/helpers/green_lanes_helper.rb
```diff
@@ -1,59 +1,52 @@
...
-      result = render_exemptions_or_no_card(1, assessments)
-      result += render_exemptions_or_no_card(2, assessments) unless assessments.cat_1_exemptions_not_met
-      result
+    category_assessments.all? do |ca|
+      category_assessment_answer = dig_category_answer(answers, category, ca.category_assessment_id)
+      category_assessment_answer.present? && category_assessment_answer != %w[none]
     end
   end
 
+  def render_exemptions(assessments, result)
```
shjohnson, line 41: This refactor works better for all scenerios

app/views/green_lanes/results/_category_assessments_card.erb
```diff
@@ -1,7 +1,7 @@
 <div class="govuk-summary-card">
   <div class="govuk-summary-card__title-wrapper">
     <h2 class="govuk-summary-card__title">
-      Why your goods are Category <%= category %>
+      Your Category <%= category %> exemptions
```
shjohnson, line 4: This is the new wording that i think was missed on the UR2 result screens

app/views/green_lanes/results/_exemptions_card.html.erb
```diff
@@ -8,7 +8,7 @@
   <div class="govuk-summary-card__content">
     <div class="govuk-summary-list">
       <% @assessments.public_send("cat#{category}_with_exemptions").each do |category_assessment| %>
-        <% category_assessment.exemptions.each do |exemption| %>
+        <% category_assessment.exemptions.select { |exemption| dig_category_answer(@answers, category, category_assessment.category_assessment_id).include?(exemption.code) }.each do |exemption| %>
```
shjohnson, line 11: This is just filtering to only show the exemptions that have been met when you have multiple for a category assessment

app/helpers/green_lanes_helper.rb
```diff
@@ -69,9 +62,14 @@ def render_all_exemptions(assessments)
   end
 
   def category_assessments_checked(category_assessment_id)
-    exemptions = params[:exemptions]
-    return if exemptions.nil? || exemptions[:category_assessments_checked].nil?
+    params.dig(:exemptions, :category_assessments_checked, category_assessment_id.to_s)
+  end
+
+  def dig_category_answer(answers, category, category_assessment_id)
```
alexdesi, line 68: maybe we don't need this method,
it's just a dig and it is called `dig_category_answer` :thinking:

app/views/green_lanes/results/_exemptions_card.html.erb
alexdesi: a bit complex for a view, maybe we should move it into an explicit method
`def exemptions_met_for_multiple_category_assessment`

alexdesi (approved):
it looks good,
just one suggestion!

app/views/green_lanes/results/_exemptions_card.html.erb
shjohnson: Yeah i agree, to be honest this part of the code i need to double check with Chris and will see what he says after the UR sessions. its something i dont think is portrayed very easily on the tickets.

PR #1990, opened 2024-08-14
GL-933: Add feature tests for scenario 7..10

alexdesi:
### What?

Add feature tests for scenarios 7..10

### Why?
Because we care


willfish (dismissed):

shjohnson (approved):
🟢 🎸 🚀

PR #1991, opened 2024-08-14
OTT-348 Re-Write javascript popup into stimulus modal

HWallenberg:
### Jira link

https://transformuk.atlassian.net/browse/OTT-348

### What?

I have

- [x] Removed code for popup including the jquery and styling
- [x] Added stimulus controllers for an Anchor controlled Modal element for use in the commodities tables for Conditions & Footnotes

### Why?

I am doing this because:

- This is part of the migration of the JQuery commodities.js file into stimulus controllers
- 
### Have you? (optional)

image


app/views/measures/_measure.html.erb
```diff
@@ -51,11 +51,15 @@
...
   <td class="conditions-col govuk-table__cell">
-    <% if measure.has_measure_conditions? %>
-      <%= link_to "Conditions",  "#", class: 'reference', 'data-popup-ref' => "#{measure.destination}-#{measure.id}-conditions", "aria-label": "Click to open a dialog with the conditions for this measure", role: "button" %>
-    <% elsif show_proofs_for_geographical_areas?(roo_schemes, measure) %>
-      <%= link_to "Conditions",  "#", class: 'reference', 'data-popup-ref' => "#{measure.destination}-#{measure.id}-cds-proofs", "aria-label": "Click to open a dialog with the conditions for this measure", role: "button" %>
-    <% end %>
+    <div data-controller="anchor">
+      <% if measure.has_measure_conditions? %>
+        <%= link_to "Conditions", "#", class: 'reference', data: { action: "click->anchor#launchModal", modal_ref: "#{measure.destination}-#{measure.id}-conditions" }, "aria-label": "Click to open a dialog with the conditions for this measure", role: "button" %>
+        <div id="modal" class="modal tariff-info" data-controller="modal" data-anchor-target="modal"></div>
```
neilmiddleton: Do we need one of these within each render of `_measure` when there's only a possibility of one modal at once?  If we can move this into some higher level spot we'll only need the one with the same Id.

app/views/measures/_measure.html.erb
```diff
@@ -51,11 +51,15 @@
...
+      <% elsif show_proofs_for_geographical_areas?(roo_schemes, measure) %>
+        <%= link_to "Conditions", "#", class: 'reference', data: { action: "click->anchor#launchModal", modal_ref: "#{measure.destination}-#{measure.id}-cds-proofs" }, "aria-label": "Click to open a dialog with the conditions for this measure", role: "button" %>
+        <div id="modal" class="modal tariff-info" data-controller="modal" data-anchor-target="modal"></div>
```
neilmiddleton: same modal div here

app/views/measures/_measure.html.erb
```diff
@@ -64,7 +68,10 @@
 
   <td class="footnotes-col govuk-table__cell govuk-table__header--numeric">
     <% if measure.has_measure_footnotes? %>
-      <%= link_to measure.footnotes.size == 1 ? measure.footnotes.first.code : "Footnotes",  "#", class: 'reference', 'data-popup-ref' => "#{measure.destination}-#{measure.id}-footnotes", "aria-label": "Click to open a dialog with the footnotes for this measure", role: "button" %>
+      <div data-controller="anchor">
+        <%= link_to measure.footnotes.size == 1 ? measure.footnotes.first.code : "Footnotes", "#", class: 'reference', data: { action: "click->anchor#launchModal", modal_ref: "#{measure.destination}-#{measure.id}-footnotes" }, "aria-label": "Click to open a dialog with the footnotes for this measure", role: "button" %>
+        <div id="modal" class="modal tariff-info" data-controller="modal" data-anchor-target="modal"></div>
```
neilmiddleton, line 72: and again

app/webpacker/src/stylesheets/_modal.scss
neilmiddleton, line 1: I wonder if everything in here should be ideally [scoped](https://developer.chrome.com/docs/css-ui/at-scope) under `modal` in some way to avoid things like `fade` _leaking_ into other stuff:

```
@scope (#modal) {
 ...
}
```

not a big issue though, just a little safer when you're using broad class names.

neilmiddleton (commented):
Functionally this seems fine, just a couple of small things I would look at potentially improving on a bit.  Styling wise the only small change I would make is to put a small black border on the modal as that seems to be the convention on other things.

Screenshot 2024-08-16 at 16 42 15

app/views/measures/_measure.html.erb
HWallenberg: Modal Div must sit within an Anchor Div.  Attempted to move Anchor Div to higher level than row cell but Anchor Controller was out of scope for modal controller to launch modal.  So have moved one of the Modal Divs out of the if statement.

app/views/measures/_measure.html.erb
HWallenberg: This modal div must sit within the anchor div which cannot move outside the scope of the row cell.

HWallenberg: see previous comment

app/views/measures/_measure.html.erb
HWallenberg: This modal div must sit within the anchor div which cannot move outside the scope of the row cell.

app/webpacker/src/stylesheets/_modal.scss
HWallenberg: I tried moving the scss under a #modal { }scope however the styling was then lost when opening the modal.  I believe this to be as the fade and show are part of the whole document and cannot be in just the scope of the <div id='moda'l>

HWallenberg:
@neilmiddleton added black border as suggested please see other comments, hope this is okay, happy to have a session to resolve anything, thanks.

neilmiddleton (dismissed):
LGTM 🎉

neilmiddleton (approved):

PR #1995, opened 2024-08-19
GL-792: GL Fix GL Commodity Form

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/GL-792

### What?

-[x] Fix service to get the commodity: now the commodity is always declarable
-[x] Add relative validation to check the commodity is declarable.
-[x] Change the wording for the validation.
-[x]  FetchGoodNomenclature service uses VCR cassette now.  

### Why?
To fix errors and improve usability.


app/forms/green_lanes/moving_requirements_form.rb
```diff
@@ -17,16 +17,9 @@ class MovingRequirementsForm
...
     def commodity_code_exists
-      GreenLanes::GoodsNomenclature.find(
-        commodity_code,
-        {
-          filter: { geographical_area_id: country_of_origin },
-          as_of: moving_date,
-        },
-        { authorization: TradeTariffFrontend.green_lanes_api_token },
-      )
+      FetchGoodsNomenclature.new(commodity_code:, country_of_origin:, moving_date:).call
```
shjohnson, line 20: 👍

app/services/green_lanes/fetch_goods_nomenclature.rb
```diff
@@ -18,6 +18,8 @@ def call
           authorization: TradeTariffFrontend.green_lanes_api_token,
         },
       )
+
+      goods_nomenclature.get_declarable
```
shjohnson, line 22: I guess this works fine for all the times we use this service?

shjohnson (approved):
Looks good just one thing to double confirm but i couldnt think of a place where it wouldnt need to be declarable

app/services/green_lanes/fetch_goods_nomenclature.rb
alexdesi: yes, if the goods is already declarable, it will be returned as it is:

```
  def get_declarable
    return self if declarable?

    descendants.find(&:declarable?)
end
```

PR #1996, opened 2024-08-19
OTT-354 Remove dead code for `js-country-picker`

neilmiddleton:
### Jira link

OTT-354

### What?

I have removed code which appeared to be doing nothing.  This ticket is talking about adding stimulus to the country picker, but from looking at it this is already done, but a load of the old code was still in place.  A search of the entire codebase showed that this code wasn't used anywhere else.

Would be handy to have @HWallenberg for a second pair of eyes.

### Why?

In order to remove dead old code

### Have you? (optional)


app/webpacker/src/javascripts/commodities.js
```diff
@@ -535,7 +434,7 @@
         @description adds behaviours
         @param {Element} content Element in which to add behaviours
       */
-    onLoad: function(context) {
+    onLoad: function (context) {
       if (context === undefined) {
         context = document.body;
       }
```
HWallenberg: Looks good but please can you remove `this.countryPicker.initialize();` from line 445 as it throws a js error, thanks. :) 
image

neilmiddleton:
@HWallenberg Should be good to go now.

HWallenberg (dismissed):
lgtm 🙌

HWallenberg (approved):

PR #2000, opened 2024-08-21
GL-925: Fix GL check eligibility start links.

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/GL-925

### What?
Fix GL check eligibility start links.
The XI of the GL link does not exist.

### Why?
To fix broken links

### DIFF
--- a/app/helpers/green_lanes_helper.rb
+++ b/app/helpers/green_lanes_helper.rb
@@ -63,6 +63,10 @@ module GreenLanesHelper
     [OpenStruct.new(id: 'yes', name: 'Yes'), OpenStruct.new(id: 'no', name: 'No'), OpenStruct.new(id: 'not_sure', name: 'Not sure')]
   end
 
+  def green_lanes_eligibility_start_path
+    '/green_lanes/start/new'
+  end
+
   private
 
   def category_assessments_checked(category_assessment_id)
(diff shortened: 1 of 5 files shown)


app/helpers/green_lanes_helper.rb
```diff
@@ -63,6 +63,10 @@ def yes_no_not_sure_options
     [OpenStruct.new(id: 'yes', name: 'Yes'), OpenStruct.new(id: 'no', name: 'No'), OpenStruct.new(id: 'not_sure', name: 'Not sure')]
   end
 
+  def green_lanes_eligibility_start_path
```
shjohnson, line 66: Just thought, do we start changing the method name now? something like;

`def spimm_eligibility_start_path`

shjohnson (commented):
I think also we need to fix:

`app/views/pages/_howto_sidebar_links.html.erb`

app/helpers/green_lanes_helper.rb
alexdesi: good point, but it's probably better to create a ticket to rename all of those,
there are 267 occurrences of 'green lanes' !
so I'd avoid to create mixed notation

shjohnson (approved):
💚

PR #2001, opened 2024-08-21
Assessed exemptions

shjohnson:
### Jira link

https://transformuk.atlassian.net/browse/GL-948
https://transformuk.atlassian.net/browse/GL-950
https://transformuk.atlassian.net/browse/GL-977

Bugs fixes mainly around the results pages displaying the wrong boxes at the bottom, this is mainly to do with how we were counting up which exemptions had been selected instead of looking at the category assessments we had.

### DIFF
--- a/app/controllers/green_lanes/check_your_answers_controller.rb
+++ b/app/controllers/green_lanes/check_your_answers_controller.rb
@@ -11,7 +11,7 @@ module GreenLanes
       @country_of_origin = check_your_answers_params[:country_of_origin] || GeographicalArea::ERGA_OMNES
       @moving_date = check_your_answers_params[:moving_date]
       @category_one_assessments = determine_category.cat1_with_exemptions
-      @category_two_assessments_without_exemptions = determine_category.cat2_without_exemptions.present?
+      @category_two_assessments_without_exemptions = determine_category.cat2_without_exemptions
       @category_two_assessments = determine_category.cat2_with_exemptions
       @answers = check_your_answers_params[:ans]
       @c1ex = check_your_answers_params[:c1ex]
--- a/app/controllers/green_lanes/check_your_answers_controller.rb
+++ b/app/controllers/green_lanes/check_your_answers_controller.rb
@@ -11,6 +11,7 @@ module GreenLanes
       @country_of_origin = check_your_answers_params[:country_of_origin] || GeographicalArea::ERGA_OMNES
       @moving_date = check_your_answers_params[:moving_date]
       @category_one_assessments = determine_category.cat1_with_exemptions
+      @category_two_assessments_without_exemptions = determine_category.cat2_without_exemptions.present?
       @category_two_assessments = determine_category.cat2_with_exemptions
       @answers = check_your_answers_params[:ans]
       @c1ex = check_your_answers_params[:c1ex]
--- a/app/helpers/green_lanes_helper.rb
+++ b/app/helpers/green_lanes_helper.rb
@@ -43,13 +43,13 @@ module GreenLanesHelper
 
     case result
     when '1'
-      view << render_exemptions_or_no_card(1, assessments, result) if assessments.public_send('cat_1_exemptions').present? || @cas_without_exemptions.present?
+      view << render_exemptions_or_no_card(1, assessments, result) if assessments.send('cat_1_exemptions').present? || @cas_without_exemptions.present?
     when '2'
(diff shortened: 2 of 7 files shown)


app/helpers/green_lanes_helper.rb
```diff
@@ -8,16 +8,16 @@ def exemption_checkbox_none?(category_assessment_id)
...
   def render_exemptions_or_no_card(category, assessments, result)
-    no_exemptions = assessments.public_send("no_cat#{category}_exemptions")
-    exemptions_met = assessments.public_send("cat_#{category}_exemptions_met")
-    total_exemptions = assessments.public_send("cat_#{category}_exemptions")
+    no_exemptions = assessments.send("no_cat#{category}_exemptions")
+    assessments_met = assessments.send("cat_#{category}_assessments_met")
+    total_assessments = assessments.send("cat_#{category}_assessments").pluck(:category_assessment_id).map(&:to_s)
 
-    all_exemptions_met = total_exemptions.count == exemptions_met.count
+    all_assessments_met = total_assessments.count == assessments_met.count
```
shjohnson, line 15: This needed to change to count the assessments rather than the exemptions as some assessments have multiple exemptions where only one exemption needs to count

app/views/green_lanes/check_your_answers/show.html.erb
```diff
@@ -22,7 +22,7 @@
       %>
     <% end %>
 
-    <% if all_exemptions_met?(1, @category_one_assessments, @answers) %>
+    <% if all_exemptions_met?(1, @category_one_assessments, @answers) && !@category_two_assessments_without_exemptions %>
```
shjohnson: We need to show this box when there are cat 2 assessments without exemptions not just when all assessments are met in cat 1

app/views/green_lanes/results/_exemptions_card.html.erb
```diff
@@ -15,8 +15,6 @@
             </dt>
             <dd class="govuk-summary-list__value">
               <%= exemption.formatted_description %>
-              <br><br>
```
shjohnson, line 18: we dont want this line anymore

app/presenters/green_lanes/assessments_presenter.rb
```diff
@@ -26,6 +26,14 @@ def cat_2_exemptions
       extract_exemptions(cat2_with_exemptions)
     end
 
+    def cat_1_assessments
```
shjohnson: Probably some refactoring can happen with this file to be honest.. as there are some bits which look repeated but rather get things working before that happens.

app/presenters/green_lanes/assessments_presenter.rb
```diff
@@ -64,11 +84,19 @@ def extract_exemptions(category_assessments)
       category_assessments.flat_map { |ca| ca.exemptions.map(&:code) }
     end
 
-    def process_exemptions(exemptions)
-      return [] if exemptions.nil?
+    def process_exemptions(answers)
```
shjohnson: chnaged here because its actually answers that are passed in as a hash with the key being the cat_assessment number

app/presenters/green_lanes/assessments_presenter.rb
alexdesi: do we need for `cat1_with_exemptions`
and `cat1_with_exemptions` ? :thinking: 
also, where are they used?

app/presenters/green_lanes/assessments_presenter.rb
alexdesi: `process_exemptions` and `process_assessment` are not meaningful method names:
you cannot deduce what they do :)

app/controllers/green_lanes/check_your_answers_controller.rb
```diff
@@ -11,6 +11,7 @@ def show
       @country_of_origin = check_your_answers_params[:country_of_origin] || GeographicalArea::ERGA_OMNES
       @moving_date = check_your_answers_params[:moving_date]
       @category_one_assessments = determine_category.cat1_with_exemptions
+      @category_two_assessments_without_exemptions = determine_category.cat2_without_exemptions.present?
```
alexdesi: it seems ` @category_two_assessments_without_exemptions ` is an array, but actually it is a true/false value

alexdesi (commented):
added a couple of comments

alexdesi (approved):
:rocket:

PR #2003, opened 2024-08-23
This allow the segments of the date (year, month and day)

alexdesi:
### What?
This allows the segments of the date (year, month and day)
to be passed inside the Form Validation as strings and not being casted to integer (allowing a fine-grained validation)

### Why?
to reckon as invalid, date like this one:  "1! 12 2024"

### DIFF
--- a/app/controllers/green_lanes/applicable_exemptions_controller.rb
+++ b/app/controllers/green_lanes/applicable_exemptions_controller.rb
@@ -102,14 +102,6 @@ module GreenLanes
       params.permit(:commodity_code, :country_of_origin, :moving_date)
     end
 
-    def moving_requirements_params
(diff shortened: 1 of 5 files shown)


app/controllers/green_lanes/applicable_exemptions_controller.rb
```diff
@@ -102,14 +102,6 @@ def goods_nomenclature_params
       params.permit(:commodity_code, :country_of_origin, :moving_date)
     end
 
-    def moving_requirements_params
```
alexdesi, line 105: weird, but this is not used!

shjohnson: Nice spot if not used think this was altered in a PR refactor

shjohnson (approved):
🟢

PR #2004, opened 2024-08-23
Fix back links and remove Pseudo codes from showing

shjohnson:
### Jira link

https://transformuk.atlassian.net/browse/GL-823
https://transformuk.atlassian.net/browse/GL-930

### DIFF
--- a/app/controllers/green_lanes/applicable_exemptions_controller.rb
+++ b/app/controllers/green_lanes/applicable_exemptions_controller.rb
@@ -9,7 +9,7 @@ module GreenLanes
     def new
       @exemptions_form = build_exemptions_form
       @category_one_assessments_without_exemptions = determine_category.cat1_without_exemptions
-      @back_link_path = determine_back_link_path
+      @back_link_path = back_link_path_for_current_page
 
       render_exemptions_questions
     end
@@ -30,24 +30,12 @@ module GreenLanes
     private
 
     def determine_back_link_path
-      permitted_params = params.permit(:commodity_code, :country_of_origin, :moving_date, :c1ex, :c2ex, ans: {})
-
-      if category == 2 && @category_one_assessments_without_exemptions.empty?
-        green_lanes_applicable_exemptions_path(
-          category: 1,
-          commodity_code: permitted_params[:commodity_code],
-          country_of_origin: permitted_params[:country_of_origin],
-          moving_date: permitted_params[:moving_date],
-          ans: parsed_ans(permitted_params[:ans]),
-          c1ex: permitted_params[:c1ex],
-        )
-      else
-        green_lanes_moving_requirements_path(
-          commodity_code: params[:commodity_code],
-          country_of_origin: params[:country_of_origin],
-          moving_date: params[:moving_date],
-        )
-      end
+      BackLinkPath.new(
+        current_page: :new_applicable_exemptions,
+        params:,
+        category_one_assessments_without_exemptions: @category_one_assessments_without_exemptions,
+        category_two_assessments_without_exemptions: determine_category.cat2_without_exemptions,
+      ).call
     end
 
     def parsed_ans(ans_param)
--- a/app/controllers/green_lanes/applicable_exemptions_controller.rb
+++ b/app/controllers/green_lanes/applicable_exemptions_controller.rb
@@ -8,6 +8,7 @@ module GreenLanes
 
     def new
       @exemptions_form = build_exemptions_form
+      @category_one_assessments_without_exemptions = determine_category.cat1_without_exemptions
       @back_link_path = determine_back_link_path
 
       render_exemptions_questions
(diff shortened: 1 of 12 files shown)


app/controllers/green_lanes/applicable_exemptions_controller.rb
```diff
@@ -31,7 +32,7 @@ def create
     def determine_back_link_path
       permitted_params = params.permit(:commodity_code, :country_of_origin, :moving_date, :c1ex, :c2ex, ans: {})
 
-      if category == 2
+      if category == 2 && @category_one_assessments_without_exemptions.empty?
```
alexdesi: maybe we can take time to move `determine_back_link_path` in the helper so we can add a test for it?

app/controllers/green_lanes/check_your_answers_controller.rb
```diff
@@ -23,7 +24,8 @@ def show
 
     def determine_back_link_path(permitted_params)
```
alexdesi: same here, this is becoming complex, we'd need a test for it.

spec/helpers/green_lanes_helper_spec.rb
```diff
@@ -15,6 +15,18 @@
     allow(helper).to receive(:render)
   end
 
+  describe '#hide_pseudo_codes' do
+    it 'returns an empty string for codes starting with WFE', :aggregate_failures do
+      expect(hide_pseudo_codes('WFE001')).to eq('')
+      expect(hide_pseudo_codes('WFE015')).to eq('')
```
alexdesi: I think one test is enough

spec/helpers/green_lanes_helper_spec.rb
```diff
@@ -15,6 +15,18 @@
...
+    end
+
+    it 'returns the original code for codes not starting with WFE', :aggregate_failures do
+      expect(hide_pseudo_codes('Y171')).to eq('Y171')
```
alexdesi: I think one test is enough

app/helpers/green_lanes_helper.rb
```diff
@@ -7,6 +7,14 @@ def exemption_checkbox_none?(category_assessment_id)
     exemption_checkbox_checked?(category_assessment_id, 'none')
   end
 
+  def hide_pseudo_codes(code)
```
alexdesi: should we move this to the exemption model?

alexdesi (dismissed):

app/controllers/green_lanes/applicable_exemptions_controller.rb
```diff
@@ -8,7 +8,9 @@ class ApplicableExemptionsController < ApplicationController
 
     def new
       @exemptions_form = build_exemptions_form
-      @back_link_path = determine_back_link_path
+      @category_one_assessments_without_exemptions = candidate_categories.cat1_without_exemptions
+      @category_two_assessments_without_exemptions = candidate_categories.cat2_without_exemptions
```
alexdesi: are the
@category_one_assessments_without_exemptions 
@category_two_assessments_without_exemptions 
used in the UI (Erb files)
?
maybe we do not need them ( BackLinkPath can get those directly or via args)

app/controllers/green_lanes/applicable_exemptions_controller.rb
```diff
@@ -20,33 +22,29 @@ def create
...
+        :country_of_origin,
+        :moving_date,
+        :c1ex,
+        :c2ex,
+        ans: {},
+      )
+    end
+
+    def back_link_path_for_current_page
+      BackLinkPath.new(params: back_link_params,
```
alexdesi, line 43: It's great you moved it into a service :rocket:

app/services/green_lanes/back_link_path.rb
```diff
@@ -0,0 +1,53 @@
...
+        commodity_code: @params[:commodity_code],
+        country_of_origin: @params[:country_of_origin],
+        moving_date: @params[:moving_date],
+      }
+
+      if category == 2 && @category_two_assessments_without_exemptions.empty?
+        # Remvoing the old answers for the back link path
+        ans.delete('2')
+
+        Rails.application.routes.url_helpers
```
alexdesi: If you include

` include Rails.application.routes.url_helpers`
you can just use the path helpers 

for example just `new_green_lanes_applicable_exemptions_path`

see `app/services/green_lanes/determine_next_page.rb`

app/controllers/green_lanes/applicable_exemptions_controller.rb
shjohnson: yeah i guess so

app/controllers/green_lanes/applicable_exemptions_controller.rb
```diff
@@ -20,33 +20,29 @@ def create
         next_page = determine_next_page
         redirect_to handle_next_page(next_page)
       else
-        @back_link_path = determine_back_link_path
+        @back_link_path = back_link_path_for_current_page
```
alexdesi, line 23: don't we need a back_link when the form is not valid?

shjohnson: We dont because when the page has an invalid form its still using the new back link and ive tested this works for cat 1 questions and cat 2 questions

alexdesi (dismissed):
:sheep: :sheep: :sheep: :sheep: :sheep: :sheep: :sheep:

alexdesi (approved):
:sheep:  :white_check_mark:  :sheep:  :white_check_mark:  :sheep:  :white_check_mark:

PR #2010, opened 2024-08-28
GL-908: Change 'green_lanes' to 'check-spimm-eligibility' in URL path.

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/GL-908

### What?
Change 'green_lanes' to 'check-spimm-eligibility' in URL path.

### Why?
To improve Web usability.

### DIFF
--- a/app/services/green_lanes/determine_next_page.rb
+++ b/app/services/green_lanes/determine_next_page.rb
@@ -1,5 +1,7 @@
 module GreenLanes
   class DetermineNextPage
+    include Rails.application.routes.url_helpers
+
     def initialize(goods_nomenclature)
       @determine_category = GreenLanes::DetermineCategory.new(goods_nomenclature)
       @categories = @determine_category.categories
@@ -63,13 +65,13 @@ module GreenLanes
     end
 
     def new_exemptions_path(category, params = {})
-      query = params.to_query
-
-      "/check-spimm-eligibility/applicable_exemptions/new?category=#{category}" + (query.present? ? "&#{query}" : '')
+      new_green_lanes_applicable_exemptions_path(
+        params.merge(category:),
+      )
     end
 
     def check_your_answers
-      Rails.application.routes.url_helpers.green_lanes_check_your_answers_path
+      green_lanes_check_your_answers_path
     end
   end
 end
--- a/app/services/green_lanes/determine_next_page.rb
+++ b/app/services/green_lanes/determine_next_page.rb
@@ -65,7 +65,7 @@ module GreenLanes
     def new_exemptions_path(category, params = {})
       query = params.to_query
 
-      "/green_lanes/applicable_exemptions/new?category=#{category}" + (query.present? ? "&#{query}" : '')
+      "/check-spimm-eligibility/applicable_exemptions/new?category=#{category}" + (query.present? ? "&#{query}" : '')
     end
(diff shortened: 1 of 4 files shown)


app/services/green_lanes/determine_next_page.rb
```diff
@@ -65,7 +65,7 @@ def question_unanswered?(cat_exemptions_apply)
     def new_exemptions_path(category, params = {})
       query = params.to_query
 
-      "/green_lanes/applicable_exemptions/new?category=#{category}" + (query.present? ? "&#{query}" : '')
+      "/check-spimm-eligibility/applicable_exemptions/new?category=#{category}" + (query.present? ? "&#{query}" : '')
```
willfish: Could we use the path helper?

config/routes.rb
```diff
@@ -108,7 +108,7 @@
   resolve('GreenLanes::CategoryAssessmentSearch') { [:category_assessments] }
 
   scope constraints: ->(_req) { TradeTariffFrontend::ServiceChooser.uk? } do
-    namespace :green_lanes do
+    namespace :green_lanes, path: 'check-spimm-eligibility' do
```
willfish: So 100% of this app uses underscores. Can we just use underscores despite gds guidance?

alexdesi: I've checked, and they still want dashes because they are easier to type and aligned with all the Gov UK websites.

willfish: Let's just stick with being consistent with this application and we can review our paths for the entire application at a later date

willfish: It's not really an option to have a different standard for this one small service

alexdesi: Ok, at the end we've agreed on _ for now

app/services/green_lanes/determine_next_page.rb
alexdesi: sure, I've also used url_helpers inside the DetermineNextPage,
much better now.

spec/requests/green_lanes/applicable_exemptions_controller_spec.rb
```diff
@@ -38,6 +38,6 @@
     end
 
     it { is_expected.to have_http_status :redirect }
-    it { is_expected.to redirect_to('http://www.example.com/green_lanes/check_your_answers?ans%5B1%5D%5B18fcbb5b75781f8a676bd84dae9c170e%5D%5B%5D=none&ans%5B1%5D%5Ba6b633a7b098132ec45c036d0e14713a%5D%5B%5D=Y997&c1ex=false&category=1&commodity_code=4114109000&country_of_origin=UA&moving_date=2024-05-29') }
+    it { is_expected.to redirect_to('http://www.example.com/check_spimm_eligibility/check_your_answers?ans%5B1%5D%5B18fcbb5b75781f8a676bd84dae9c170e%5D%5B%5D=none&ans%5B1%5D%5Ba6b633a7b098132ec45c036d0e14713a%5D%5B%5D=Y997&c1ex=false&category=1&commodity_code=4114109000&country_of_origin=UA&moving_date=2024-05-29') }
```
willfish: I wonder if we want to use path helpers, here, too?

willfish (dismissed):

willfish (approved):

PR #2019, opened 2024-09-05
GL-929: Get the right checkboxes values from param

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/GL-929

### What?
It was not reading the right param to compose the checkboxes.
This PR fixes it.

### DIFF
--- a/app/helpers/green_lanes_helper.rb
+++ b/app/helpers/green_lanes_helper.rb
@@ -1,6 +1,7 @@
 module GreenLanesHelper
   def exemption_checkbox_checked?(resource_id, exemption_code)
-    category_assessments_checked(resource_id)&.include?(exemption_code)
+    params.dig(:exemptions, resource_id.to_s)
+          &.include?(exemption_code)
   end
 
   def exemption_checkbox_none?(resource_id)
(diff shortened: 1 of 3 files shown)


app/helpers/green_lanes_helper.rb
```diff
@@ -1,10 +1,12 @@
 module GreenLanesHelper
   def exemption_checkbox_checked?(resource_id, exemption_code)
-    category_assessments_checked(resource_id)&.include?(exemption_code)
+    params.dig(:exemptions, resource_id.to_s)
+          &.include?(exemption_code)
```
alexdesi, line 4: We should not use the `&`,
it is defensive, to avoid errors in case exemptions is missing,
which happen when clicking "back" link.

app/helpers/green_lanes_helper.rb
```diff
@@ -74,10 +76,6 @@ def prettify_category(category)
 
   private
 
-  def category_assessments_checked(resource_id)
-    params.dig(:exemptions, :category_assessments_checked, resource_id.to_s)
```
alexdesi, line 78: :category_assessments_checked does not exist in params, and neither in the code elsewhere,
maybe legacy code (first version of the form)

willfish (dismissed):

willfish (dismissed):

shjohnson (approved):

PR #2034, opened 2024-09-12
GL-979: This adds the expiry in some urls in the green lanes (SPIMM).

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/GL-979

### Why
If the user bookmarks the page, and it reloads the page after 10 hours, they must restart the journey from the beginning
(/check_spimm_eligibility).

### DIFF
--- a/app/controllers/green_lanes/applicable_exemptions_controller.rb
+++ b/app/controllers/green_lanes/applicable_exemptions_controller.rb
@@ -2,9 +2,13 @@ module GreenLanes
   class ApplicableExemptionsController < ApplicationController
     include GreenLanesHelper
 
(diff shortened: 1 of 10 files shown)


willfish (dismissed):
I don't know why we'd build this. Because the criteria have expired/changed based on data points?

Why 10 hours later?

shjohnson (dismissed):

shjohnson:
@willfish I cant remember why 10 hrs i think it was because slightly longer than a work day but the idea is that if a certain time has passed we should expire the URL so people arent bookmarking outdated data.

willfish (approved):

shjohnson (approved):

PR #2035, opened 2024-09-12
OTT-496 Speed up search lookups

neilmiddleton:
### Jira link

OTT-496

### What?

The search box has a debounce function set to avoid hammering the API  with every key press, however, it's set to quite a slow level which is what is causing the issue in this issue.

This PR speeds up the debounce to 200ms which doesn't solve the issue, but should make it much much harder to trigger.

We _could_ fix this properly and stop the form submitting if the lookup is still running, but doing this would make the form effectively break if the API lookup is a little slow or fails.


HWallenberg (approved):
lgtm 🚀

willfish (approved):

PR #2039, opened 2024-09-13
OTT-495 Correct URL formation for search suggestions

neilmiddleton:
### Jira link

OTT-495

### What?

Corrects URL formation for searchSuggestions URL

### Why?

Previously the code would take any existing URL params and try to append on the search query, and this was incorrect.  For instance:

`/find_commodity?foo=bar` then tried to turn into `/find_commodity?foo=bar?query=chicken`

which is an invalid URL.
This PR uses the JS URL obejct to ensure we're building a proper URL regardless.


HWallenberg (approved):
nice one! 👍

PR #2041, opened 2024-09-14
BAU: Enable nixos

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [x] Added support for nixos

### Why?

I am doing this because:

- This is needed specifically for my dev environment and stops me from needing to use a different OS


spec/spec_helper.rb
```diff
@@ -24,6 +24,7 @@
 Capybara.javascript_driver = :cuprite
 Capybara.register_driver(:cuprite) do |app|
   opts = {
+    browser_name: :chrome,
```
willfish, line 27: Just making this default explicit. No impact of change

PR #2048, opened 2024-09-18
GL-1040: Fixes sequence handling of moving date form params

willfish:
### Jira link

GL-1040

### What?

I have added/removed/altered:

- [x] Fixes parsing of the moving date params

### Why?

I am doing this because:

- Before we were fetching the goods nomenclature with unparsed moving date params. The net effect of this is we were determining the next page with today's version of the goods nomenclature. This is distinct from determinations that come in subsequent controllers which were using the correct date and thus showing no data. The fix was to adjust the params handling and to just use params for the moving date that had been parsed by the form object using ActiveRecord::AttributeAssignment.


alexdesi (approved):
Bellissimo! :tada:

PR #2049, opened 2024-09-18
OTT-179 Fix form date validation

neilmiddleton:
### Jira link

[OTT-179](https://transformuk.atlassian.net/browse/OTT-179)

### What?

This PR changes how the search forms work for commodities and quota search in order to allow us to return the fact that a date is erroneous rather than just return an error page.

https://github.com/user-attachments/assets/17c65ae6-08ab-4905-badb-a61e7b65e654

Note, the test suite is green, but there are some fairly deep changes here.  This whole part of the app at some point needs a re-working as there's a lot of things here that are very brittle.  There are few tests removed that are no longer appropriate


HWallenberg (dismissed):
It gets the 👍 from me.  I can understand what you've done and why.

willfish (dismissed):

HWallenberg (approved):

PR #2072, opened 2024-09-30
OTT-516 Replace method of fix for general rules of intepretation

neilmiddleton:
### Jira link

https://transformuk.atlassian.net/browse/OTT-516

### What?

Reverted previous change to General Interpretative Rules and isolated that fix to just the General Interpretative Rules wording.  This stops the leaking of formatting across to Notes numbering


HWallenberg (approved):
🚀 🙌

PR #2073, opened 2024-09-30
GL-1068: …

alexdesi:
### Jira link
https://transformuk.atlassian.net/browse/GL-1068
https://transformuk.atlassian.net/browse/GL-1080

### What?

- Add new validation for the ambiguous exemption when there are more than one exemptions that are equal.
- Move validation errors on top of the header of the page.


app/forms/green_lanes/applicable_exemptions_form.rb
```diff
@@ -46,5 +54,28 @@ def presented_answers
...
+
+    def exemptions_count
+      all_exemptions = category_assessments.flat_map do |ca|
+        ca.exemptions.map(&:resource_id)
+      end
+
+      all_exemptions.group_by { |code| code }
+                    .transform_values(&:count)
+    end
   end
```
alexdesi, line 80: This is the core of this PR.
What it does is to count the checked answers,
and count the exemptions available,
then they are compared.

If they do not match, a validation error will be shown to the user.

neilmiddleton (dismissed):

rasikasri (approved):

PR #2080, opened 2024-10-02
OTT-526 Don't double encode special chars

neilmiddleton:
### Jira link

OTT-526

### What?

This removes double-encoding of params in the search which allows us to search for keys such as `1,4-dinitrobenzene`

Previously, `1,4` was being encoded to `1%252C4` which equates to `1%,4` due to the double encode.  Removing this means we end up with what we should have.


HWallenberg (approved):
nice one thank you! 😄

PR #2105, opened 2024-11-04
HMRC-304 | Refactor url_options 

EraliaKats:
### Jira link

[HMRC-304](https://transformuk.atlassian.net/browse/HMRC-304)
See comment for screen recording of behaviour after changes [here](https://transformuk.atlassian.net/browse/HMRC-304?focusedCommentId=27367)

### What?

I have added/removed/altered:

- [x] Refactored `url_options` to be defined in `application_controller` again

### Why?

I am doing this because:

- Users need the ability to filter by both country and date as desired
- We should keep all updates of [OTT-179](https://transformuk.atlassian.net/browse/OTT-179) without the bug introduced

### Have you? (optional)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -772,6 +772,7 @@ GEM
 
 PLATFORMS
   arm64-darwin-23
+  arm64-darwin-24
   x86_64-darwin-22
   x86_64-linux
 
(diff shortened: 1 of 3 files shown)


HWallenberg (approved):
LGTM! 🙌

PR #2108, opened 2024-11-11
HMRC-307 Fix issue with fast typing in search

neilmiddleton:
### Jira link

HMRC-307

### What?

I have removed a line that is trying to overwrite search terms when the suggestions API comes back.

### Why?

I am doing this because the delay on the suggestions API meant it was replacing what people had typed into the box on a submit if made before the suggestions API had returned.


HWallenberg (approved):
🙌

PR #2165, opened 2025-01-22
HMRC-584 Resolve search issues

neilmiddleton:
### Jira link

HMRC-584

### What?

Fixed the issues with the autocomplete and the search button.


app/views/shared/search/_search_form.html.erb
```diff
@@ -79,4 +73,4 @@
   </div>
   <% end %>
 </div>
-<% end %>
+<% end %>
```
EraliaKats: Could you pls add a line here :)

EraliaKats (approved):

PR #2168, opened 2025-01-23
HMRC-271 Disable V1 API routing

neilmiddleton:
### Jira link

HMRC-271

### What?

Disable routing for the v1 API on a temporary basis.  This PR is intended to be used for brown-out, and can be revert'ed or superseded by another PR to make the disabling permanent.


neilmiddleton:
Closing as this is not appropriate anymore with API routing changes

PR #2183, opened 2025-02-07
BAU: Bump deps

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [x] Bumps deps
- [x] Fixes rubocop violations
- [ ] Fixes test failures

### Why?

I am doing this because:

- To keep us up to date


willfish:
This defeated me. I'm going to have to break this out into specific pieces

PR #2188, opened 2025-02-11
HMRC-657: Migrate to new routes

willfish:
### Jira link

HMRC-657

### What?

I noticed that refreshed cassettes for the ./spec/features/green_lanes_category_assessments_spec.rb specs break some scenarios so we'll need to find up-to-date examples of these. I've skipped them for now. But it makes sense to speak with a BA about what good data for these scenarios is.

I have added/removed/altered:

- [x] Remove handling of api requests
- [x] Reset all VCR cassettes
- [x] Reset all tests to match new VCR cassettes

### Why?

I am doing this because:

- This is cleanup that's coming out of the separation of the frontend and the backend
- We should really be removing the frontend routes that support .json but this needs socialisation/review with the wider team


willfish:
Okay, I just verified from both the application and cloudfront logs that there's not a single user of the json endpoints in production that aren't on the `/api` paths so we can just kill the request forwarder completely and remove any reference to these paths.

```
fields `cs-uri-stem` as path, `cs-method` as method, `sc-content-type` as content, @logStream
| filter @logStream = "CloudFront_E385MK7DG1LKK"
| filter path not like /^\/api/
| filter path not like /^\/xi\/api/
| filter path not like /search_suggestions/
| filter content like /json/
| stats count(*) by path
| sort `count(*)` desc
```

neilmiddleton (dismissed):

HWallenberg (approved):

PR #2194, opened 2025-02-18
HMRC-669 | Remove CHIEF guidance from export conditions and certificate search

EraliaKats:
### Jira link

[HMRC-669](https://transformuk.atlassian.net/browse/HMRC-669)

### What?

I have added/removed/altered:

- [x] Removed CHIEF guidance from export conditions and certificate search

### Why?

I am doing this because:

- Service request from Scott, CHIEF guidance isn't used anymore


willfish (approved):
LGTM

PR #2228, opened 2025-03-25
HMRC-823: Enable maintenance mode

willfish:
### Jira link

[HMRC-823](https://transformuk.atlassian.net/browse/HMRC-823)

### What?

I have a...

- Enabled maintenance mode in all environments

### Why?

I am doing this because...

- We're going to have 2 minutes of outage


HWallenberg (approved):
lgtm but do you need more senior approval too?

willfish:
Going to kill this as it actually takes longer to enable maintenance mode than it does do a full restore

PR #2229, opened 2025-03-27
HMRC-818 Fix country autocomplete

vertism:
### Jira link

[HMRC-818](https://transformuk.atlassian.net/browse/HMRC-818)

### What?

Added missing module reference which was stopping the autocomplete functionality working and causing a console error.

### Why?

https://www.trade-tariff.service.gov.uk/quota_search country dropdown is meant to have autocomplete enabled.


HWallenberg (approved):
lgtm! 🙌

PR #2230, opened 2025-03-27
HMRC-913 Disable analytics unless consent cookie is present

vertism:
### Jira link

[HMRC-913](https://transformuk.atlassian.net/browse/HMRC-913)

### What?

- Removes gtm-loader script as it was duplicating the script tag following #2213 change
- Adds helper that returns boolean as to whether analytics are allowed based on the consent cookie value

### Why?

If the user has not consented to analytics cookies, the GTM script should not be loaded.
This adds some logic duplication between backend and frontend, but it is necessary in the case where javascript is disabled in the client.


neilmiddleton (approved):

EraliaKats (approved):

HWallenberg (approved):
🚀

PR #2248, opened 2025-04-08
HMRC-962: Simplify and update gems/js deps

willfish:
### Jira link

[HMRC-962](https://transformuk.atlassian.net/browse/HMRC-962)

### What?

I have a...

- Removed some unused/barely used gems
- Removed some unused forwarding behaviour (since we separated backend/frontend)
- Bumped development/test gems

### Why?

I am doing this because...

- This is just passing simplification


neilmiddleton:
Might need a rebase or something

HWallenberg (approved):

PR #2295, opened 2025-05-12
HMRC-1025 MyOTT Chapter Selection and Check Your Answers pages

HWallenberg:
### Jira link

[HMRC-1025](https://transformuk.atlassian.net/browse/HMRC-1025)
[HMRC-1028](https://transformuk.atlassian.net/browse/HMRC-1028)

### What?

I have added MyOTT Chapter Selection and Check Your Answers pages

### Why?

I am doing this because it's a requirement for MyOTT project

image
image

### DIFF
--- a/app/controllers/myott/subscriptions_controller.rb
+++ b/app/controllers/myott/subscriptions_controller.rb
@@ -2,10 +2,9 @@ module Myott
   class SubscriptionsController < ApplicationController
     before_action :disable_search_form,
                   :disable_switch_service_banner,
-                  :disable_last_updated_footnote,
-                  :cache_sections_chapters
+                  :disable_last_updated_footnote
 
-    before_action :clear_stale_flash_error, only: %i[dashboard check_your_answers]
+    before_action :sections_chapters, only: %i[chapter_selection]
 
     def dashboard
       @email = 'test@email.com'
@@ -13,16 +12,15 @@ module Myott
     end
 
     def chapter_selection
-      @selected_chapter_ids = Array(session[:chapter_ids])
-      all_chapters = @sections_chapters.values.flatten
-      @selected_chapters = all_chapters.select { |chapter| @selected_chapter_ids.include?(chapter.to_param) }
+      selected_chapter_ids = Array(session[:chapter_ids])
+      @selected_chapters = all_chapters.select { |chapter| selected_chapter_ids.include?(chapter.to_param) }
     end
 
     def check_your_answers
       selected_ids = Array(params[:chapter_ids])
 
       if selected_ids.empty?
-        flash[:error] = 'Select the chapters you want tariff updates about.'
+        flash.now[:error] = 'Select the chapters you want tariff updates about.'
         @selected_chapters = []
         render :chapter_selection
         return
@@ -30,28 +28,22 @@ module Myott
 
       session[:chapter_ids] = params[:chapter_ids]
 
-      all_chapters = @sections_chapters.values.flatten
       @selected_chapters = all_chapters.select { |chapter| selected_ids.include?(chapter.to_param) }
-    end
-
-    def remove_chapter_selection
-      session[:chapter_ids] = Array(session[:chapter_ids]) - [params[:chapter_id]]
-      head :ok
     end
 
     private
 
-    def cache_sections_chapters
-      @sections_chapters = Rails.cache.fetch('sections_chapters', expires_in: 1.hour) do
+    def all_chapters
+      sections_chapters.values.flatten
+    end
+
+    def sections_chapters
+      @sections_chapters ||= Rails.cache.fetch('sections_chapters', expires_in: 1.day) do
         Section.all.each_with_object({}) do |section, hash|
           chapters = Section.find(section.resource_id).chapters
           hash[section] = chapters
         end
       end
-    end
-
-    def clear_stale_flash_error
-      flash[:error] = nil
     end
   end
 end
--- a/app/controllers/myott/subscriptions_controller.rb
+++ b/app/controllers/myott/subscriptions_controller.rb
@@ -2,10 +2,53 @@ module Myott
   class SubscriptionsController < ApplicationController
     before_action :disable_search_form,
                   :disable_switch_service_banner,
-                  :disable_last_updated_footnote
+                  :disable_last_updated_footnote,
+                  :cache_sections_chapters
 
     def dashboard
       @email = 'test@email.com'
+      session[:chapter_ids] = nil
+      flash[:error] = nil
+    end
+
+    def chapter_selection
+      @selected_chapter_ids = Array(session[:chapter_ids])
+      all_chapters = @sections_chapters.values.flatten
+      @selected_chapters = all_chapters.select { |chapter| @selected_chapter_ids.include?(chapter.to_param) }
+    end
+
+    def check_your_answers
+      selected_ids = Array(params[:chapter_ids])
+
+      if selected_ids.empty?
+        flash[:error] = 'Select the chapters you want tariff updates about.'
+        @selected_chapters = []
+        render :chapter_selection
+        return
+      else
+        flash[:error] = nil
+      end
+
+      session[:chapter_ids] = params[:chapter_ids]
+
+      all_chapters = @sections_chapters.values.flatten
+      @selected_chapters = all_chapters.select { |chapter| selected_ids.include?(chapter.to_param) }
+    end
+
+    def remove_chapter_selection
+      session[:chapter_ids] = Array(session[:chapter_ids]) - [params[:chapter_id]]
+      head :ok
+    end
+
+    private
+
+    def cache_sections_chapters
+      @sections_chapters = Rails.cache.fetch('sections_chapters', expires_in: 1.hour) do
+        Section.all.each_with_object({}) do |section, hash|
+          chapters = Section.find(section.resource_id).chapters
+          hash[section] = chapters
+        end
+      end
     end
   end
 end
--- a/app/controllers/myott/subscriptions_controller.rb
+++ b/app/controllers/myott/subscriptions_controller.rb
@@ -5,10 +5,11 @@ module Myott
                   :disable_last_updated_footnote,
                   :cache_sections_chapters
 
+    before_action :clear_stale_flash_error, only: %i[dashboard check_your_answers]
+
     def dashboard
       @email = 'test@email.com'
       session[:chapter_ids] = nil
-      flash[:error] = nil
     end
 
     def chapter_selection
@@ -25,8 +26,6 @@ module Myott
         @selected_chapters = []
         render :chapter_selection
         return
-      else
-        flash[:error] = nil
       end
 
       session[:chapter_ids] = params[:chapter_ids]
@@ -49,6 +48,10 @@ module Myott
           hash[section] = chapters
         end
       end
+    end
+
+    def clear_stale_flash_error
+      flash[:error] = nil
     end
   end
 end
--- a/app/views/myott/subscriptions/_accordion_section_content.html.erb
+++ b/app/views/myott/subscriptions/_accordion_section_content.html.erb
@@ -1,0 +1,32 @@
+<div id="section-chapter-accordion-sections-content-<%=section.resource_id%>" class="govuk-accordion__section-content">
+    <table class="govuk-table">
+    <tbody class="govuk-table__body">
+        <% chapters.each do |chapter| %>
+        <tr class="govuk-table__row">
+            <td class="govuk-table__cell govuk-!-padding-top-0 govuk-!-padding-right-1 set_checkbox_width no_border">
+            <div class="govuk-checkboxes govuk-checkboxes--small" data-module="govuk-checkboxes">
+                <div class="govuk-checkboxes__item">
+                <input class="govuk-checkboxes__input"
+                    id="chapter_<%= chapter.to_param %>"
+                    name="chapter_ids[]"
+                    type="checkbox"
+                    data-chapter-selection-target="chapterCheckbox"
+                    value="<%= chapter.to_param %>"
+                    <%= "checked" if selected_chapters.any? { |c| c.to_param == chapter.to_param } %>>
+                <label class="govuk-checkboxes__label" for="chapter_<%= chapter.to_param %>"></label>
+                </div>
+            </div>
+            </td>
+            <td class="govuk-table__cell section-id-col govuk-!-padding-top-2 no_border">
+            <div class="chapter-code">
+                <span class="code-text"><%= chapter.short_code %></span>
+            </div>
+            </td>
+            <td class="govuk-table__cell govuk-!-padding-top-2 no_border">
+            <p><%= chapter.to_s %></p>
+            </td>
+        </tr>
+        <% end %>
+    </tbody>
+    </table>
+</div>
--- a/app/views/myott/subscriptions/chapter_selection.html.erb
+++ b/app/views/myott/subscriptions/chapter_selection.html.erb
@@ -1,94 +1,27 @@
 <% content_for :back_link do %>
   <% back_link myott_path %>
 <% end %>
-
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-  <% if flash.now[:error] %>
-    <div class="govuk-error-summary" data-module="govuk-error-summary">
-      <div role="alert">
-        <h2 class="govuk-error-summary__title">
-          There is a problem
-        </h2>
-        <div class="govuk-error-summary__body">
-          <ul class="govuk-list govuk-error-summary__list">
-            <li>
-              <a href="#"><%= flash.now[:error] %></a>
-            </li>
-          </ul>
-        </div>
-      </div>
-    </div>
-  <% end %>
+  <%= render 'error' if flash.now[:error] %>
     <h1 class="govuk-heading-l">Select the chapters you want tariff updates about</h1>
     <div data-controller="chapter-selection">
-      <div class="govuk-body">
+      <div class="govuk-body myott-subscriptions">
         <p>If you only want to receive updates on specific areas, you can select the relevant chapters below.  These can be updated at any time.</p>
         <p>If you don't select any chapters below, you will continue to receive all tariff updates.</p>
-        <div class="govuk-grid-row govuk-!-padding-bottom-3">
-          <div class="govuk-grid-column-two-thirds">
-            <div class="govuk-grid-row">
-              <div class="govuk-grid-column-one-half">
-                <button class="link_button" data-action="click->chapter-selection#toggleAllChapterCheckboxes" value="true" >Select all chapters</button>
-              </div>
-              <div class="govuk-grid-column-one-half">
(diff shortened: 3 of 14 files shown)


app/controllers/myott/subscriptions_controller.rb
```diff
@@ -2,10 +2,53 @@ module Myott
...
     before_action :disable_search_form,
                   :disable_switch_service_banner,
-                  :disable_last_updated_footnote
+                  :disable_last_updated_footnote,
+                  :cache_sections_chapters
 
     def dashboard
       @email = 'test@email.com'
+      session[:chapter_ids] = nil
+      flash[:error] = nil
```
EraliaKats: Why do you need to set this to nil? I don't think it should be expecting flash errors anywhere? I reckon you've done this because in your view you're checking if it exists, but if it's not set it'd return nil anyway no?

app/controllers/myott/subscriptions_controller.rb
```diff
@@ -2,10 +2,53 @@ module Myott
...
+    def check_your_answers
+      selected_ids = Array(params[:chapter_ids])
+
+      if selected_ids.empty?
+        flash[:error] = 'Select the chapters you want tariff updates about.'
+        @selected_chapters = []
+        render :chapter_selection
+        return
+      else
+        flash[:error] = nil
```
EraliaKats: same here

app/controllers/myott/subscriptions_controller.rb
```diff
@@ -2,10 +2,53 @@ module Myott
...
+      end
+
+      session[:chapter_ids] = params[:chapter_ids]
+
+      all_chapters = @sections_chapters.values.flatten
+      @selected_chapters = all_chapters.select { |chapter| selected_ids.include?(chapter.to_param) }
+    end
+
+    def remove_chapter_selection
+      session[:chapter_ids] = Array(session[:chapter_ids]) - [params[:chapter_id]]
```
EraliaKats: This syntax feels wrong/unreadable. Is there any way you could remove the params from the session without altering the session itself? Setting the params to empty or nil ?

app/webpacker/controllers/chapter_selection_controller.js
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  removeChapter(event) {
+    event.preventDefault()
+    const chapterId = event.target.dataset.chapterId
+    fetch("/subscriptions/remove_chapter_selection", {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+        'X-CSRF-Token': document.querySelector("meta[name='csrf-token']").content
```
EraliaKats: probs not relevant for this PR but why CSRF and not whitelisting only our service in origin headers?

EraliaKats (commented):
Thanks for this Hadleigh!

p.s. I ducking hell I hate javascript so much, why so many lines are needed to handle a response

app/controllers/myott/subscriptions_controller.rb
```diff
@@ -2,10 +2,53 @@ module Myott
...
+      head :ok
+    end
+
+    private
+
+    def cache_sections_chapters
+      @sections_chapters = Rails.cache.fetch('sections_chapters', expires_in: 1.hour) do
```
vertism: Probably worth memoizing this as it's called more than once.
```suggestion
      @sections_chapters ||= Rails.cache.fetch('sections_chapters', expires_in: 1.hour) do
```

app/controllers/myott/subscriptions_controller.rb
```diff
@@ -2,10 +2,53 @@ module Myott
...
+                  :cache_sections_chapters
 
     def dashboard
       @email = 'test@email.com'
+      session[:chapter_ids] = nil
+      flash[:error] = nil
+    end
+
+    def chapter_selection
+      @selected_chapter_ids = Array(session[:chapter_ids])
```
vertism: Does this need to be an instance variable?

app/views/myott/subscriptions/chapter_selection.html.erb
vertism, line 1: This is quite long, what do you think about breaking it up by adding some partials? Maybe where there are loops for the accordions and the chapter checkboxs?

EraliaKats: Plus, in dev hub we used this framework made from govuk developers - maybe it'd be helpful here too
https://govuk-components.netlify.app/components/accordion/

app/webpacker/controllers/chapter_selection_controller.js
```diff
@@ -0,0 +1,44 @@
+import {Controller} from '@hotwired/stimulus';
+
+export default class extends Controller {
+  connect() {
+    // accordion doesn't allow sections to be open by default
```
vertism: You can use `govuk-accordion__section--expanded` style to have the accordion open initially

app/controllers/myott/subscriptions_controller.rb
HWallenberg: Yeah, it’s a bit of a safeguard. I was seeing a stale flash[:error] sticking around even after returning to the start of the journey. Resetting it manually here avoids showing outdated errors if the user decides to change their selection. It’s a belt-and-braces thing, but it resolved an issue I hit during testing.

app/controllers/myott/subscriptions_controller.rb
HWallenberg: When running the code locally visiting different screens the flash error persisted when the currently hidden feature 'remove chapter' was active or if the user decided to edit. Setting it to nil if the user decided to go back resolved this error.

I've refactored this to a `before_action` though so it isn't repeated, thanks.

app/controllers/myott/subscriptions_controller.rb
HWallenberg: We do need to alter the session here — it’s acting as the store for the user’s chapter selections as they move through the flow. The params just pass the data between pages, but the session is what keeps their preferences around. So clearing the param wouldn’t help unless we also update the session to match.

app/webpacker/controllers/chapter_selection_controller.js
HWallenberg: As far as I understand it..Rails expects the CSRF token even for same-origin POSTs as part of its default forgery protection. It’s mainly to make sure the request is coming from the actual user/session and not something dodgy like a form or script on another site.

app/controllers/myott/subscriptions_controller.rb
HWallenberg: 🙏

app/controllers/myott/subscriptions_controller.rb
HWallenberg: 🙏

app/controllers/myott/subscriptions_controller.rb
HWallenberg: no! 🙂

app/views/myott/subscriptions/chapter_selection.html.erb
HWallenberg: Thanks @EraliaKats is this the same thing as this is what we've been using https://design-system.service.gov.uk/components/accordion/ 

@vertism I'll look at breaking it up as I agree.

app/webpacker/controllers/chapter_selection_controller.js
HWallenberg: Yes I saw that but it didn't work on my accordion so I copied the example (plain html) adding it to:

` <div class="govuk-accordion__section govuk-accordion__section--expanded"`

and trying 

` <div class="govuk-accordion__section--expanded"`

on it's own but still wouldn't work it broke the accordion buttons functionality.  

I'll give it another go after breaking it up.  

Have you seen this work as intended?  

Thanks.

app/views/myott/subscriptions/chapter_selection.html.erb
HWallenberg: https://github.com/trade-tariff/trade-tariff-frontend/pull/2295/commits/5ee850f8ef5c7011c69d57580760f77d303ec547

app/controllers/myott/subscriptions_controller.rb
EraliaKats: Thanks for the refactoring!

app/controllers/myott/subscriptions_controller.rb
EraliaKats: Hmm, again though this feels a little hacky. We're setting what we need from the session by removing the params from the session. Why do we need to pass these params in the session and not have them be params we can remove easily?

app/webpacker/controllers/chapter_selection_controller.js
EraliaKats: First I'm hearing of it 😅

app/controllers/myott/subscriptions_controller.rb
EraliaKats: Something like this? session.delete(:chapter_ids) ?

HWallenberg: I don't think we're removing the params from the session if I understand what you're asking correctly?
 
We have an array of chapter_ids in the session that the user has selected.  This is used to check any relevant checkboxes in the chapter_selection page.

If the user wants to remove a chapter we do an AJAX postback to the server to remove it from their selection (ie. remove the chapter from the array in the session) in case they want to go back to the chapter_selection page.  

So when clicking 'remove chapter' two things happen:

1. The AJAX call posts a chapterId as a param back to the server and the server removes the chapter_id from the session array of chapter_Ids
2. The row is removed from the current screen.

When the user finally finishes the journey the chapter_ids in the session array will be committed to the backend (not yet written) the session array will then be destroyed.

app/webpacker/controllers/chapter_selection_controller.js
HWallenberg: `govuk-accordion__section--expanded `does work as discovered the status is stored in session storage which I wasn't clearing out!

Accordion code updated and js removed thanks @vertism 🙌

app/views/myott/subscriptions/chapter_selection.html.erb
```diff
@@ -0,0 +1,29 @@
+<% content_for :back_link do %>
+  <% back_link myott_path %>
+<% end %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+  <% if flash.now[:error] %>
+    <%= render partial: 'myott/subscriptions/partials/error' %>
```
vertism: If you put the partials in the same directory then you can just use:
```
<%= render 'error' %>
```

This applies to all the partials

app/controllers/myott/subscriptions_controller.rb
```diff
@@ -2,10 +2,56 @@ module Myott
...
+      selected_chapter_ids = Array(session[:chapter_ids])
+      all_chapters = @sections_chapters.values.flatten
+      @selected_chapters = all_chapters.select { |chapter| selected_chapter_ids.include?(chapter.to_param) }
+    end
+
+    def check_your_answers
+      selected_ids = Array(params[:chapter_ids])
+
+      if selected_ids.empty?
+        flash[:error] = 'Select the chapters you want tariff updates about.'
```
vertism: Maybe try this so you don't need to worry about clearing the flash?
```suggestion
        flash.now[:error] = 'Select the chapters you want tariff updates about.'
```

app/views/myott/subscriptions/chapter_selection.html.erb
HWallenberg: moved partials into same directory and updated paths,thanks.

app/controllers/myott/subscriptions_controller.rb
HWallenberg: Yep seems to work! Thanks 🙂

app/controllers/myott/subscriptions_controller.rb
HWallenberg: Removed this as `flash.now[:error]` negates the need for setting the flash error to nil

app/controllers/myott/subscriptions_controller.rb
HWallenberg: Removed this as `flash.now[:error]` negates the need for setting the flash error to nil

app/controllers/myott/subscriptions_controller.rb
HWallenberg: Removed delete chapter functionality so AJAX functionality not required

app/webpacker/controllers/chapter_selection_controller.js
HWallenberg: Removed delete chapter functionality so AJAX not required

app/controllers/myott/subscriptions_controller.rb
```diff
@@ -2,10 +2,45 @@ module Myott
   class SubscriptionsController < ApplicationController
     before_action :disable_search_form,
                   :disable_switch_service_banner,
-                  :disable_last_updated_footnote
+                  :disable_last_updated_footnote,
+                  :sections_chapters
```
vertism, line 6: I think you could avoid needing a before_action by calling the method instead of the instance variable

vertism: I think the before_action is only needed for the chapter_selection action.

You could use a method call below instead of using the instance variable, or even create an `all_chapters` method.

app/views/myott/subscriptions/chapter_selection.html.erb
```diff
@@ -0,0 +1,29 @@
+<% content_for :back_link do %>
+  <% back_link myott_path %>
+<% end %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+  <% if flash.now[:error] %>
+    <%= render partial: 'error' %>
+  <% end %>
```
vertism: You can combine these:
```suggestion
  <%= render 'error' if flash.now[:error] %>
```

app/views/myott/subscriptions/chapter_selection.html.erb
```diff
@@ -0,0 +1,29 @@
...
+      <div class="govuk-body">
+        <p>If you only want to receive updates on specific areas, you can select the relevant chapters below.  These can be updated at any time.</p>
+        <p>If you don't select any chapters below, you will continue to receive all tariff updates.</p>
+        <%= render partial: 'checkbox_toggles' %>
+        <%= form_with url: myott_check_your_answers_path, method: :post, local: true do %>
+          <div class="govuk-accordion" data-module="govuk-accordion" id="section-chapter-accordion">
+            <% @sections_chapters.each do |section, chapters| %>
+              <div class="govuk-accordion__section <%= section.resource_id == "1" ? 'govuk-accordion__section--expanded' : ''%>" id="accordion-section-<%=section.resource_id%>">
+                <%= render partial: 'accordion_header', locals: {section: section} %>
+                <%= render partial: 'accordion_section_content', locals: {section: section, chapters: chapters} %>
```
vertism: You can make these a little more succinct
```suggestion
                <%= render 'accordion_header', section: %>
                <%= render 'accordion_section_content', section:, chapters: %>
```

app/views/myott/subscriptions/_accordion_section_content.html.erb
```diff
@@ -0,0 +1,32 @@
...
+            <td class="govuk-table__cell govuk-!-padding-top-0 govuk-!-padding-right-1 set_checkbox_width no_border">
+            <div class="govuk-checkboxes govuk-checkboxes--small" data-module="govuk-checkboxes">
+                <div class="govuk-checkboxes__item">
+                <input class="govuk-checkboxes__input"
+                    id="chapter_<%= chapter.to_param %>"
+                    name="chapter_ids[]"
+                    type="checkbox"
+                    data-chapter-selection-target="chapterCheckbox"
+                    value="<%= chapter.to_param %>"
+                    <%= "checked" if @selected_chapters.any? { |c| c.to_param == chapter.to_param } %>>
```
vertism: It would be preferable to pass @selected_chapters as a variable

app/controllers/myott/subscriptions_controller.rb
HWallenberg: Agree, it was there as I considered the data would be loaded when a user 'logs in' however they may not want to change their preferences.  It is required for the CYA page in case there is no selection and the chapter_selection page is rendered.  When I ran it locally the collection didn't exist without it.

app/controllers/myott/subscriptions_controller.rb
```diff
@@ -4,8 +4,46 @@ class SubscriptionsController < ApplicationController
...
+    end
+
+    private
+
+    def all_chapters
+      sections_chapters.values.flatten
+    end
+
+    def sections_chapters
+      @sections_chapters ||= Rails.cache.fetch('sections_chapters', expires_in: 1.hour) do
```
vertism: Could the expiry be extended to a day?

HWallenberg: yes!

app/controllers/myott/subscriptions_controller.rb
```diff
@@ -4,8 +4,46 @@ class SubscriptionsController < ApplicationController
                   :disable_switch_service_banner,
                   :disable_last_updated_footnote
 
+    before_action :sections_chapters, only: %i[chapter_selection check_your_answers]
```
vertism: I don't think this is required for `check_your_answers`

config/routes.rb
```diff
@@ -62,6 +62,9 @@
   if TradeTariffFrontend.myott?
     namespace :myott, path: 'subscriptions' do
       get '/', to: 'subscriptions#dashboard'
+      get '/chapter_selection', to: 'subscriptions#chapter_selection'
+      post '/check_your_answers', to: 'subscriptions#check_your_answers'
+      post '/remove_chapter_selection', to: 'subscriptions#remove_chapter_selection'
```
vertism: Needs removing

app/controllers/myott/subscriptions_controller.rb
HWallenberg: no you're right thanks

app/webpacker/controllers/chapter_selection_controller.js
```diff
@@ -0,0 +1,34 @@
...
+  static targets = ['chapterCheckbox'];
+
+  toggleAllChapterCheckboxes(event) {
+    let checkboxes = this.chapterCheckboxTargets;
+    checkboxes.forEach(checkbox => {
+      checkbox.checked = event.currentTarget.value === 'true';
+    });
+  }
+
+  removeChapter(event) {
```
vertism: Are you going to remove this for now?

app/webpacker/src/stylesheets/_chapter_selection.scss
vertism, line 1: Should these classes be restricted so they can only apply to these new pages?

app/webpacker/controllers/chapter_selection_controller.js
HWallenberg: yes sorry should've done this

app/webpacker/src/stylesheets/_chapter_selection.scss
HWallenberg: yes good idea I've moved them into a `.myott-subscriptions` class 👍

vertism (approved):

PR #2311, opened 2025-05-15
HMRC-1048: Preview Environments

AigulDj:
### Jira link

[HMRC-1048](https://transformuk.atlassian.net/browse/HMRC-1048)

### What?

This PR integrates pull request previews using [Preevy](https://preevy.dev/) and Livecycle. 
Specifically, it includes:

- A new GitHub Actions workflow for Preevy.
- Docker Compose configuration to support previews.
- Updates to development.rb to allow all preview subdomains from livecycle.run.

### Why?

I am doing this because
This enables automatic preview environments for each pull request, making it easier to review and test changes in isolation.


.github/workflows/ci.yml
```diff
@@ -1,12 +1,13 @@
...
-  pull_request:
-    branches-ignore:
-      - main
+  ## Uncomment when testing is done
+  # push:
+  #   branches-ignore:
+  #     - main
+  # pull_request:
+  #   branches-ignore:
+  #     - main
```
EraliaKats: is testing done? 👀

.github/workflows/preevy-down.yml
```diff
@@ -0,0 +1,59 @@
...
+      #     app-id: ${{ vars.GH_APP_ID }}
+      #     private-key: ${{ secrets.GH_APP_PRIVATE_KEY }}
+
+      # - name: 🗑 Delete deployment environment
+      #   uses: strumwolf/delete-deployment-environment@v2.2.3
+      #   with:
+      #     # Use a JWT created with your GitHub App's private key
+      #     token: ${{ steps.get-token.outputs.token }}
+      #     environment: pr-${{ github.event.number }}
+      #     ref: ${{ github.ref_name }}
```
EraliaKats: what are these comments giving us? feeling it's too much noise?

compose.yml
```diff
@@ -0,0 +1,29 @@
...
+    build:
+      context: .
+      dockerfile: Dockerfile
+    ports:
+      - "8080:8080"
+    environment:
+      - GOVUK_APP_DOMAIN="localhost"
+      - GOVUK_WEBSITE_ROOT=http://localhost/
+      - RAILS_ENV=production
+      - PREVIEW_MODE=true # Set to true to enable preview mode and prevent redirecting to the https://www.gov.uk/trade-tariff
```
EraliaKats: this comment out of context doesn't make too much sense. I knew you mean you're preventing a redirect we expect in prod env

lib/trade_tariff_frontend.rb
```diff
@@ -6,7 +6,7 @@ module TradeTariffFrontend
   module_function
 
   def production?
-    environment == 'production'
+    environment == 'production' && ENV['PREVIEW_MODE'] != 'true'
```
EraliaKats: since this is a boolean value, I would prefer to state it as false instead of not true. What do you think?

.github/workflows/ci.yml
AigulDj: I would say yes at a certain stage,
will uncomment that

.github/workflows/preevy-down.yml
AigulDj: Yeah, it is just something I might investigate and apply as one of the next step.
Can remove that for now

.github/workflows/preevy-up.yml
```diff
@@ -0,0 +1,43 @@
+name: Deploy Preevy environment
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
```
EraliaKats, line 4: what's the synchronize type here? Can you give me an example?

.github/workflows/preevy-up.yml
```diff
@@ -0,0 +1,43 @@
...
+  ENVIRONMENT: development
+
+# allow a single job to run per PR
+concurrency: preevy-${{ github.event.number }}
+
+jobs:
+  deploy:
+    environment:
+          name: pr-${{ github.event.number }}
+          # url: ${{ fromJson(steps.preevy_up.outputs.urls-map).frontend[3000] }}
```
EraliaKats: do we need this url comment?

EraliaKats (commented):
This is great work Aigul! Let me know what you think about the comments

.github/workflows/preevy-up.yml
AigulDj: It is when you update/push new commits to your branch

.github/workflows/preevy-up.yml
AigulDj: Yeah, it's something I want to try later

.github/workflows/preevy-up.yml
EraliaKats: here is the thing that I'm fearing. Every time I make a commit I have a new preevy env instead of the one I already have to be updated with the latest code. If that's not possible,  we need to kill the previous version

EraliaKats: Aigul explained to me that it's just rerunning the up command but the AWS Lightsail instance is the same. So I'm happy

EraliaKats (dismissed):

lib/trade_tariff_frontend.rb
willfish: I’m convinced we don’t actually need it since we can just set the environment to preview or something

willfish:
Have you ran through the deployed app and made sure it’s working fully? We can maybe do this together if helpful

.github/workflows/preevy-down.yml
```diff
@@ -0,0 +1,36 @@
+name: Teardown Preevy environment
```
willfish: How do you feel about coupling to the concept and not the technology? So we could call this review app down or something and mention the tech less?

AigulDj:
> This is great work Aigul! Let me know what you think about the comments

Thanks a lot! I went through all your comments and made the updates. Really appreciate the review and the approval!

.github/workflows/preevy-down.yml
AigulDj: > Sounds reasonable

Sounds reasonable to me

willfish:
So I have a few thoughts really:
1. The deployed apps aren't actually using the redis sidecart so this can probably just be deleted/not get deployed
2. We should probably delete the development ECS service for the frontend since we've now got a review app to evaluate frontend changes which is just wonderful
3. It would be good to clearly document which secret/configuration the preevy service is using to configure the application environment since this is absolutely something a developer will want to do to evaluate their code changes

AigulDj:
> PREVIEW_MODE

AigulDj:
> So I have a few thoughts really:
> 
> 1. The deployed apps aren't actually using the redis sidecart so this can probably just be deleted/not get deployed
> 2. We should probably delete the development ECS service for the frontend since we've now got a review app to evaluate frontend changes which is just wonderful
> 3. It would be good to clearly document which secret/configuration the preevy service is using to configure the application environment since this is absolutely something a developer will want to do to evaluate their code changes

That's great, thank you, Will. That also reminded me to create a new role for preview

willfish (approved):
Thanks for helping me to understand the change and get to the bottom of a few issues.

This is great and will help improve the velocity of the team.

Great work!

:sheep: :it: :tada:

PR #2322, opened 2025-05-20
HMRC-1134 Retrieve user from backend

vertism:
### Jira link

[HMRC-1134](https://transformuk.atlassian.net/browse/HMRC-1134)

### What?

- Adds User model with authentication via id token
- Removes decryption of id token

### Why?

Enabling MyOTT to work for multiple users


app/models/user.rb
```diff
@@ -0,0 +1,32 @@
+class User
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  attr_accessor :email, :chapter_ids, :stop_press_subscription
+
+  def self.get(token)
+    response = self.class.get(
```
willfish: How comes we can't just use ApiEntity?

vertism: It's still a bit WIP but I've added ApiEntity now. I need to look into it more to understand if we can handle the custom path and auth headers more directly in ApiEntity.

willfish: Ah fair enough. You ought to be able to yeah. Either by setting the singular or collection path and calling some of the client methods directly

app/models/user.rb
```diff
@@ -0,0 +1,25 @@
+require 'api_entity'
+class User
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+  include ApiEntity
+
+  set_singular_path 'user/users'
```
willfish, line 4: This is a bit unusual for a path in the backend but I must have missed that PR/not noticed

app/models/user.rb
```diff
@@ -0,0 +1,25 @@
...
+    return nil if token.nil?
+
+    super(nil, {}, headers(token))
+  rescue Faraday::UnauthorizedError
+    nil
+  end
+
+  def self.headers(token)
+    {
+      'Content-Type' => 'application/json',
```
willfish: I think this header is implicit as part of the client configuration

spec/models/user_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      let(:body) do
+        {
+          'email' => user.email,
+          'chapter_ids' => user.chapter_ids,
+          'stop_press_subscription' => user.stop_press_subscription,
+        }
+      end
+
+      before do
+        allow_any_instance_of(ApiEntity::ClassMethods).to receive(:api).and_return( # rubocop:disable RSpec/AnyInstance
```
willfish: You could just use webmock here?

spec/models/user_spec.rb
```diff
@@ -0,0 +1,54 @@
+require 'spec_helper'
```
willfish: I don't think you need this chipolata (:eyes: .rspec)

app/models/user.rb
```diff
@@ -0,0 +1,25 @@
+require 'api_entity'
```
willfish: You actually shouldn't need this require with zeitwerk :thinking:

app/models/user.rb
```diff
@@ -0,0 +1,25 @@
...
+class User
+  include ActiveModel::Model
```
willfish: I think ApiEntity includes these two module includes

app/models/user.rb
vertism: It is a new path. I created it as this api is non-public and uses different authentication.

willfish: This will never work in a review app thinking about it

The review app relies on publicly routable ALB paths to avoid needing access to internal networks.

Currently the only externally routable paths in the ALB to the backend under under `/api`

Not sure if this is an actual issue since you can always deploy to development by just adding the `needs-deployment` label

spec/models/user_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    context 'when there is no token' do
+      let(:token) { nil }
+
+      it { is_expected.to be_nil }
+    end
+
+    context 'when response is successful' do
+      before do
+        stub_api_request('/user/users').and_return(jsonapi_response(:user, attributes_for(:user)))
```
willfish, line 15: Nice :)

willfish (approved):

PR #2324, opened 2025-05-20
BAU: Silence skipped production notifications

willfish:
### Jira link

BAU

### What?

I have added/removed/altered:

- [x] Silence skipped notification in production deploys

### Why?

I am doing this because:

- This is noise not signal

### DIFF
--- a/.github/workflows/deploy-to-production.yml
+++ b/.github/workflows/deploy-to-production.yml
@@ -83,7 +83,7 @@ jobs:
   notifications:
       runs-on: ubuntu-latest
       needs: deploy
-      if: always()
+      if: ${{ github.event.workflow_run.conclusion == 'success' || github.event_name == 'workflow_dispatch' }}
       steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-to-production.yml
```diff
@@ -83,7 +83,7 @@ jobs:
   notifications:
       runs-on: ubuntu-latest
       needs: deploy
-      if: always()
+      if: ${{ github.event.workflow_run.conclusion == 'success' || github.event_name == 'workflow_dispatch' }}
```
vertism, line 86: Does this mean deploy failures won't be notified on?

willfish: No it doesn't. It means that if we've skipped the production deployment due to a failing staging run we'll avoid notifications about a skipped run

vertism (approved):

PR #2330, opened 2025-05-21
HMRC-1098 Add MyOTT Subscription Preferences Page & Functionality

HWallenberg:
### Jira link

[HMRC-1098](https://transformuk.atlassian.net/browse/HMRC-1098)

### What?

I have added subscription preferences and integrated functionality with other myott journey pages

### Why?

I am doing this because it is a requirement
image


app/controllers/myott/subscriptions_controller.rb
```diff
@@ -26,11 +28,26 @@ def check_your_answers
...
+        if selection == 'selectChapters'
+          redirect_to myott_chapter_selection_path
+        elsif selection == 'allChapters'
+          redirect_to myott_check_your_answers_path(all_tariff_updates: true)
+        end
+      else
+        flash.now[:error] = 'Please select an option.'
+        render :preference_selection
+      end
+    end
```
vertism: What do you think about doing this as a case statement?

app/views/myott/subscriptions/preference_selection.html.erb
```diff
@@ -0,0 +1,44 @@
...
+            <h2 class="govuk-fieldset__heading govuk-heading-m">What tariff updates would you like to receive?</h2>
+          </legend>
+          <div id="select-hint" class="govuk-hint">
+            Select an option below. You can change your subscription preferences at any time.
+          </div>
+          <%if flash.now[:error] %>
+            <p class="govuk-error-message">
+              Please select an option
+            </p>
+          <%end%>
```
vertism: This is super minor, but I think there should be some extra whitespace here:
```suggestion
          <% if flash.now[:error] %>
            <p class="govuk-error-message">
              Please select an option
            </p>
          <% end %>
```

vertism (dismissed):

vertism (approved):

PR #2332, opened 2025-05-21
Simplify the JS

neilmiddleton:
This PR removes around 500 dependencies that aren't required, and simplifies a bunch of what's left.

There's a bunch of code here that's been removed which didn't appear to be doing anything at all, mostly in `commodities.js`


app/javascript/src/commodities.js
```diff
@@ -1,178 +1,13 @@
...
-          var evt = document.createEventObject();
-          this.trigger('click');
-        }
-      },
-      /**
-          @name cookies
-          @object
-          @description controls the getting and setting of cookies
-        */
-      cookies: {
```
willfish, line 52: Amazing. I can't see that this is used anywhere must have been missed in a migration :tada:

app/javascript/src/commodities.js
```diff
@@ -1,178 +1,13 @@
...
-        this.$tables.each(function() {
-          const table = $(this);
-
-          table.find('dt.has_children').each(function() {
-            const dt = $(this);
-            const secondColumn = dt.closest('td').next();
-            let height = 0;
-            secondColumn.find('span.table-line').each(function() {
-              height += $(this).outerHeight();
+document.addEventListener('DOMContentLoaded', () => {
```
willfish, line 1: I do wonder if this file now could have a different name :thinking:

config/environments/development.rb
```diff
@@ -65,4 +65,6 @@
 
   # Allow all preview subdomains from livecycle.run
   config.hosts << /[a-z0-9-]+\.livecycle\.run/
+
+  config.file_watcher = ActiveSupport::EventedFileUpdateChecker
```
willfish, line 69: I figured this was automatic :confused: but I have had issues locally with the mapped paths triggering an automatic reload/refingerprinting

vendor/javascript/govuk-frontend.js
```diff
@@ -0,0 +1,4 @@
+// govuk-frontend@5.10.0 downloaded from https://ga.jspm.io/npm:govuk-frontend@5.10.0/dist/govuk/all.mjs
```
willfish, line 1: Not sure why we're vendoring this chipolata?

willfish:
I'm guessing search is still working on chrome 86?

willfish (dismissed):

willfish:
This is awesome, I deployed it to dev and can see that a the clipboard copy stuff is working as expected

vendor/javascript/govuk-frontend.js
neilmiddleton: Pinning will pull it into vendor

config/environments/development.rb
neilmiddleton: It may not be needed, but I was getting big problems without it
https://guides.rubyonrails.org/configuring.html#config-file-watcher

package.json
```diff
@@ -1,22 +1,39 @@
 {
   "name": "trade-tariff-frontend",
   "private": true,
+  "dependencies": {
+    "@stimulus/polyfills": "^2.0.0",
```
willfish, line 5: So all of these are actually development dependencies/shouldn't end up in the docker image

config/environments/development.rb
willfish: Okay, let's leave it in

package.json
neilmiddleton: I think these are needed for older browsers.

willfish: Okey poke

willfish (approved):

PR #2338, opened 2025-05-27
Implement accessibility checks for quota search feature

Sandy25design:
### Jira link

[HMRC-<1090> <https://transformuk.atlassian.net/browse/HMRC-1090>)]

### What?
This PR adds an automated accessibility test for the quota search functionality. The test ensures that the user interface for this feature is compliant with accessibility standards, particularly focusing on:

- Proper use of ARIA roles and attributes
- Keyboard navigation and focus handling
- Screen reader compatibility and semantic structure

The test helps identify any accessibility barriers in the quota search screen, ensuring it's usable for all users, including those relying on assistive technologies.

### Why?

I am doing this because it complies with WCAG accessibility standards.
- Improves usability for users with disabilities
- Supports broader accessibility efforts across the application
- Prevents regressions as UI evolves

### DIFF
--- a/spec/javascript/accessibility/quota_search.spec.js
+++ b/spec/javascript/accessibility/quota_search.spec.js
@@ -1,0 +1,40 @@
+const { test, expect } = require("@playwright/test");
+const { LoginPage } = require("./pages/loginPage");
+const AxeBuilder = require("@axe-core/playwright").default;
+
+test.describe("Search for quotas", () => {
+  test("Validate UK quota Search Results", async ({ page }, testInfo) => {
+    await new LoginPage("/quota_search", page, testInfo).login();
+
+    await expect(
+      page.getByRole("heading", { name: "Search for quotas" })
+    ).toBeVisible();
+    await page
+      .getByRole("textbox", { name: "Enter the 6-digit quota order" })
+      .fill("052016");
+
+    //Select a country to which the quota applies
+    await page
+      .getByRole("combobox", { name: "Select a country to which the" })
+      .click();
+    await page.getByRole("option", { name: "Albania (AL)" }).click();
+
+    //search for quotas
+    await page.getByRole("button", { name: "Search for quotas" }).click();
+
+    //Validation
+    await expect(
+      page.getByRole("heading", { name: "Quota search results" })
+    ).toBeVisible();
+    await expect(
+      page.getByRole("columnheader", { name: "Order number" })
+    ).toBeVisible();
+    await expect(
+      page.getByRole("button", { name: "Opens in a popup" })
+    ).toHaveText("052016");
+
+    const accessibilityScanResults = await new AxeBuilder({ page }).analyze();
+
+    expect(accessibilityScanResults.violations.length).toBeLessThan(5);
+  });
+});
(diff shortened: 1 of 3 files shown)


spec/javascript/accessibility/quota_search.spec.js
```diff
@@ -0,0 +1,40 @@
+const { test, expect } = require("@playwright/test");
+const { LoginPage } = require("./pages/loginPage");
+const AxeBuilder = require("@axe-core/playwright").default;
+
+test.describe("Search for quotas", () => {
+  test("Validate UK quota Search Results", async ({ page }, testInfo) => {
```
willfish, line 6: So I don't think we need any of this orchestration when displaying the results since results are just URL based.

We can just navigate here, say:

```https://dev.trade-tariff.service.gov.uk/quota_search?order_number=052012&goods_nomenclature_item_id=&geographical_area_id=&day=29&month=5&year=2025&critical=&status=```

And save loads of time

File: spec/javascript/accessibility/quota_search.spec.js [hunk 2]
```diff
@@ -0,0 +1,40 @@
...
+    await expect(
+      page.getByRole("columnheader", { name: "Order number" })
+    ).toBeVisible();
+    await expect(
+      page.getByRole("button", { name: "Opens in a popup" })
+    ).toHaveText("052016");
+
+    const accessibilityScanResults = await new AxeBuilder({ page }).analyze();
+
+    expect(accessibilityScanResults.violations.length).toBeLessThan(5);
```
willfish, line 11: So I think we can strip this right back to just validating the accessiblity of rendered results and reduce the runtime

spec/javascript/accessibility/quota_search.spec.js
Sandy25design: That makes sense if the results page can be directly accessed via URL without any prior interactions. In this case though, the page we’re testing requires login and user input before the results are shown, so we need to go through the full flow (login, fill the form, select a country, search) before axe can analyse the correct content. So it's not just about visiting a static URL, it's about getting the page into the right state for an accurate accessibility check.

willfish: I'll have a play but I suspect we don't :)

Sandy25design: @willfish, are you suggesting we skip the login and form submission steps, and just run the accessibility test by navigating straight to that URL? I can do that if the page loads correctly on its own, but I want to make sure it reflects the same state as when a user goes through the full flow. Let me know your thoughts about it.

willfish: So what we're doing is rendering some html elements in different states in the page. When there are rows to display we show the results section but the experience is the same for whatever method you get there.

The notion you're describing is really a behavioural one and not about axxy and is more something an e2e test would drive. If we can get to the state we want quickly/with fewer steps we should definitely take it . We won't always be able too but we can here

willfish (approved):

spec/javascript/accessibility/quota_search.spec.js
Sandy25design: Got it, that makes sense. Since the accessibility check is based on the final HTML state, and not the full user flow, it’s more efficient to go straight to the results page URL when we can. I’ll adjust the test to use the direct URL in this case, and we can always fall back to the full flow if needed for other scenarios.

PR #2340, opened 2025-05-28
HMRC-1028 MyOTT CYA Screen Chapter Preferences Update and Retrieval Functionality

HWallenberg:
### Jira link

[HMRC-1028](https://transformuk.atlassian.net/browse/HMRC-1028)

### What?

I have added Chapter update, confirmation and retrieval functionality with screens

### Why?

I am doing this because it is a project requirement


app/controllers/myott/subscriptions_controller.rb
```diff
@@ -6,14 +6,40 @@ class SubscriptionsController < MyottController
 
     before_action :sections_chapters, only: %i[chapter_selection check_your_answers]
 
+    before_action :ensure_subscription_in_progress, only: %i[chapter_selection check_your_answers subscribe]
```
vertism: What's the purpose of this? Can't the user change their selection at any time?

app/controllers/myott/subscriptions_controller.rb
```diff
@@ -6,14 +6,40 @@ class SubscriptionsController < MyottController
...
+      subscribed_to_stop_press = current_user&.stop_press_subscription || false
+
+      return redirect_to myott_preference_selection_path unless subscribed_to_stop_press
+
+      session[:chapter_ids] = if current_user&.chapter_ids&.split(',')&.any?
+                                current_user&.chapter_ids&.split(',')
+                              else
+                                all_chapters.map(&:to_param)
+                              end
+      @selected_chapters = get_selected_chapters(Array(session[:chapter_ids]))
```
vertism, line 19: Does this need to check the session? Can't the value from the user be used?

app/controllers/myott/subscriptions_controller.rb
```diff
@@ -28,25 +54,31 @@ def check_your_answers
...
+      if params[:all_tariff_updates] == 'true'
+        session[:chapter_ids] = []
+      end
+      updated_user = User.update(cookies[:id_token], chapter_ids: session[:chapter_ids].join(','), stop_press_subscription: 'true')
+      Rails.logger.info("User updated: #{updated_user.inspect}")
+      if updated_user
+        session.delete(:chapter_ids)
+        session.delete(:all_tariff_updates)
+        session[:subscription_in_progress] = false
+        render :subscription_confirmation
```
vertism: I think this should redirect to subscription_confirmation otherwise refreshing the page will try to re-subscribe.

app/controllers/myott/subscriptions_controller.rb
HWallenberg: Good question! I was thinking the user (once updated and confirmed) might decide to click on the 'back' button and go through the journey again.  
It's supposed to be a guard against that and also potentially trying to navigate one of the stages without going to the start of the journey.  A determined user might find a way around these things but I thought it would be useful to guard against more casual users.  What do you think?

app/controllers/myott/subscriptions_controller.rb
HWallenberg: Adding to the session here so they are available if someone decides to edit their preferences.

It's a link not a submit to the chapter_selection page and this will persist the preferences without a POST back to server.  

Same pattern used to edit selection from the Check Your Answers page.

Also UX requires all chapters are shown if user originally selected 'all chapters' (ie `user.chapter_ids == nil`)

So it seems correct to add chapter_ids to session when we can for manipulation by the user.

vertism: On this page, won't we only show the saved preferences?

app/controllers/myott/subscriptions_controller.rb
vertism: Could you do the same thing by checking whether session[:chapter_ids] exists instead of needing to create a new variable?

app/controllers/myott/subscriptions_controller.rb
HWallenberg: Yes, however the saved preferences of chapter_ids might equal '' (or nil) if they originally chose 'all chapters'. 

We need a way of showing 'all of the chapters' and I'm adding to the session in case they decide to edit their preferences.

Assigning to `@selected_chapters` to display saved preferences as it's reusing partial `_details_selected_chapters`

Hope this makes sense?

app/controllers/myott/subscriptions_controller.rb
HWallenberg: Do you mean something like this..?

```
def ensure_subscription_in_progress
  redirect_to myott_path unless session[:chapter_ids]
end
```

vertism: Yeah, you might have to use nil? so it doesn't match on an empty array.

app/controllers/myott/subscriptions_controller.rb
vertism: OK, I understand what you're saying now. Makes sense.

app/views/myott/subscriptions/subscription_confirmation.html.erb
```diff
@@ -0,0 +1,31 @@
+<script>
+  window.history.replaceState({}, document.title, window.location.href);
+</script>
```
vertism: Is this still needed?

app/controllers/myott/subscriptions_controller.rb
HWallenberg: that's done thanks.

app/views/myott/subscriptions/subscription_confirmation.html.erb
HWallenberg: no!

app/controllers/myott/subscriptions_controller.rb
```diff
@@ -28,25 +53,32 @@ def check_your_answers
...
+      if updated_user
+        session.delete(:chapter_ids)
+        session.delete(:all_tariff_updates)
+        redirect_to myott_subscription_confirmation_path
       else
-        flash.now[:error] = 'Please select an option.'
-        render :preference_selection
+        flash.now[:error] = 'There was an error updating your subscription. Please try again.'
+        @selected_chapters = get_selected_chapters(Array(session[:chapter_ids]))
+        render :check_your_answers
```
vertism: Does @all_tariff_updates need to be set if we are re-rendering check your answers?

HWallenberg: yes, committed some new code.

vertism (approved):

PR #2341, opened 2025-05-29
Configure SMTP to use AWS SES for email delivery

AigulDj:
### Jira link

[HMRC1125](https://transformuk.atlassian.net/browse/HMRC-1125)

### What?

I have configured SMTP settings to enable sending emails via AWS SES.
Secrets (SMTP credentials) are fetched securely from AWS Secrets Manager.

### Why?

Feedback submission emails were failing because SES v2 was not accessible from the Lightsail preview environment.
This change ensures authenticated SMTP access to SES, allowing feedback to be sent properly in preview deployments.


.github/workflows/preevy-up.yml
```diff
@@ -29,6 +29,11 @@ jobs:
 
       - uses: actions/checkout@v4
 
+      - uses: trade-tariff/trade-tariff-tools/.github/actions/fetch-secrets@main
+        with:
+          secret-name: frontend-configuration
+          env-file: GITHUB_ENV
```
willfish: Does this work :exploding_head:

config/environments/production.rb
```diff
@@ -101,4 +101,16 @@
   }
 
   config.action_mailer.delivery_method = :ses_v2
+
+  if ENV['USE_SMTP'] == 'true'
```
willfish, line 103: Is this either or re the delivery method above?

.github/workflows/preevy-up.yml
AigulDj: Seems like not, I'm looking for a way to provide SMTP secrets saved in the secrets manager, don't know the way yet

config/environments/production.rb
AigulDj: > Is this either or re the delivery method above?
if ENV['USE_SMTP'] == 'true' is selecting between two delivery methods:
ses_v2 or :smtp

In our case, we’ve set USE_SMTP=true in the Docker Compose file to enable actual feedback delivery via SMTP. I tested this with explicit SMTP_USERNAME and SMTP_PASSWORD, and it worked as expected

.github/workflows/preview-up.yml
```diff
@@ -29,6 +30,17 @@ jobs:
 
       - uses: actions/checkout@v4
 
+      - run: sudo apt-get install -y jq
```
willfish: I think this comes with the runner image? https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

willfish (dismissed):

willfish (approved):

PR #2343, opened 2025-06-02
HMRC-1092 Start page

vertism:
### Jira link

[HMRC-1092](https://transformuk.atlassian.net/browse/HMRC-1092)

### What?

Start page for subscribing to stop press alerts

Screenshot 2025-06-02 at 12 00 50


willfish (approved):
Do we need a controller/request spec for it? TBH it does very little so maybe not

vertism:
> Do we need a controller/request spec for it? TBH it does very little so maybe not

I didn't think it was worth it, but there will eventually be some feature tests that will cover this page.

PR #2346, opened 2025-06-02
HMRC-1092 Redirect to login if user is not valid

vertism:
### Jira link

[HMRC-1092](https://transformuk.atlassian.net/browse/HMRC-1092)

### What?

The user is taken to the login page if their session is not valid. This is disabled in the development environment.

### Why?

Users need to be authenticated to login

### DIFF
--- a/app/controllers/myott/subscriptions_controller.rb
+++ b/app/controllers/myott/subscriptions_controller.rb
@@ -5,16 +5,17 @@ module Myott
                   :disable_last_updated_footnote
 
     before_action :all_sections_chapters, only: %i[chapter_selection check_your_answers]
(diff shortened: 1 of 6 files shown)


app/controllers/myott/subscriptions_controller.rb
```diff
@@ -5,16 +5,17 @@ class SubscriptionsController < MyottController
                   :disable_last_updated_footnote
 
     before_action :all_sections_chapters, only: %i[chapter_selection check_your_answers]
+    before_action :authenticate, except: %i[start]
```
HWallenberg, line 8: nice. 👍

HWallenberg (approved):
LGTM 👍

PR #2347, opened 2025-06-02
HMRC-1093 Add subscribe links to Stop Press News Collection Items

HWallenberg:
### Jira link

[HMRC-1093](https://transformuk.atlassian.net/browse/HMRC-1093)

### What?

I have added subscribe links for stop press news collection type

### Why?

I am doing this because it is required for MyOTT when Public Beta Testing Starts

This means the Stop Press News Collection Description can be set to nil in the database by a backend data migration with this PR:
https://github.com/trade-tariff/trade-tariff-backend/pull/2293

**Change to Stop Press News** 
image

### DIFF
--- a/app/models/news/collection.rb
+++ b/app/models/news/collection.rb
@@ -12,6 +12,13 @@ module News
                   :description,
                   :priority
 
+    enum :id, {
+      trade_news: [1],
+      tariff_notices: [2],
+      tariff_stop_press: [3],
+      service_updates: [4],
+    }
+
     def id
       @id ||= resource_id.presence&.to_i
     end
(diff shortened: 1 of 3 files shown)


app/models/news/collection.rb
```diff
@@ -29,5 +29,9 @@ def matches_param?(collection_id)
         false
       end
     end
+
+    def stop_press?
```
willfish: You could use the enum functionality here? Can be convenient for this kind of thing

HWallenberg: Interesting learning point with `APIEntity>Enum` thanks. 🙌

PR #2366, opened 2025-06-05
Create Live Issues page with hard-coded data

EraliaKats:
### Jira link

[HMRC-<TODO>](https://transformuk.atlassian.net/browse/HMRC-<TODO>)

### What?

I have a...

### Why?

I am doing this because...

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -59,7 +59,7 @@ Rails.application.routes.draw do
   get '/news/stories/:id', to: 'news_items#show', as: :news_item
   resources :news_items, only: %i[index show], path: '/news'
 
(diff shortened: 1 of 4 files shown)


config/routes.rb
```diff
@@ -59,6 +59,8 @@
   get '/news/stories/:id', to: 'news_items#show', as: :news_item
   resources :news_items, only: %i[index show], path: '/news'
 
+  get '/live_issues', to: 'live_issues#index', as: :live_issues
```
willfish:
```suggestion
  get '/live_issues', to: 'live_issues#index'

PR #2390, opened 2025-06-11
HMRC-1151: set up tests

AigulDj:
### Jira link

[HMRC-1151](https://transformuk.atlassian.net/browse/HMRC-1151)

### What?

- Added an `e2e-tests` job to the `preview-up` workflow.
- Removed docker compose version lock steps

### Why?

- To ensure end-to-end tests run automatically during the preview deployment process.
- We don't need to lock compose version anymore after Preevy source code was updated


.github/workflows/preview-tests.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Get preview environment URL
+        id: get-preview-url
+        run: |
+          touch .env
+          PREVIEW_URL=$(npx preevy urls --json --wait --profile "$PREEVY_PROFILE_URL" | jq -r '.[].url')
+          echo "✅ Using preview URL: $PREVIEW_URL"
+          echo "BASE_URL=$PREVIEW_URL" >> "$GITHUB_ENV"
+
+      - name: Run Playwright tests
+        run: npx playwright test spec/javascript/accessibility/preview-check.spec.js
```
willfish:
```suggestion
        run: yarn run axxy
```

spec/javascript/accessibility/preview-check.spec.js
```diff
@@ -0,0 +1,70 @@
+const { test, expect } = require("@playwright/test");
+
+test.describe('Preview environment smoke test', () => {
```
willfish: So the preview environment is exactly the same as the regular environment just with a different URL. We ought to be able to just call the e2e test workflow with the preevy url

willfish: Also, this isn't an accessibility test

willfish:
So I think you can just use the following and save on writing e2e tests inside of the frontend accessibility tests:

```
  e2e-tests:
    uses: trade-tariff/trade-tariff-tools/.github/workflows/e2e-tests.yml@main
    with:
      test-url: "<preevy-url>"
      admin-test-url: "https://admin.dev.trade-tariff.service.gov.uk" # Unaffected but just to make it work
    secrets:
      basic_password: ${{ secrets.BASIC_PASSWORD }}
```

spec/javascript/accessibility/preview-check.spec.js
AigulDj: Yeah, that was one of the questions, if I place it in another directory then will I have to create another config file or update the existing one etc

.github/workflows/preview-up.yml
```diff
@@ -68,3 +72,17 @@ jobs:
...
+          touch .env
+          PREVIEW_URL=$(npx preevy urls --json --wait --profile "$PREEVY_PROFILE_URL" | jq -r '.[0].url')
+          echo "✅ Using preview URL: $PREVIEW_URL"
+          echo "preview_url=$PREVIEW_URL" >> "$GITHUB_OUTPUT"
+
+  e2e-tests:
+    needs: deploy
+    uses: trade-tariff/trade-tariff-tools/.github/workflows/e2e-tests.yml@main
+    with:
+      test-environment: preview
```
willfish: So you can set the environment variables:
```suggestion
      test-environment: development
   env:
      BASE_URL: ${{ needs.deploy.outputs.preview_url }}
   secrets:
      basic-password: ${{ secrets.BASIC_PASSWORD }}
```

.github/workflows/preview-down.yml
```diff
@@ -24,8 +24,8 @@ jobs:
       - uses: actions/checkout@v4
       - run: touch .env # To prevent preevy from failing due to missing .env file
 
-      - name: Lock Docker Compose to v2.36.0
-        uses: trade-tariff/trade-tariff-tools/.github/actions/lock-docker-compose@main
+      # - name: Lock Docker Compose to v2.36.0
```
willfish: Should we just remove this?

.github/workflows/preview-up.yml
```diff
@@ -42,29 +45,33 @@ jobs:
           echo "$SECRET_JSON" | jq -r 'to_entries[] | "::add-mask::\(.value)"'
           echo "$SECRET_JSON" | jq -r 'to_entries[] | "\(.key)=\(.value)"' >> .env
 
-      - name: Lock Docker Compose to v2.36.0
-        uses: trade-tariff/trade-tariff-tools/.github/actions/lock-docker-compose@main
+      # - name: Lock Docker Compose to v2.36.0
```
willfish: Same here :)

willfish (approved):

PR #2397, opened 2025-06-16
HMRC-1259 Verified User Bug Fix

HWallenberg:
### Jira link

[HMRC-1259](https://transformuk.atlassian.net/browse/HMRC-1259)

### What?

I have update `invalid` controller function to check for a current_user.
If there is user is redirected to dashboard or preference selection.

### Why?

I am doing this so a user can click on their password-less identity link to verify, close the window and still be authenticated.

### DIFF
--- a/app/controllers/myott/subscriptions_controller.rb
+++ b/app/controllers/myott/subscriptions_controller.rb
@@ -60,10 +60,7 @@ module Myott
       @selected_sections_chapters = get_selected_sections_chapters(Array(session[:chapter_ids]))
     end
 
-    def preference_selection
-      Rails.logger.info("cookies[:id_token]: #{cookies[:id_token].inspect}")
-      Rails.logger.info("authenticate: #{authenticate.inspect}")
-    end
+    def preference_selection; end
 
     def subscribe
       if params[:all_tariff_updates] == 'true'
--- a/app/controllers/myott/subscriptions_controller.rb
+++ b/app/controllers/myott/subscriptions_controller.rb
@@ -6,8 +6,7 @@ module Myott
     def start; end
 
     def invalid
-      # HMRC-1259 user re-using token could still be valid
-      if cookies[:id_token].present?
+      unless current_user.nil?
         redirect_to myott_path
       end
     end
--- a/app/controllers/myott/subscriptions_controller.rb
+++ b/app/controllers/myott/subscriptions_controller.rb
@@ -6,7 +6,7 @@ module Myott
     def start; end
 
     def invalid
-      unless current_user.nil?
+      if current_user.present?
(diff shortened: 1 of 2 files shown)


willfish (dismissed):

app/controllers/myott/subscriptions_controller.rb
```diff
@@ -54,7 +60,10 @@ def check_your_answers
       @selected_sections_chapters = get_selected_sections_chapters(Array(session[:chapter_ids]))
     end
 
-    def preference_selection; end
+    def preference_selection
+      Rails.logger.info("cookies[:id_token]: #{cookies[:id_token].inspect}")
+      Rails.logger.info("authenticate: #{authenticate.inspect}")
```
vertism: Did you mean to leave these here?

app/controllers/myott/subscriptions_controller.rb
```diff
@@ -4,7 +4,13 @@ class SubscriptionsController < MyottController
     before_action :authenticate, except: %i[start invalid]
 
     def start; end
-    def invalid; end
+
+    def invalid
+      # HMRC-1259 user re-using token could still be valid
+      if cookies[:id_token].present?
+        redirect_to myott_path
+      end
```
vertism: Do you think it's worth hitting the API to also check the JWT is valid?

app/controllers/myott/subscriptions_controller.rb
HWallenberg: d'oh! 🤦

app/controllers/myott/subscriptions_controller.rb
HWallenberg: I'm not sure, does `authenticate` do this when re-directed to dashboard?

vertism: authenticate will redirect which we don't want, but checking if current_user exists should do the trick.

HWallenberg: okay will go with that 👍

app/controllers/myott/subscriptions_controller.rb
```diff
@@ -4,7 +4,12 @@ class SubscriptionsController < MyottController
     before_action :authenticate, except: %i[start invalid]
 
     def start; end
-    def invalid; end
+
+    def invalid
+      unless current_user.nil?
```
vertism: Personal preference, but I find the double negative confusing. Up to you though.
```suggestion
      if current_user.present?
```

HWallenberg: no worries!

vertism (approved):

PR #2410, opened 2025-06-18
Refactor subscriptions controller

vertism:
### What?

- Creates a PreferencesController for selecting chapter preferences
- Creates SessionChaptersDecorator to handle presentation methods
- Feature tests

### Why?

I am doing this to make future changes easier


HWallenberg (dismissed):

HWallenberg (dismissed):
This is a much more elegant solution @vertism thank you for refactoring.  It looks good to me, I like the way you've introduced decorators and separated out subscriptions from preferences.  Interesting to see what you've done with the routing to make it more 'railsy'.
As you say it will make it easier to expand upon. 🙌

HWallenberg (approved):

PR #2413, opened 2025-06-19
Cache the commodity page

neilmiddleton:
This is an experiment in fragment caching on one of the heaviest pages we have.

We're adding a bunch of caching here, all centered around the date that `/uk/api/v2/updates/latest` returns.

* The page content itself is cached as much as possible.  Once we've rendered it once, we use it as much as we can
* We cache things like ROO as much as possible for the day for use across different codes
* We cache individual headings/commodities for the day
* We cache geographical areas/countries for the day for re-use in multiple places.

Initial testing locally drops performance from 100-200ms+ to sub 10ms on a warm cache.


HWallenberg (approved):
lgtm

PR #2414, opened 2025-06-19
HMRC-1210: set up preview monitoring

AigulDj:
### Jira link

[HMRC-1210](https://transformuk.atlassian.net/browse/HMRC-1210)

### What?

- Added log capture for preview envs
- Added cost estimates and cleanup behaviour in the PR comment for visibility.

### Why?

- I am doing this because failing silently is annoying, and we need the logs when things break.
Logs now auto-upload as artifacts for 7 days.
- Adding cleanup policy info helps the team understand when and how environments are removed and how to keep them around when needed.

### DIFF
--- a/.github/workflows/preview-down.yml
+++ b/.github/workflows/preview-down.yml
@@ -23,6 +23,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - run: touch .env # To prevent preevy from failing due to missing .env file
+      - name: Lock Docker Compose to v2.36.0
+        run: |
+          mkdir -p ~/.docker/cli-plugins/
+          curl -sSL https://github.com/docker/compose/releases/download/v2.36.0/docker-compose-linux-x86_64 -o ~/.docker/cli-plugins/docker-compose
+          chmod +x ~/.docker/cli-plugins/docker-compose
 
       - uses: aws-actions/configure-aws-credentials@v4
         with:
--- a/.github/workflows/preview-down.yml
+++ b/.github/workflows/preview-down.yml
@@ -17,7 +17,7 @@ env:
 
 jobs:
   teardown:
-    if: github.event.action == 'closed' || (github.event.action == 'unlabeled' && github.event.label.name == 'needs-preview')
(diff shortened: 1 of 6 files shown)


.github/workflows/preview-down.yml
```diff
@@ -17,12 +17,17 @@ env:
...
 jobs:
   teardown:
-    if: github.event.action == 'closed' || (github.event.action == 'unlabeled' && github.event.label.name == 'needs-preview')
+    if: github.event.action == 'closed' || (github.event.action == 'unlabeled' && (github.event.label.name == 'needs-preview' || github.event.label.name == 'keep-preview'))
     permissions: write-all
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
       - run: touch .env # To prevent preevy from failing due to missing .env file
+      - name: Lock Docker Compose to v2.36.0
```
willfish, line 27: I wonder if we could put this inside an action? We might want to share it later

.github/workflows/scripts/capture-log.sh
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+echo "📄 Starting log capture process..."
+LOG_DIR="logs"
```
willfish:
```suggestion
LOG_DIR="log"
```

Might as well piggy back off of the directory with logs in already?

willfish (dismissed):

.github/workflows/scripts/capture-log.sh
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
```
willfish: Just a preference thing but with the top line I don't think we need .sh suffix on the file

.github/workflows/scripts/capture-log.sh
AigulDj: Where is that?

willfish: I was assuming we’d checked out the code so in the repo already

willfish (approved):

PR #2435, opened 2025-06-24
More FE caching

neilmiddleton:
Adds a cache to headings, improves the cache key for a better hit ratio, and also caches news items in the headers.


app/views/layouts/application.html.erb
```diff
@@ -59,7 +59,11 @@
       <%= render 'shared/govuk_logo' %>
       <%= yield :proposition_header %>
     </div>
-    <%= render 'news_items/header_banner', news_item: News::Item.latest_banner unless @skip_news_banner %>
+
+    <%= Rails.cache.fetch(['news_items/header_banner', cache_key, expires_in: 1.hour]) do %>
```
willfish, line 63: Might be worth Gabriel knowing about this. Effectively the change will lead to a 1 hour lag in news items appearing on the banner

willfish (approved):

PR #2436, opened 2025-06-25
Fix cache key on import cache

neilmiddleton:
Adds extra clarity to a cache key preventing collisions.

### DIFF
--- a/spec/models/commodity_spec.rb
+++ b/spec/models/commodity_spec.rb
@@ -18,6 +18,10 @@ RSpec.describe Commodity do
       ]
     end
 
+    before do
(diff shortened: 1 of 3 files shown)


willfish (dismissed):

spec/models/commodity_spec.rb
```diff
@@ -18,6 +18,10 @@
       ]
     end
 
+    before do
```
willfish, line 21: Do we want to validate the message sending to this guy or can we remove it?

neilmiddleton: I think we're good here.

willfish (approved):

PR #2438, opened 2025-06-25
HMRC-1306: Drop long-lived frontend in development

willfish:
### Jira link

[HMRC-1306](https://transformuk.atlassian.net/browse/HMRC-1306)

### What?

I have a...

- Removed the ECS frontend in development 

### Why?

I am doing this because...

- This has been replaced by preevy which will build our image/precompile assets, etc, and without developer collisions


willfish:
Going for a different approach

PR #2444, opened 2025-06-30
HMRC-1384: Cleanup preview

AigulDj:
### Jira link

[HMRC-1384](https://transformuk.atlassian.net/browse/HMRC-1384)

### What?

I have:
- Removed the cleanup workflow.
- Removed the unused database volume from compose.yml.

### Why?

I am doing this because:
- The preview-cleanup.yml workflow has been moved to the trade-tariff-tools repository, so these are no longer needed here.


compose.yml
```diff
@@ -17,9 +17,6 @@ services:
       - USE_SMTP=true
     extra_hosts:
       - "telemetry.preevy.dev:127.0.0.1" # Disables telemetry
-volumes:
```
willfish, line 20: :dancers:

willfish (approved):

compose.yml
AigulDj: That's odd, I don't see that in my code

PR #2459, opened 2025-07-09
PEX-2 | Add live issues feature

EraliaKats:
### Jira link

[PEX-2](https://transformuk.atlassian.net/browse/PEX-2)

### What?

- [x] Add the live issues feature

### DIFF
--- a/app/models/live_issue.rb
+++ b/app/models/live_issue.rb
@@ -3,7 +3,6 @@ class LiveIssue
 
   set_collection_path 'api/v2/live_issues'
 
-  attr_reader :id
   attr_accessor :title,
                 :description,
                 :status,
--- a/app/models/live_issue.rb
+++ b/app/models/live_issue.rb
@@ -1,0 +1,15 @@
+class LiveIssue
+  include UkOnlyApiEntity
+
+  set_collection_path 'api/v2/live_issues'
+
+  attr_reader :id
+  attr_accessor :title,
+                :description,
+                :status,
+                :commodities,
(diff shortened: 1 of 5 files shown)


app/models/live_issue.rb
```diff
@@ -0,0 +1,15 @@
+class LiveIssue
+  include UkOnlyApiEntity
+
+  set_collection_path 'api/v2/live_issues'
+
+  attr_reader :id
```
willfish: I think you'll have this as resource_id by default so shouldn't need this

app/models/live_issue.rb
```diff
@@ -0,0 +1,14 @@
+class LiveIssue
+  include UkOnlyApiEntity
+
+  set_collection_path 'api/v2/live_issues'
```
willfish, line 4: You actually don't need this :)

willfish (approved):
Not sure if we want to add a test but its simple enough that I'm honestly not that bothered

app/models/live_issue.rb
EraliaKats: It's a readability thing for me. There's to places the data could come from, having this there helps understand which one without thinking much

willfish: So this was basically added as a feature to enable special snowflake routes that were essentially non-json-api compliant - think convention over configuration

willfish: Everything about json api semantics is predictable which is why its awesome

PR #2491, opened 2025-07-31
PRDEX-70 Generic Enquiry Form

HWallenberg:
### Jira link

[HMRC-PRDEX-70](https://transformuk.atlassian.net/browse/HMRC-PRDEX-70)

### What?

I have implemented generic enquiry form for Product Experience

### Why?

I am doing this because it was required for the ticket.

The backend [PR](https://github.com/trade-tariff/trade-tariff-backend/pull/2432) for this.


app/controllers/product_experience/enquiry_form_controller.rb
```diff
@@ -0,0 +1,121 @@
+module ProductExperience
+  class EnquiryFormController < ApplicationController
```
willfish, line 2: So I'm not sure what we did for the subscription wizard but how do you feel about using wizard steps for the enquiry form just to reduce the number of approaches in the frontend?

Appreciate this is a PITA

app/models/enquiry_form.rb
```diff
@@ -0,0 +1,23 @@
...
+      data: {
+        attributes: attributes,
+      },
+    }
+
+    body = json_api_params.to_json
+    headers = { 'Content-Type' => 'application/json' }
+
+    super(body, headers)
+  rescue Faraday::ClientError => e
```
willfish: I think this error will get handled by super

willfish: So probably can remove handling here

app/views/product_experience/enquiry_form/check_your_answers.html.erb
```diff
@@ -0,0 +1,31 @@
+<% enquiry_form_page_title "Enquiry Form Check Your Answers" %>
```
willfish: I'm guessing a title is not specific to this feature and we can reuse the existing implementation for titling and just add support for error presentation?

app/views/product_experience/enquiry_form/check_your_answers.html.erb
```diff
@@ -0,0 +1,31 @@
...
+<% content_for :back_link do %>
+  <%= back_link url_for(controller: 'product_experience/enquiry_form', action: 'form', field: @prev_field) %>
```
willfish: Confused why you can't use the standard path helper here?

config/brakeman.ignore
```diff
@@ -1,5 +1,39 @@
...
+      "file": "app/views/product_experience/enquiry_form/form.html.erb",
+      "line": 14,
+      "link": "https://brakemanscanner.org/docs/warning_types/dynamic_render_path/",
+      "code": "render(partial => partial_for_field(params[:field]), { :locals => ({ :field => params[:field] }) })",
+      "render_path": [
+        {
+          "type": "controller",
+          "class": "ProductExperience::EnquiryFormController",
+          "method": "form",
+          "line": 19,
```
willfish, line 18: I think you've actually fixed this?

willfish: Ah okay reading into it its a false positive anyway since you constrain the list of templates using a fixed map

config/routes.rb
```diff
@@ -79,6 +79,19 @@
...
   end
 
+  if TradeTariffFrontend.enquiry_form?
+    namespace :product_experience, path: '', as: 'product_experience' do
+      scope path: 'enquiry_form', as: 'enquiry_form', controller: 'enquiry_form' do
+        get '/', action: 'show'
+        get 'check_your_answers'
+        post 'submit', action: 'submit_form', as: 'submit_form'
+        get 'confirmation'
+        get ':field', action: 'form'
```
willfish: Do we want to add the constraints for possible steps, here?

spec/spec_helper.rb
```diff
@@ -1,5 +1,7 @@
 ENV['RAILS_ENV'] ||= 'test'
 
+ENV['ENQUIRY_FORM_ENABLED'] = 'true'
```
willfish: We can move this to `.env.test`

config/routes.rb
HWallenberg: Yes, can do.  It was what I did originally and then removed it in favour of `validate_field` which redirected back to the start page if an erroneous path (disallowed field) was entered.

A page not found will show if someone tries to type in a forbidden path, with this routing change:

```
 constraints field: /full_name|company_name|occupation|email_address|category|query/ do
          get ':field', action: 'form'
          post ':field', action: 'submit'
        end
```

Do you want me to add this and keep `validate_field`

app/controllers/product_experience/enquiry_form_controller.rb
HWallenberg: The subscription pages aren't using a wizard AFAIK.

Similar routing that I've tried to make as generic as possible with show, form, CYA and submit actions.  This is similar to `subscriptions_controller.rb`

I've mentioned more about routing and the constraints in another comment.

Happy to discuss about what you think is best.

config/brakeman.ignore
HWallenberg: yeah brakeman strict but dumb IMO but hope this is okay.

app/controllers/product_experience/enquiry_form_controller.rb
HWallenberg: Discussion with NM and EK yesterday who decided not to refactor into wizard

app/views/product_experience/enquiry_form/check_your_answers.html.erb
HWallenberg: yeah added to the existing titling implementation with error handling.

app/views/product_experience/enquiry_form/check_your_answers.html.erb
HWallenberg: The form field routes weren't explicitly constrained so the generated url was badly formed.  
Fixed.

app/views/product_experience/enquiry_form/_error.html.erb
```diff
@@ -0,0 +1,14 @@
+<div class="govuk-error-summary" data-module="govuk-error-summary">
```
willfish: This feels super generic and probably not specific to enquiry form submission?

Could it go here or somewhere shared: `app/views/application/_flashes.html.erb`?

app/views/product_experience/enquiry_form/_error.html.erb
```diff
@@ -0,0 +1,14 @@
...
+  <div role="alert">
```
willfish: We could do something like

```erb
<div class="flashes">
  <% if notice %>
    <div class="govuk-notification-banner govuk-notification-banner--success" role="alert" aria-labelledby="govuk-notification-banner-title" data-module="govuk-notification-banner">
      <div class="govuk-notification-banner__header">
        <h2 class="govuk-notification-banner__title" id="govuk-notification-banner-title">
          Success
        </h2>
      </div>
      <div class="govuk-notification-banner__content">
        <p><%= notice %></p>
      </div>
    </div>
  <% end %>

  <% if alert %>
    <div class="govuk-notification-banner govuk-notification-banner--error" role="alert" aria-labelledby="govuk-notification-banner-title" data-module="govuk-notification-banner">
      <div class="govuk-notification-banner__header">
        <h2 class="govuk-notification-banner__title" id="govuk-notification-banner-title">
          There is a problem
        </h2>
      </div>
      <div class="govuk-notification-banner__content">
        <p><%= alert %></p>
      </div>
    </div>
  <% end %>
</div>
```

I think errors typically tie into something form specific (e.g. a known field was validated and incorrect) versus flash errors which might be "we got to the end and someone unexpected happened"

willfish: It ends up producing something like 

Image

willfish: Image

app/views/product_experience/enquiry_form/_error.html.erb
HWallenberg: good idea done and integrated with myott too

app/views/product_experience/enquiry_form/_error.html.erb
HWallenberg: yes but on re-directs we need to use flash[:alert] otherwise the instance variable is swallowed up. so handling both.

willfish (approved):

PR #2505, opened 2025-08-06
BAU: Tidy the basic session UI

willfish:
### Jira link

BAU

### What?

I have a...

- Tidied the basic session UI

### Why?

I am doing this because...

- This is a arbitrarily messy and doesn't need inaccessible components on the page


HWallenberg (approved):
🚀

PR #2524, opened 2025-08-26
PRDEX-136 Enquiry Form Confirmation Page Use Session For Reference Number

HWallenberg:
### Jira link

[PRDEX-136](https://transformuk.atlassian.net/browse/PRDEX-136)

### What?

I have updated controller to store reference_number in session and use in confirmation page instead of inline parameter

### Why?

I am doing this because it prevents users from generating a confirmation page from erroneous URLs such as https://staging.trade-tariff.service.gov.uk/enquiry_form/confirmation?reference_number=foobar:

image

### DIFF
--- a/app/controllers/product_experience/enquiry_form_controller.rb
+++ b/app/controllers/product_experience/enquiry_form_controller.rb
@@ -64,7 +64,9 @@ module ProductExperience
         if response['resource_id']
           Rails.cache.delete(form_data['query']) if form_data['query'].present?
           session.delete(:enquiry_data)
-          session[:reference_number] = response['resource_id']
+          session[:product_experience_enquiry] = {
+            reference_number: response['resource_id'],
+          }
           redirect_to product_experience_enquiry_form_confirmation_path
         else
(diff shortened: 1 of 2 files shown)


app/controllers/product_experience/enquiry_form_controller.rb
```diff
@@ -64,7 +64,8 @@ def submit_form
         if response['resource_id']
           Rails.cache.delete(form_data['query']) if form_data['query'].present?
           session.delete(:enquiry_data)
-          redirect_to product_experience_enquiry_form_confirmation_path(reference_number: response['resource_id'])
+          session[:reference_number] = response['resource_id']
```
willfish: I'm wondering about future us and whether we want to namespace this under product_experience_enquiry or something?

```ruby
session[:product_experience_enquiry] = {
  reference_number: response['resource_id']
}
```

HWallenberg: Yes, I like that thanks 🚀 please can you re-review @willfish.

willfish (approved):

PR #2527, opened 2025-08-29
PRDEX-161 Enquiry Form Submission Token Validation Fix

HWallenberg:
### Jira link

[PRDEX-161](https://transformuk.atlassian.net/browse/PRDEX-161)

### What?

I have simplified the submission session token validation process.  Form can load without a submission token but will need a valid one to submit pages to the server or else be re-directed to the start.

### Why?

I am doing this because before the token in the param is not available during form load which was making the check irrelevant thus giving a false positive.

https://github.com/user-attachments/assets/60464048-4099-4a27-b510-f3e6c1c64e0c

### DIFF
--- a/spec/controllers/product_experience/enquiry_form_controller_spec.rb
+++ b/spec/controllers/product_experience/enquiry_form_controller_spec.rb
@@ -49,13 +49,6 @@ RSpec.describe ProductExperience::EnquiryFormController, type: :controller do
         expect(response).to render_template(:form)
       end
     end
-
-    # context 'with invalid token' do
-    #   it 'redirects back to the start' do
-    #     get :form, params: { field: 'full_name', submission_token: 'wrong-token' }
-    #     expect(response).to redirect_to(product_experience_enquiry_form_path)
-    #   end
-    # end
   end
 
   describe 'POST #submit' do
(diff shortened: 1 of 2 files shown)


spec/controllers/product_experience/enquiry_form_controller_spec.rb
```diff
@@ -50,12 +50,12 @@
       end
     end
 
-    context 'with invalid token' do
-      it 'redirects back to the start' do
-        get :form, params: { field: 'full_name', submission_token: 'wrong-token' }
-        expect(response).to redirect_to(product_experience_enquiry_form_path)
-      end
-    end
+    # context 'with invalid token' do
```
willfish: I guess we can delete this chipolata?

willfish (dismissed):

spec/controllers/product_experience/enquiry_form_controller_spec.rb
HWallenberg: doh

HWallenberg: I thought linter would've smashed that up tbh! 😆

willfish (approved):

PR #2570, opened 2025-10-08
HMRC-1577: Ignore known errors in NR

AigulDj:
### Jira link

[HMRC-1577](https://transformuk.atlassian.net/browse/HMRC-1577)

### What?

I have 
- Added several error classes to `error_collector.expected_classes` in `newrelic.yml` so that these known or harmless exceptions are treated as expected by the New Relic Ruby agent.

### Why?

I am doing this because
- This reduces false positives in error-rate alerts while still allowing the team to triage them in the Errors Inbox.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 ---
 repos:
   - repo: https://github.com/antonbabenko/pre-commit-terraform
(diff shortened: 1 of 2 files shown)


willfish (dismissed):
Let's give it a go

willfish (approved):

PR #2578, opened 2025-10-16
HMRC-1683: HMRC-1681: ActionView::Template::Error in frontend

rasikasri:
### Jira link

[HMRC-1683](https://transformuk.atlassian.net/browse/HMRC-1683)
[HMRC-1681](https://transformuk.atlassian.net/browse/HMRC-1681)

### What?

I have added filter to remove null value from the param list in heading and commodity page cache key
have added condition to ignore request with nested params to heading and commodity being cached 

### Why?

I am doing this because null values causing an error in sorting param list and caching thorrowly filtered results would get less hits

### DIFF
--- a/app/views/headings/show.html.erb
+++ b/app/views/headings/show.html.erb
@@ -1,4 +1,6 @@
-<%= cache ['headings#show', cache_key, params.values.sort.join('/')] do %>
+<% cacheable = params.values.compact.none? { |v| v.is_a?(ActionController::Parameters) } %>
+
+<% content = capture do %>
   <% content_for :title, goods_nomenclature_title(@heading) %>
 
   <% content_for :head do %>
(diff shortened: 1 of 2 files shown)


app/views/headings/show.html.erb
```diff
@@ -1,4 +1,6 @@
-<%= cache ['headings#show', cache_key, params.values.sort.join('/')] do %>
+<% cacheable = params.values.compact.none? { |v| v.is_a?(ActionController::Parameters) } %>
```
willfish, line 1: We don't really want to cache for every permutation of keys the user can supply as a query param/in the body. That's actually kind of mad since this means they could easily flood our cache with an unholy number of requests and bring down the service by flooding us with fragments

I think the targeted approach makes way more sense because the alternative is every sortable key can flood our cache with documents on each combination.

I also think we should be validating what goes into the cache key.

We have three variables that are interesting to the cache key currently:

```ruby
code = params[:id] # This will reduce the number of fragments to the number of things that exist as rows in the database for the given date
as_of = "#{params[:year]}-#{params[:month]}-#{params[:day]}" # Very permissive - we should probably validate this
country = params[:geographical_area_id) # Very permissive - we should probably validate this
```

Anyway the current approach I could bring down the site in < 5 minutes on my machine by knocking out the cache and then flooding the frontend with requests it had to personally service via the backend

willfish: Okay, reviewed this in a live environment and anyone flooding the caches gets rate limited pretty quick

We're also way overpowered in staging

willfish (approved):

PR #2596, opened 2025-11-10
MYOTT-206 Grouped measure changes

vertism:
### Jira link

[MYOTT-206](https://transformuk.atlassian.net/browse/MYOTT-206)

### What?

I have:
- added a GroupedMeasureChange model
- Extracted methods for getting excluded country list into a concern as it is needed by 2 models.

### Why?

I am doing this because it is needed for MyOTT

### DIFF
--- a/lib/api_entity.rb
+++ b/lib/api_entity.rb
@@ -125,8 +125,8 @@ private
       collection("#{collection_path}/search", opts)
     end
 
-    def collection(collection_path, opts = {})
-      resp = api.get(collection_path, opts)
+    def collection(collection_path, opts = {}, headers = {})
+      resp = api.get(collection_path, opts, headers)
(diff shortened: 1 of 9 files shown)


lib/api_entity.rb
```diff
@@ -125,8 +125,8 @@ def search(opts = {})
       collection("#{collection_path}/search", opts)
     end
 
-    def collection(collection_path, opts = {})
-      resp = api.get(collection_path, opts)
+    def collection(collection_path, opts = {}, headers = {})
+      resp = api.get(collection_path, opts, headers)
```
HWallenberg, line 129: Excellent, I've added this on my branch too! 👍

HWallenberg (approved):

PR #2601, opened 2025-11-13
MyOTT-117 Upload Commodity Codes

HWallenberg:
### Jira link

[MyOTT-117](https://transformuk.atlassian.net/browse/MyOTT-117)

### What?

I have 
- [x] Added `my_commodities` controller and views with `apientity` integration with myott `user`, `subscription` and `subscription_target` models.

### Why?

I am doing this because its required to upload a users commodity codes for their commodity watchlist

https://github.com/user-attachments/assets/d43ee41f-6b0c-4053-a3fa-75fcc3546039


app/controllers/myott/myott_controller.rb
```diff
@@ -24,5 +24,11 @@ def current_subscription
     def user_id_token
       cookies[:id_token]
     end
+
+    def get_subscription(subscription_type, token)
+      subscription_id = current_user[:subscriptions]&.find { |s| s['subscription_type'] == subscription_type }&.fetch('id', nil)
+      Subscription.find(subscription_id, token) if subscription_id
```
vertism: You can use user_id_token here rather than passing the token as an argument

app/controllers/myott/mycommodities_controller.rb
```diff
@@ -0,0 +1,144 @@
...
+      get_subscription_targets __method__.to_s
+    end
+
+    def index
+      my_commodities_subscription = get_subscription('my_commodities', cookies[:id_token])
+      redirect_to new_myott_mycommodity_path and return unless my_commodities_subscription
+
+      meta = my_commodities_subscription[:meta]
+
+      cache_session_commodity_codes({ 'active_commodity_codes' => meta['active'],
```
vertism: If we're caching this, does that mean we don't have to retrieve the subscription from backend every time?

vertism: Saying that, I think I might be missing the point of the caching. 
The codes are cached in line 38, and then we read from the cache in the next line.
Is this cache used anywhere else?

app/controllers/myott/mycommodities_controller.rb
```diff
@@ -0,0 +1,144 @@
...
+                          get_subscription('my_commodities', cookies[:id_token]).resource_id
+                        else
+                          subscription.resource_id
+                        end
+      Subscription.batch(subscription_id,
+                         targets: TariffJsonapiParser.new(commodity_codes.uniq).parse,
+                         subscription_type: 'my_commodities')
+    end
+
+    def extract_codes_from_file(file)
```
vertism, line 73: I think this should be extracted to a service object. Maybe other stuff like validation too.
I don't mind if this is done post merge.

app/controllers/myott/mycommodities_controller.rb
```diff
@@ -0,0 +1,144 @@
...
+      Rails.cache.write(redis_key, commodity_codes, expires_in: 1.hour)
+      session[:commodity_codes_key] = redis_key
+    end
+
+    def parse_numbers_only(string)
+      string.gsub(/[^0-9]/, '').strip
+    end
+
+    def candidate_commodity_code?(code)
+      code.present? && code.length.between?(8, 10)
```
vertism: Should we be handling codes that are < 10 digits?

vertism: Or should we in fact be handling all entries and displaying errors in the Upload Errors page?

app/controllers/myott/mycommodities_controller.rb
```diff
@@ -0,0 +1,144 @@
...
+      get_subscription_targets __method__.to_s
+    end
+
+    def expired
+      get_subscription_targets __method__.to_s
+    end
+
+    def invalid
+      get_subscription_targets __method__.to_s
+    end
```
vertism, line 26: Personally I would prefer to explicitly hardcode the strings here.

app/models/subscription.rb
```diff
@@ -3,5 +3,24 @@ class Subscription
...
+      data: {
+        attributes: attributes,
+      },
+    }
+    super(json_api_params, headers(token))
+  rescue Faraday::UnauthorizedError
+    nil
+  end
+
+  def self.headers(token)
```
vertism: This method is already in AuthenticatableApiEntity so can be removed from here

app/models/subscription_target.rb
```diff
@@ -0,0 +1,24 @@
+class SubscriptionTarget
+  include ApiEntity
```
vertism: This gives you the header method:
```suggestion 
  include AuthenticatableApiEntity
```

app/models/subscription_target.rb
```diff
@@ -0,0 +1,24 @@
...
+  attr_accessor :target_type,
+                :chapter_short_code,
+                :goods_nomenclature_item_id,
+                :hierarchical_description,
+                :meta
+
+  def self.all(token, target, page: 1, per_page: 10)
+    return nil if token.nil? && !Rails.env.development?
+
+    path = "/uk/user/subscription_targets/#{token}?filter[active_commodities_type]=#{target}&page=#{page}&per_page=#{per_page}"
```
vertism: You can set the path using 
`set_collection_path '/uk/user/subscription_targets'`

The params can be passed to collection in the 2nd argument (options)

Also, I think it would be better to pass params to this method and just pass it through. The backend should handle setting default values if necessary.

app/views/myott/mycommodities/index.html.erb
```diff
@@ -0,0 +1,51 @@
+<% myott_page_title "Upload commodities", error: @alert %>
+<div class="govuk-body myott-mycommodities">
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+      <h1 class="govuk-heading-l">Your commodity watchlist</h1>
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+          <h2>Total commodities uploaded: <%= number_with_delimiter(@total_commodity_codes, :delimeter => ',') %></h2>
```
vertism: Missing a class
```suggestion
          <h2 class="govuk-heading-m">Total commodities uploaded: <%= number_with_delimiter(@total_commodity_codes, :delimeter => ',') %></h2>
```

app/views/myott/mycommodities/index.html.erb
```diff
@@ -0,0 +1,51 @@
...
+          </div>
+        </div>
+        <div class="govuk-grid-column-one-third">
+          <div class="summary_container">
+            <strong>Expired commodity codes:</strong>
+            <br />
+             <% if @expired_commodity_codes.count == 0 %>
+              0
+            <% else %>
+              <%= link_to link_to  number_with_delimiter(@expired_commodity_codes.count, :delimeter => ','), expired_myott_mycommodities_path %>
```
vertism:
```suggestion
              <%= link_to number_with_delimiter(@expired_commodity_codes.count, :delimeter => ','), expired_myott_mycommodities_path %>
```

app/views/myott/mycommodities/index.html.erb
```diff
@@ -0,0 +1,51 @@
...
+          </div>
+        </div>
+        <div class="govuk-grid-column-one-third">
+          <div class="summary_container">
+            <strong>Upload errors:</strong>
+            <br />
+            <% if @invalid_commodity_codes.count == 0 %>
+              0
+            <% else %>
+              <%= link_to link_to  number_with_delimiter(@invalid_commodity_codes.count, :delimeter => ','), invalid_myott_mycommodities_path %>
```
vertism:
```suggestion
              <%= link_to number_with_delimiter(@invalid_commodity_codes.count, :delimeter => ','), invalid_myott_mycommodities_path %>
```

lib/api_entity.rb
```diff
@@ -93,6 +93,13 @@ def create!(params = {}, headers = {})
       new parse_jsonapi(response)
     end
 
+    def batch(params = {}, headers = {})
+      request = prepare_json_request(params, headers)
+      batch_path = "#{singular_path.sub(':id', headers[:authorization].split.last)}/batch"
```
vertism: This doesn't seem right. This is getting the user token, not the subscription ID.

app/controllers/myott/mycommodities_controller.rb
```diff
@@ -0,0 +1,144 @@
...
+      subscription = get_subscription('my_commodities', cookies[:id_token])
+
+      subscription_id = if subscription.nil? && User.update(cookies[:id_token], my_commodities_subscription: 'true')
+                          get_subscription('my_commodities', cookies[:id_token]).resource_id
+                        else
+                          subscription.resource_id
+                        end
+      Subscription.batch(subscription_id,
+                         targets: TariffJsonapiParser.new(commodity_codes.uniq).parse,
+                         subscription_type: 'my_commodities')
```
vertism, line 70: I don't think subscription_type should be needed because that is known from the subscription.
This might need a change to backend, so could do after merging.

.env.development
```diff
@@ -1,5 +1,5 @@
 ALLOW_SEARCH=true
-API_SERVICE_BACKEND_URL_OPTIONS={"uk":"http://localhost:3000/uk/api","xi":"http://localhost:3000/xi/api"}
+API_SERVICE_BACKEND_URL_OPTIONS={"uk":"http://localhost:3000/api/v2","xi":"http://localhost:3000/api/v2"}
```
vertism: I don't think this should be changed.
You can add to .env.development.local for yourself if you need to change it for some reason

app/controllers/myott/mycommodities_controller.rb
HWallenberg: Oh I see what you mean, it's caching every time when the page loads so if someone clicks < back it's getting the subscription again and then caching again which seems a little unnecessary.  

This is a hangover from when I was passing the commodity codes around instead of relying on the new `Subscription` and `SubscriptionTarget` APIs to get the users CCs.

So I think that there should be a check to see if a subscription has been cached and if not load one & cache it?  That way index will load quicker?

app/controllers/myott/mycommodities_controller.rb
HWallenberg: Yes the `batch` API requires the `subscription_type` but what you're saying is that the `subscription_id` will give us the `my_commodities` subscription anyway.
So backend change to update that API post merge if I understand correctly.

app/controllers/myott/mycommodities_controller.rb
HWallenberg: yes agree ticket raised. https://transformuk.atlassian.net/browse/MYOTT-244

app/controllers/myott/mycommodities_controller.rb
HWallenberg: Yes I think we were just ignoring anything other than a 8 or 10 digit CC at one point during prototyping.
This code isn't needed as these should be in the errors page if the number isn't a recognised CC.

app/controllers/myott/mycommodities_controller.rb
HWallenberg: Jira created to updated the front end and backend post merge. https://transformuk.atlassian.net/browse/MYOTT-245

app/models/subscription_target.rb
```diff
@@ -0,0 +1,18 @@
...
+  attr_accessor :target_type,
+                :chapter_short_code,
+                :goods_nomenclature_item_id,
+                :hierarchical_description,
+                :meta
+
+  def self.all(token, params)
+    return nil if token.nil? && !Rails.env.development?
+
+    set_collection_path "/uk/user/subscription_targets/#{token}"
```
vertism: I think this could be a problem as set_collection_path is a class method. 
Better to call the collection method from here with the path you've created.

app/controllers/myott/mycommodities_controller.rb
```diff
@@ -0,0 +1,161 @@
...
+    def expired
+      get_subscription_targets 'expired'
+    end
+
+    def invalid
+      get_subscription_targets 'invalid'
+    end
+
+    def index
+      meta = Rails.cache.read(session[:subscription_meta_key]) if session[:subscription_meta_key].present?
```
vertism: I'm finding the caching hard to follow. My questions would be:
1. Is it necessary, is there any improvement we can make to the backend
2. If yes, then can it be moved to the model or a service to keep the controller cleaner?

HWallenberg: I believe that the caching is probably a case of engineering out a problem that doesn't exist.  I've removed it and if we get issues with load times then we can revisit this.

vertism (approved):
LGTM

PR #2608, opened 2025-11-18
MyOTT-244 Implement Commodity Codes Extraction Service

HWallenberg:
### Jira link

[MyOTT-244](https://transformuk.atlassian.net/browse/MyOTT-244)

### What?

I have refactored `extract_codes_from_file` into a service

### Why?

I am doing this because the function should be separated from the `my_commodities_controller`.


vertism (approved):
Nice job!

PR #2610, opened 2025-11-19
Extract commodity counts logic and partial

vertism:
### What?

- Moves loading of subscription into a single method
- Moves view into a partial
- Uses 1 instance variable to hold all counts of commodities in the various buckets

### Why?

Simplifying the index action before adding more components to this page.


HWallenberg (approved):
🙌

PR #2612, opened 2025-11-20
MYOTT-206 Add grouped measure changes

vertism:
### Jira link

[MYOTT-206](https://transformuk.atlassian.net/browse/MYOTT-206)

### What?

Added grouped measure changes for a single day to the index page for MyOTT


HWallenberg (approved):
nice one! 🚀

PR #2616, opened 2025-11-21
HMRC-1769: NoMethodError in frontend

rasikasri:
### Jira link

[HMRC-1769](https://transformuk.atlassian.net/browse/HMRC-1769)

### What?

I have added default import_date value to duty calculator

### Why?

I am doing this because, invalid session cause NoMethodError in import_date

### DIFF
--- a/app/models/duty_calculator/steps/import_date.rb
+++ b/app/models/duty_calculator/steps/import_date.rb
@@ -30,7 +30,7 @@ module DutyCalculator
       end
 
       def import_date
-        super || user_session.import_date
+        super || user_session.import_date || Time.zone.now.strftime('%Y-%m-%d')
       end
 
       def save!


willfish (approved):
I don't know how this is possible:

```ruby
      def initial_date_params
        {
          'import_date(3i)' => day.to_s,
          'import_date(2i)' => month.to_s,
          'import_date(1i)' => year.to_s,
        }
      end

      def day
        params[:day] || now.day
      end

      def month
        params[:month] || now.month
      end

      def year
        params[:year] || now.year
      end

      def now
        @now ||= Time.zone.now
      end
```

willfish:
So the controller will always pass the date unless for whatever reason the date that's getting parsed can't be processed like so:

```ruby

      def permitted_params
        params.require(:duty_calculator_steps_import_date).permit(
          :'import_date(3i)',
          :'import_date(2i)',
          :'import_date(1i)',
        )
      end

```

rasikasri:
Yeah I went through this code, and cannot explain the issue from the code. But when I copy the url can send it from different window, I managed to recreate the issue. It has to be something with messed up session. Just added the safety line to get rid of the error message. But I agree with you, don't know how it happened. 

Screenshot 2025-11-21 at 17 50 53

PR #2617, opened 2025-11-21
MyOTT-200 Refactor User API Subscriptions

HWallenberg:
### Jira link

[MyOTT-200](https://transformuk.atlassian.net/browse/MyOTT-200)

### What?

I have refactored `current_subscription` into `myott_controller` so `subscription_controller` and `my_commodities_controller` can use the shared function

### Why?

I am doing this because both subscription types need to be managed the same way and the backend has been changed for this.


app/controllers/myott/myott_controller.rb
```diff
@@ -17,18 +17,20 @@ def current_user
...
     helper_method :current_user
 
-    def current_subscription
-      @current_subscription ||= Subscription.find(params[:id], user_id_token)
+    def current_subscription(subscription_type)
+      @current_subscriptions ||= {}
+      @current_subscriptions[subscription_type] ||= get_subscription(subscription_type)
     end
 
+    helper_method :current_subscription
```
vertism: Is this needed?

HWallenberg: no, removed.

vertism (approved):

PR #2624, opened 2025-11-25
MyOTT-255 Iterate List Views

HWallenberg:
### Jira link

[MyOTT-255](https://transformuk.atlassian.net/browse/MyOTT-255)

### What?

I have added partials for each list category.  Helper method for page title and `validity_end_date` to `subscriptiontarget` model

### Why?

I am doing this because it's need for presentation of each category.

Invalid
image

Expired
image

Active
image


app/views/myott/mycommodities/_active.html.erb
```diff
@@ -0,0 +1,21 @@
+<table class="govuk-table">
+  <thead class="govuk-table__head">
+    <tr class="govuk-table__row">
+      <th class="govuk-table__header">Commodity</th>
+      <th class="govuk-table__header">Classification</th>
+    </tr>
+  </thead>
+  <tbody class="govuk-table__body">
+      <% @commodities.each do |commodity| %>
```
vertism: Minor niggle - you can use commodities here (rather than @commodities) as you are passing that variable into the partial. Same with the other partials.
Same with @total_commodities_count

app/views/myott/mycommodities/list.html.erb
```diff
@@ -5,51 +5,8 @@
...
-              </td>
-               <td class="govuk-table__cell">
-                <%= commodity.try(:classification_description) %>
-              </td>
-            </tr>
-          <% end %>
-        </tbody>
-      </table>
-      <%= paginate @commodities %>
+      <h1 class="govuk-heading-xl"><%= my_commodities_page_heading(@category, @total_commodities_count) %></h1>
```
vertism: Can you move this into the partial instead of needing a helper method?

HWallenberg: yeah seems like overkill 👍

app/views/myott/mycommodities/_active.html.erb
```diff
@@ -0,0 +1,22 @@
+<h1 class="govuk-heading-xl">Active commodities: <%= number_with_delimiter(total_commodities_count, delimiter: ',') %></h1>
```
vertism: I don't think you need to specify the delimiter as comma is the default. 
Also, as commodities is being passed to this partial, could you use commodities.total_count here rather than having to create an instance variable?

Same applies to the other partials

app/controllers/myott/mycommodities_controller.rb
vertism, line 1: As most of the content for active, invalid and expired is in the partial, what do you think about removing the list view completely and just having the 3 views?
You could change get_subscription_targets to return the result of SubscriptionTarget.all and assign to @commodities in the action
e.g.
```
def active
  @commodities = get_subscription_targets('active')
end
```

HWallenberg: 100% agree

vertism (approved):
👍

PR #2632, opened 2025-12-01
MYOTT-288 Add main watch lists page

vertism:
### Jira link

[MYOTT-288](https://transformuk.atlassian.net/browse/MYOTT-288)

### What?

Added a new watch list index page that serves as an entry point for MyOTT once logged in.
A new stop_press controller has been created specifically for pages that relate to the stop press subscription.

### Why?

To enable a user to access both the subscriptions they might have.


app/controllers/myott/subscriptions_controller.rb
```diff
@@ -8,51 +8,11 @@ def invalid
...
-      session[:chapter_ids] = session_chapters.all_chapters.map(&:to_param) if session[:chapter_ids].blank?
-
-      @selected_chapters_heading = session_chapters.selected_chapters_heading
-      @selected_sections_chapters = session_chapters.selected_sections_chapters
-    end
-
-    def session_chapters
-      @session_chapters ||= SessionChaptersDecorator.new(session)
+      @stop_press = current_subscription('stop_press')
+      @my_commodities = current_subscription('my_commodities')
```
HWallenberg, line 15: nice! 👍

HWallenberg:
Thanks @vertism looks good, one small suggestion after playing with the 'journey' 
It may not be on the design but it feels like commodity watch list and stop press watch list both now need  a `< Back` button like the other pages to get back to the new landing page.

HWallenberg (approved):

PR #2637, opened 2025-12-03
MyOTT-263 & MyOTT-264 Iterate Upload CCs journey

HWallenberg:
### Jira link

[MyOTT-263](https://transformuk.atlassian.net/browse/MyOTT-263)
[MyOTT-264](https://transformuk.atlassian.net/browse/MyOTT-264)

### What?

I have added dynamic `confirmation` view and dynamic `new` view based on user subscription status

### Why?

I am doing this because it's part of the journey design iteration.

**new subscriber upload page**
image

**new subscriber confirmation page**
image

**existing subscriber upload page**
image

**existing subscriber confirmation page**
image

### DIFF
--- a/app/controllers/myott/mycommodities_controller.rb
+++ b/app/controllers/myott/mycommodities_controller.rb
@@ -1,6 +1,9 @@
 module Myott
   class MycommoditiesController < MyottController
-    before_action :authenticate, only: %i[new create index confirmation]
+    before_action :authenticate
+    before_action except: %i[new create] do
+      redirect_to myott_path unless current_subscription('my_commodities')
+    end
 
     def new; end
 
@@ -17,8 +20,6 @@ module Myott
     end
 
     def index
-      redirect_to new_myott_mycommodity_path and return unless current_subscription('my_commodities')
-
       @meta = metadata_from_subscription
       @grouped_measure_changes = TariffChanges::GroupedMeasureChange.all(user_id_token, { as_of: as_of.strftime('%Y-%m-%d') })
       @commodity_changes = TariffChanges::CommodityChange.all(user_id_token, { as_of: as_of.strftime('%Y-%m-%d') })
@@ -39,7 +40,6 @@ module Myott
 
     def confirmation
       @new_subscriber = params[:new_subscriber] == 'true'
-      redirect_to myott_path unless current_subscription('my_commodities')
     end
 
     private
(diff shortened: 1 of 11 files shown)


app/controllers/myott/mycommodities_controller.rb
```diff
@@ -1,6 +1,6 @@
 module Myott
   class MycommoditiesController < MyottController
-    before_action :authenticate, only: %i[new create index]
+    before_action :authenticate, only: %i[new create index confirmation]
```
vertism: Is there a reason not to be authenticating on all actions?

app/controllers/myott/mycommodities_controller.rb
```diff
@@ -33,7 +34,12 @@ def create
       end
 
       update_user_commodity_codes(result.codes)
-      redirect_to myott_mycommodities_path
+      redirect_to confirmation_myott_mycommodities_path params: { new_subscriber: new_subscriber } and return
+    end
+
+    def confirmation
+      @new_subscriber = params[:new_subscriber] == 'true'
+      redirect_to myott_path unless current_subscription('my_commodities')
```
vertism: I think this redirect should be the same as on index when there is no subscription, and it should probably be a before_action on all actions apart from new.

app/controllers/myott/mycommodities_controller.rb
HWallenberg: nope, was being over cautious during dev

app/controllers/myott/mycommodities_controller.rb
HWallenberg: agree, moved to a `before_action` updated specs for controllers that inherit from this controller and link for a new `my_commodities` subscriber.

app/controllers/myott/mycommodities_controller.rb
```diff
@@ -1,6 +1,9 @@
 module Myott
   class MycommoditiesController < MyottController
-    before_action :authenticate, only: %i[new create index]
+    before_action :authenticate
+    before_action except: %i[new create] do
+      redirect_to myott_path unless current_subscription('my_commodities')
```
vertism: Should this be
```suggestion
      redirect_to new_myott_mycommodity_path unless current_subscription('my_commodities')
```
I think it makes sense to take them to the upload page if they are in this controller without a subscription

HWallenberg: yes 👍

vertism (approved):

PR #2645, opened 2025-12-11
MYOTT-299 Remove bullet from list

vertism:
### Jira link

[MYOTT-299](https://transformuk.atlassian.net/browse/MYOTT-299)

### What?

I have removed a bullet from the list that should not be there.
Also adds govuk-list class and changes h3 to h2

Screenshot 2025-12-11 at 12 37 54

### Why?

To remove incorrect content and improve accessibility


HWallenberg (approved):
🚀

PR #2657, opened 2025-12-29
HMRC-1819: Fixes bug in meursing results

willfish:
### Jira link

[HMRC-1819](https://transformuk.atlassian.net/browse/HMRC-1819)

### What?

I have a...

- Moved to singular results for the meursing path and jettisoned the :id param on the show action

### Why?

I am doing this because...

- This enables us to just pass a goods nomenclature code to the params helper on clearing that isn't interpreted as an id field
- It makes more sense since we're not showing multiple results - just one


shjohnson (approved):
Makes sense

PR #2664, opened 2025-12-30
HMRC-1844: Adds rack-timeout with a 15 second timeout

willfish:
### Jira link

[HMRC-1844](https://transformuk.atlassian.net/browse/HMRC-1844)

### What?

- Added the `rack-timeout` gem and configured it with a 15-second service timeout.

### Why?

- To ensure requests don't hang indefinitely and to enforce a strict timeout policy for the application.


shjohnson (dismissed):
Makes sense 🏅

PR #2665, opened 2026-01-05
MyOTT-291 Integrate Unsubscribe Journey for my_commodities and stop_press

HWallenberg:
### Jira link

[MyOTT-291](https://transformuk.atlassian.net/browse/MyOTT-291)

### What?

I have added unsubscribe journey for `my_commodities` and integrated `stop_press` unsubscribe

Added `def get_subscription_type_by_id(subscription_id)` as user's subscription UUID is known at this point of the journey.

### Why?

I am doing this because the app must allow a user to unsubscribe from either a stop press or a Commodity Watch List subscription independently. 

https://github.com/user-attachments/assets/e7070eed-598b-4208-a13e-40ebf0530552

### DIFF
--- a/app/controllers/myott/myott_controller.rb
+++ b/app/controllers/myott/myott_controller.rb
@@ -31,10 +31,5 @@ module Myott
       subscription_id = current_user[:subscriptions]&.find { |s| s['subscription_type'] == subscription_type && s['active'] }&.fetch('id', nil)
       Subscription.find(subscription_id, user_id_token) if subscription_id
     end
-
-    def get_subscription_type_by_id(subscription_id)
-      subscription = Subscription.find(subscription_id, user_id_token)
-      subscription['subscription_type'] if subscription
-    end
   end
 end
--- a/app/controllers/myott/myott_controller.rb
+++ b/app/controllers/myott/myott_controller.rb
@@ -31,5 +31,10 @@ module Myott
       subscription_id = current_user[:subscriptions]&.find { |s| s['subscription_type'] == subscription_type && s['active'] }&.fetch('id', nil)
       Subscription.find(subscription_id, user_id_token) if subscription_id
     end
+
+    def get_subscription_type_by_id(subscription_id)
+      subscription = Subscription.find(subscription_id, user_id_token)
+      subscription['subscription_type'] if subscription
+    end
   end
 end
--- a/app/controllers/myott/unsubscribes_controller.rb
+++ b/app/controllers/myott/unsubscribes_controller.rb
@@ -4,11 +4,6 @@ module Myott
 
     before_action :authentication, except: :confirmation
 
-    SUBSCRIPTION_TYPES = {
-      stop_press: 'stop_press',
-      my_commodities: 'my_commodities',
-    }.freeze
-
     def show
       render subscription_type
     end
@@ -37,12 +32,8 @@ module Myott
       @subscription_type ||= Subscription.find(params[:id], user_id_token)&.subscription_type_name
     end
 
-    def stop_press_subscription?
-      subscription_type == SUBSCRIPTION_TYPES[:stop_press]
-    end
-
     def my_commodities_subscription?
-      subscription_type == SUBSCRIPTION_TYPES[:my_commodities]
+      subscription_type == Subscription::SUBSCRIPTION_TYPES[:my_commodities]
     end
 
     def delete_cookie
@@ -85,13 +76,13 @@ module Myott
       errors = unsubscribe_error_messages
       @alert = errors[:confirmation]
       flash.now[:select_error] = @alert
-      render template: "myott/unsubscribes/#{subscription_type}"
+      render subscription_type
     end
 
     def show_deletion_error
       errors = unsubscribe_error_messages
       @alert = errors[:deletion]
-      render template: "myott/unsubscribes/#{subscription_type}"
+      render subscription_type
     end
   end
 end
--- a/app/controllers/myott/unsubscribes_controller.rb
+++ b/app/controllers/myott/unsubscribes_controller.rb
@@ -34,7 +34,7 @@ module Myott
     end
 
     def subscription_type
-      @subscription_type ||= get_subscription_type_by_id(params[:id])['name']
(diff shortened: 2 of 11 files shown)


app/controllers/myott/unsubscribes_controller.rb
```diff
@@ -1,32 +1,97 @@
...
     before_action :authentication, except: :confirmation
 
-    def show; end
+    SUBSCRIPTION_TYPES = {
+      stop_press: 'stop_press',
+      my_commodities: 'my_commodities',
+    }.freeze
+
+    def show
+      render template: "myott/unsubscribes/#{subscription_type}"
```
vertism: Will this work?
```suggestion
      render subscription_type
```

app/controllers/myott/unsubscribes_controller.rb
```diff
@@ -1,32 +1,97 @@
...
     def confirmation
-      domain = ".#{request.host.sub(/^www\./, '')}"
-      cookies.delete(:id_token, domain:)
-      @header = 'You have unsubscribed'
-      @message = 'You will no longer receive any Stop Press emails from the UK Trade Tariff Service.'
+      @subscription_type = params[:subscription_type]
+      content = unsubscribe_confirmation_content(@subscription_type)
+      @header = content[:header]
+      @message = content[:message]
+      delete_cookie_for_stop_press if params[:subscription_type] == 'stop_press'
```
vertism: Should deleting the cookie be dependant on whether they have any subscription?

app/controllers/myott/unsubscribes_controller.rb
```diff
@@ -1,32 +1,97 @@
...
 
     def authentication
-      if params[:id].nil? || current_subscription('stop_press').nil?
+      if params[:id].nil? || subscription_type.nil?
         redirect_to myott_start_path
       end
     end
+
+    def subscription_type
+      @subscription_type ||= get_subscription_type_by_id(params[:id])['name']
```
vertism: Doesn't `get_subscription_type_by_id` return a string?

app/controllers/myott/myott_controller.rb
```diff
@@ -31,5 +31,10 @@ def get_subscription(subscription_type)
       subscription_id = current_user[:subscriptions]&.find { |s| s['subscription_type'] == subscription_type && s['active'] }&.fetch('id', nil)
       Subscription.find(subscription_id, user_id_token) if subscription_id
     end
+
+    def get_subscription_type_by_id(subscription_id)
```
vertism: Does this have a test?

app/controllers/myott/unsubscribes_controller.rb
HWallenberg: Yes!

app/controllers/myott/unsubscribes_controller.rb
HWallenberg: no it returns a hash:

```
get_subscription_type_by_id(params[:id])

=> {"resource_type" => "subscription_type", "resource_id" => "1", "name" => "stop_press"}
```

`subscription_type` is not a hydrated object.  I've created a Jira to do this as low priority [MYOTT-354](https://transformuk.atlassian.net/browse/MYOTT-354).

app/controllers/myott/unsubscribes_controller.rb
HWallenberg: no the cookie is relevant to the subscription that the user is unsubscribing from. 👍

app/controllers/myott/myott_controller.rb
HWallenberg: it does now! 🙂

app/controllers/myott/unsubscribes_controller.rb
vertism: Maybe change 'get_subscription_type_by_id' to return the name as that's what you need here?

app/controllers/myott/myott_controller.rb
```diff
@@ -31,5 +31,10 @@ def get_subscription(subscription_type)
...
+      subscription = Subscription.find(subscription_id, user_id_token)
+      subscription['subscription_type'] if subscription
```
vertism: This should work as you've added it to attr_accessor
```suggestion
      subscription&.subscription_type
```

HWallenberg: @vertism 

I've removed `get_subscription_type_by_id` from `myott_controller.rb` .

I've added a method `subscription_type_name` to the `subscription` model for single responsibility. 

I've updated `unsubscribes_controller.rb>subscription_type` to use the new  method.

What do you think?

app/controllers/myott/unsubscribes_controller.rb
HWallenberg: see other [comment](https://github.com/trade-tariff/trade-tariff-frontend/pull/2665#discussion_r2664534082)

app/controllers/myott/unsubscribes_controller.rb
```diff
@@ -1,32 +1,97 @@
...
   class UnsubscribesController < MyottController
+    include UnsubscribesHelper
+
     before_action :authentication, except: :confirmation
 
-    def show; end
+    SUBSCRIPTION_TYPES = {
+      stop_press: 'stop_press',
+      my_commodities: 'my_commodities',
+    }.freeze
```
vertism: What do you think about moving this to the Subscription model?

app/controllers/myott/unsubscribes_controller.rb
```diff
@@ -1,32 +1,97 @@
...
+
+    def user_confirmed?
+      params[:decision] == 'true'
+    end
+
+    def show_confirmation_error
+      errors = unsubscribe_error_messages
+      @alert = errors[:confirmation]
+      flash.now[:select_error] = @alert
+      render template: "myott/unsubscribes/#{subscription_type}"
```
vertism:
```suggestion
      render subscription_type
```
Same for show_deletion_error

app/controllers/myott/myott_controller.rb
vertism: Looks good!

vertism (approved):

PR #2670, opened 2026-01-06
MYOTT-356 Add cancel link to upload

vertism:
### Jira link

[MYOTT-356](https://transformuk.atlassian.net/browse/MYOTT-356)

### What?

Adds a cancel link that allows a user to return to the previous page

Screenshot 2026-01-06 at 14 24 21

When the user does not have a subscription, cancel takes them back to the watch lists dashboard, when they do it takes them to commodities watch list dashboard.

### Why?

To help with navigation


HWallenberg (approved):
nice! 🙂

PR #2674, opened 2026-01-08
MYOTT-378 Store URL whilst user authenticates

vertism:
### Jira link

[MYOTT-378](https://transformuk.atlassian.net/browse/MYOTT-378)

### What?

Currently when a user is not authenticated they are redirected to the Identity service. After logging in they are always returned to the watch lists dashboard page, not the page they were trying to access.

Now their intended URL is stored in the session so they are redirected to that page after logging in.

### Why?

This allows links to specific pages to be added to notification emails, and for these links to work when the user needs to login to the service.

### DIFF
--- a/app/controllers/myott/myott_controller.rb
+++ b/app/controllers/myott/myott_controller.rb
@@ -9,7 +9,10 @@ module Myott
 
     def authenticate
       if current_user.nil?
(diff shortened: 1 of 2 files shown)


HWallenberg (approved):
cool!

PR #2678, opened 2026-01-08
HMRC-1870: Handle authentication failures and clear cookies in frontend

willfish:
# Jira link

[HMRC-1870](https://transformuk.atlassian.net/browse/HMRC-1870)

## What?

I have:

- [x] Created `AuthenticationError` class with `should_clear_cookies?` logic
- [x] Updated `AuthenticatableApiEntity` to parse JSON:API error responses from backend
- [x] Updated `MyottController` to handle `AuthenticationError` and selectively clear cookies
- [x] Added `identity_cookie_domain` helper to `TradeTariffFrontend` (derived from `base_domain`)

## Why?

I am doing this because:

- When the backend returns specific authentication error codes, the frontend needs to take appropriate action
- For `not_in_group`, `invalid_token`, and `missing_jwks_keys` errors, cookies must be cleared to force clean re-authentication
- For `expired` token errors, cookies should be preserved so the identity service can refresh them automatically
- Without clearing cookies for invalid authentication states, users get stuck in redirect loops between services
- This completes the authentication failure handling flow started in the backend

The frontend now:
1. Catches `Faraday::UnauthorizedError` from API calls
2. Parses the JSON:API error response to extract the error code
3. Raises `AuthenticationError` with the specific reason
4. Clears `id_token` and `refresh_token` cookies if needed
5. Redirects to identity service for re-authentication

## Related PRs

- Backend: https://github.com/trade-tariff/trade-tariff-backend/pull/2691

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,5 @@ yarn-debug.log*
 /test-results
 /playwright-report/
 /dist
+
+.claude/
--- a/app/controllers/myott/myott_controller.rb
+++ b/app/controllers/myott/myott_controller.rb
@@ -5,6 +5,8 @@ module Myott
                   :disable_switch_service_banner,
                   :disable_last_updated_footnote
 
+    rescue_from AuthenticationError, with: :handle_authentication_error
+
   private
 
     def authenticate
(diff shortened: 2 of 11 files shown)


montychristo (dismissed):

vertism (commented):
This looks great! I think it could benefit from some tests.
Also, the Unsubscribes controller has a delete_cookie method. Can you update that to use the clear_authentication_cookies you have added?

vertism (approved):
LGTM

PR #2697, opened 2026-01-20
Concurrency control on deploys

vertism:
### What?

Add concurrency control to deploys. 

### Why?

Currently there is no guarantee what order deploys will be made, so where 2 PRs are merged around the same time, a race condition means the 1st PR may get deployed 2nd so the changes in the 2nd PR are no longer deployed.

### DIFF
--- a/.github/workflows/deploy-to-development.yml
+++ b/.github/workflows/deploy-to-development.yml
@@ -10,6 +10,10 @@ on:
       - labeled
       - unlabeled
 
+concurrency:
+  group: deploy-development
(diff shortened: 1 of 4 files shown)


amberstarlight (approved):

.github/workflows/deploy-to-development.yml
```diff
@@ -10,6 +10,10 @@ on:
       - labeled
       - unlabeled
 
+concurrency:
```
willfish, line 13: This is already controlled by terraform. You should only need this in development

PR #2753, opened 2026-02-11
Handle CSV files more liberally

vertism:
### What?

The CSV library is strict about the file having correct line endings. We only need to extract the first value in each line so this can be done simply without the library.


HWallenberg (approved):
🚀

PR #2758, opened 2026-02-12
MYOTT-482 Refactor preferences form

vertism:
### Jira link

[MYOTT-482](https://transformuk.atlassian.net/browse/MYOTT-482)

### What?

Refactors the preference selection form to use GOVUKDesignSystemFormBuilder and a form object.

### Why?

- Simplifies the frontend code
- Moves the validation logic out of the controller
- Moves the submit button lower for better accessibility

### DIFF
--- a/app/forms/myott/stop_press_preference_form.rb
+++ b/app/forms/myott/stop_press_preference_form.rb
@@ -1,0 +1,20 @@
+class Myott::StopPressPreferenceForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  SELECT_CHAPTERS = 'selectChapters'.freeze
+  ALL_CHAPTERS = 'allChapters'.freeze
+
+  attribute :preference, :string
+
+  validates :preference, presence: { message: 'Select a subscription preference to continue' },
+                         inclusion: { in: [SELECT_CHAPTERS, ALL_CHAPTERS], message: 'Select a subscription preference to continue' }
+
+  def select_chapters?
+    valid? && preference == SELECT_CHAPTERS
+  end
+
+  def all_chapters?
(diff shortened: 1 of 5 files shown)


willfish (dismissed):
Sexy

app/forms/myott/stop_press_preference_form.rb
```diff
@@ -0,0 +1,20 @@
+class Myott::StopPressPreferenceForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  SELECT_CHAPTERS = 'selectChapters'.freeze
+  ALL_CHAPTERS = 'allChapters'.freeze
+
+  attribute :preference, :string
+
+  validates :preference, presence: { message: 'Select a subscription preference to continue' },
```
HWallenberg, line 10: I'm not sure how we should be handling invalid errors.

If you look at `app/views/myott/unsubscribes/my_commodities.html.erb` 

The flash error is added by `app/controllers/myott/unsubscribes_controller.rb` with the `@div_id` used with `app/views/shared/_flashes.html.erb`

I understand that `GOVUKDesignSystemFormBuilder` validates differently.  So now we have two ways of validating and displaying an error.

However we decide to do it going forward UR said that the link in the error should be clickable so it takes you to the part of the page like in my commodities unsubscribe below 👇 

https://github.com/user-attachments/assets/c576b40c-13cd-4354-8e22-0f41b9dfd407

Hope this makes sense.

vertism: That's a great spot. It does have a link but it's not being set to the right ID for some reason. I've manually added in the correct ID to get it working.

HWallenberg (approved):

PR #2759, opened 2026-02-12
Revert "HMRC-1967: Upgrade Ruby from 3.4.7 to 4.0.1"

willfish:
Reverts trade-tariff/trade-tariff-frontend#2757


vertism (approved):

willfish:
Okay, this isn't needed

PR #2760, opened 2026-02-12
HMRC-1958: Different excluded countries showing on OTT - Potential Caching issue

rasikasri:
### Jira link

[HMRC-1958](https://transformuk.atlassian.net/browse/HMRC-1958)

### What?

I have added new cache key to european union geo area description.
Added logs to investigate how older description is fetched

### Why?

I am doing this because...
OTT is showing two different excluded countries description for EU

### DIFF
--- a/app/models/geographical_area.rb
+++ b/app/models/geographical_area.rb
@@ -23,7 +23,19 @@ class GeographicalArea
 
   class << self
     def european_union
-      @european_union ||= find(EUROPEAN_UNION_ID)
+      @european_union ||= Rails.cache.fetch("european_union", EUROPEAN_UNION_ID) do
+        find(EUROPEAN_UNION_ID).tap do |eu|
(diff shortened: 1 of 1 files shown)


app/models/geographical_area.rb
```diff
@@ -6,6 +6,7 @@ class GeographicalArea
   ERGA_OMNES = '1011'.freeze
 
   include ApiEntity
+  extend CacheHelper
```
willfish, line 9: Does this constant exist?

rasikasri: do you mean the values compose the cache key?
` [TradeTariffFrontend::ServiceChooser.cache_prefix, @tariff_last_updated, TradeTariffFrontend.revision].join('/')
`

willfish: Ah okay

willfish (approved):

PR #2765, opened 2026-02-13
HMRC-1937: Missing Application Security Headers

rasikasri:
### Jira link

[HMRC-1937](https://transformuk.atlassian.net/browse/HMRC-1937)

### What?

I have added/removed/altered:

- [ ] Added security response headers 
- [ ] Added report only content security policy

### Why?

I am doing this because:

- PEN test report recomended web server to return thesecurity-related HTTP response headers

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -35,7 +35,12 @@ Rails.application.configure do
   config.assume_ssl = ENV.fetch('RAILS_ASSUME_SSL', 'true') == 'true'
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  config.force_ssl = false
+  config.force_ssl = true
+
+  # HSTS should only be set over HTTPS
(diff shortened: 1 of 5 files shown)


config/environments/production.rb
```diff
@@ -35,7 +35,12 @@
   config.assume_ssl = ENV.fetch('RAILS_ASSUME_SSL', 'true') == 'true'
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  config.force_ssl = false
+  config.force_ssl = true
+
+  # HSTS should only be set over HTTPS
```
willfish: As I understand it this is what you get for free with force ssl

rasikasri: Yes, that's right

rasikasri: Is it still worth to leave it to change default values

willfish (approved):

PR #2777, opened 2026-02-17
MyOTT-354 Refactor SubscriptionType 

HWallenberg:
### Jira link

[MyOTT-354](https://transformuk.atlassian.net/browse/MyOTT-354)

### What?

I have added `SubscriptionType` model to host delegated methods and refactored `Subscription` and `Unsubscribe` to use delegated methods.

### Why?

I am doing this to make `Subscription` and `SubscriptionType` cleaner and more object-oriented.

### DIFF
--- a/app/controllers/myott/unsubscribes_controller.rb
+++ b/app/controllers/myott/unsubscribes_controller.rb
@@ -5,8 +5,8 @@ module Myott
     skip_before_action :authenticate, only: :confirmation
 
     def show
-      @form = Myott::UnsubscribeMyCommoditiesForm.new if my_commodities_subscription?
-      render subscription_type
+      @form = Myott::UnsubscribeMyCommoditiesForm.new if subscription_type.my_commodities_subscription?
+      render subscription_type&.name
     end
 
     def destroy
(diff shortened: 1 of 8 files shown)


app/controllers/myott/unsubscribes_controller.rb
vertism, line 1: I like the creation of the SubscriptionType. What do you think about loading a subscription object in this controller and then delegating the methods needed from SubscriptionType. e.g. allow the controller to call `subscription.my_commodities_subscription?` and keep the `subscription_type_name` method (but have it call the new class.

app/controllers/myott/unsubscribes_controller.rb
```diff
@@ -25,21 +25,17 @@ def confirmation
   private
 
     def authenticate
-      if params[:id].nil? || subscription_type.nil?
+      if params[:id].nil? || subscription.nil?
```
vertism: I know this doesn't relate to your change but I think the subscription method should handle where params[:id] is nil. 
Although I'm not sure that would be a valid route anyway.

HWallenberg: Thanks @vertism , how about this?

```suggestion
 
      if subscription.nil?

    def subscription
      if params[:id]
        @subscription ||= Subscription.find(params[:id], user_id_token)
      end
    end
```

vertism: That looks good

vertism (approved):

PR #2781, opened 2026-02-18
Refactor my commodities controllers

vertism:
### What?

Various refactoring of My Commodities controllers including:
- Use send_data to stream file to browser
- Extract options
- Add subscription method


HWallenberg (approved):
nice thanks.

PR #2784, opened 2026-02-20
MyOTT-516 Iterate Tariff Changes Download Link

HWallenberg:
### Jira link

[MyOTT-516](https://transformuk.atlassian.net/browse/MyOTT-516)

### What?

I have added icon and styling for download link

### Why?

I am doing this because it is required for iteration of the commodity watch list page and will be reused for downloading watchlist in the hero section.

BEFORE
image

AFTER
image


app/views/myott/mycommodities/_download.html.erb
```diff
@@ -1,3 +1,10 @@
 <div class="govuk-!-margin-bottom-6">
-  <%= link_to 'Download tariff changes as spreadsheet', download_myott_mycommodities_path(as_of: as_of.to_fs(:dashed)) %>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-full govuk-!-font-weight-bold">
+      <%= link_to download_myott_mycommodities_path(as_of: as_of.to_fs(:dashed)), class: "download_link #{link_class || ""}" do %>
+        <%= inline_svg_tag("download_icon.svg", class: "download_icon", aria: { hidden: true }) %>
+        <span><%= download_text ||= "Download tariff changes as spreadsheet" %></span>
```
vertism: I don't think you need the = here
```suggestion
        <span><%= download_text || "Download tariff changes as spreadsheet" %></span>
```

vertism: Also, is there a reason the text needs to be passed into the partial?

HWallenberg: I was thinking re-using the partial for the download watch list in the hero section which has different wording.

PR #2785, opened 2026-02-23
[HMRC-1994] Rails 8.1

spencerldixon:
### Jira link

[HMRC-1994](https://transformuk.atlassian.net/browse/HMRC-1994)

### What?

- Upgrades to Rails 8.1.2
- Uses new 8.1 compatible `trade-tariff/routing-filter` gem

### Why?

Dependabot has tried to upgrade Rails on FE and Admin and it’s thrown up a load of test failures.


bin/ci
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env ruby
```
willfish, line 1: Is the plan to hook this into the workflows once they're available for all rails apps?

willfish (approved):

bin/ci
spencerldixon: That would be nice, we could get rid of it but it seems like a nice direction to explore. It would be handy to eventually make this a "run 90% of what you need locally" command

PR #2801, opened 2026-02-27
Skip CSRF verification for CSP violation reports

willfish:
### What?

- [x] Skip forgery protection on `CspReportsController#create`

### Why?

CSP violation reports are POST requests sent automatically by the browser. They don't carry a CSRF token, so Rails rejects them with a 422. This is expected behaviour for browser-initiated reporting endpoints.


HWallenberg (approved):
thank you this will reduce some of the logging noise 🙏

PR #2815, opened 2026-03-06
[HMRC-1997] Remove hardcoded default for secret

spencerldixon:
### Jira link

[HMRC-1997](https://transformuk.atlassian.net/browse/HMRC-1997)

### What?

- Removes hardcoded fallback default for `SECRET_KEY_BASE`
- Requires ENV loaded with `ENV.fetch` to ensure an error is raised if the env var is not set
- Use `XXXX` in `.env.development` to be explicit and scrub the old key (we should rotate this in production just to be safe too)

### Why?

AMR CyberSecurity recommends removing the default hardcoded secret and requiring environment variables to be explicitly configured.

The application should fail to start if required secrets are not configured, rather than falling back to insecure defaults.

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -19,8 +19,8 @@ PERMUTATIONS=true
 PORT=3001
 ROO_WIZARD=true
 SEARCH_BANNER=false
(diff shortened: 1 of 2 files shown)


willfish (approved):
This is definitely a non-issue/not a live secret

PR #2816, opened 2026-03-06
Revert "[HMRC-1997] Remove hardcoded default for secret"

willfish:
Reverts trade-tariff/trade-tariff-frontend#2815


willfish:
Having to revert as this is breaking deployments since the CI build process depends on this value existing (even if its different in the actual live environment runtime)

PR #2828, opened 2026-03-10
MYOTT-552 Open links in new tab

vertism:
### Jira link

[MYOTT-552](https://transformuk.atlassian.net/browse/MYOTT-552)

### What?

Opens links in a new tab, and adds copy to explain it.


HWallenberg (approved):
lgtm!

PR #2829, opened 2026-03-10
OTTIMP-367 Corrected 6 links to the new feedback method

shjohnson:
### Jira link

[OTTIMP-367-2](https://transformuk.atlassian.net/browse/OTTIMP-367-2)

### What?

I have a change that makes all in-app “give feedback” links go to the external survey (`https://surveys.transformuk.com/s3/17fead99a348`) instead of the internal feedback page (`/feedback`). Updated in six places: footer, interactive phase banner, interactive useful banner, commodities interactive search feedback, Help page, and MyOTT unsubscribe confirmation. Links open in a new tab with `target="_blank"` and `rel="noopener noreferrer"`. The spec that asserted `request_id` on the commodities feedback link was updated to assert the survey URL instead.

### Why?

I am doing this because we want a single feedback destination (the survey) and to avoid users landing on `/feedback?day=10&month=3&year=2026` when they follow feedback links from dated or search pages. Sending everyone to the same survey URL keeps feedback collection consistent and removes dependence on the in-app feedback form for these entry points.

### DIFF
--- a/app/views/commodities/_interactive_search_feedback.html.erb
+++ b/app/views/commodities/_interactive_search_feedback.html.erb
@@ -2,7 +2,7 @@
   <div class="govuk-inset-text app-inset-text--blue">
     <h2 class="govuk-heading-m govuk-!-margin-bottom-1">Give feedback about this service</h2>
     <p class="govuk-body">Help us improve this service by <%= link_to "giving your feedback (opens in a new tab)",
-        feedback_path(request_id: params[:request_id]),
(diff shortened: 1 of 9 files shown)


willfish (dismissed):

willfish:
It's worth catching up with Chris Booth before we ship this. My understanding is that they're specifically using feedback to gather information about myott

willfish (approved):

PR #2832, opened 2026-03-11
Remove duplicate interactive search feedback partial

willfish:
### What?

- [x] Remove `commodities/_interactive_search_feedback` partial and its spec
- [x] Remove the render call from `commodities/show.html.erb`

### Why?

The commodities page rendered its own inline "Give feedback about this service" inset when interactive search was enabled, duplicating the global feedback banner already rendered by the layout. The layout-level banner is sufficient.


amberstarlight (approved):

shjohnson (approved):
💚

PR #2868, opened 2026-03-24
HMRC:-1965: Remove HTTP fallback and tighten security groups

rasikasri:
# Jira link
[HMRC-1965](https://transformuk.atlassian.net/browse/HMRC-1965)

## What?
Remove the HTTP listener from all services and close port 8080 in security groups. After this, only encrypted HTTPS traffic reaches the containers.

I have:

- Update config/puma.rb to restrict the HTTP port bind only to development - only keep ssl_bind on 8443
- Update Dockerfile HEALTHCHECK to check port 8443
- Update PORT environment variable to 8443

## Why?

I am doing this because:

- Only encrypted HTTPS traffic allowed in ecs cluster

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,8 +62,6 @@ USER tariff
 COPY --chown=tariff:tariff --from=builder /build .
 COPY --chown=tariff:tariff --from=builder /usr/local/bundle/ /usr/local/bundle/
 
-EXPOSE 8443
-
 HEALTHCHECK CMD nc -z 0.0.0.0 $SSL_PORT
 
 CMD ["bundle", "exec", "puma", "-C", "config/puma.rb"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,7 +49,7 @@ RUN apk add --no-cache \
 
 ENV RAILS_SERVE_STATIC_FILES=true \
     RAILS_ENV=production \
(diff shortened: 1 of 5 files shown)


config/puma.rb
```diff
@@ -31,8 +31,12 @@
 
 environment ENV['RACK_ENV'] || 'development'
 
+rails_env = ENV.fetch("RAILS_ENV", "development")
+
 # Explicit HTTP bind,  default is 3000.
-bind "tcp://0.0.0.0:#{ENV.fetch('PORT', 3000)}"
+if rails_env == "development"
+  bind "tcp://0.0.0.0:#{ENV.fetch('PORT', 3000)}"
```
willfish, line 38:
```suggestion
  bind "tcp://0.0.0.0:#{ENV.fetch('PORT', 3001)}"
```

Dockerfile
```diff
@@ -62,11 +62,10 @@ USER tariff
 COPY --chown=tariff:tariff --from=builder /build .
 COPY --chown=tariff:tariff --from=builder /usr/local/bundle/ /usr/local/bundle/
 
-EXPOSE 8080
```
willfish, line 65: I think we should remove this to be consistent with the other Dockerfiles

willfish (approved):

PR #2870, opened 2026-03-24
[HMRC-2129] - Add critical footnotes to import/export tabs

spencerldixon:
### Jira link

[HMRC-2129](https://transformuk.atlassian.net/browse/HMRC-2129)

### What?

- Adds critical footnotes to the import and export tabs
- Corrects "export" tab heading to be consistent font size with import tab heading

### Why?
CDS has made the changes so we can amend of Banner for CCs in Chapter 99

We need to add the same critical footnote to the export tab as requested in the request above. The banner should reflect the message below on the export tab.

### DIFF
--- a/app/views/measures/_measures.html.erb
+++ b/app/views/measures/_measures.html.erb
@@ -23,8 +23,11 @@
 
   <!-- Import tab -->
   <div class="govuk-tabs__panel" id="import">
-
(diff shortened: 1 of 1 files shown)


HWallenberg (approved):
lgtm

Image

PR #2871, opened 2026-03-24
Refactor ErrorsController to eliminate repeated respond_to boilerplate

neilmiddleton:
## Summary

- Extract the repeated `respond_to` pattern across 7 error actions into a single private `render_error` method
- Drive the 7 structurally identical actions from an `ERROR_RESPONSES` config table using `define_method`
- `not_found` and `maintenance` remain explicit as they render their own named templates with no header/message locals
- No behaviour changes — all 27 existing request specs pass

## Test plan

- [x] All existing request specs pass (`spec/requests/errors_controller_spec.rb`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -9,76 +9,70 @@ class ErrorsController < ApplicationController
                 :disable_switch_service_banner,
                 :skip_news_banner
 
+  ERROR_RESPONSES = {
+    unprocessable_entity: {
+      status: :unprocessable_entity,
+      header: 'Unprocessable entity',
+      message: "We're sorry, but we cannot process your request at this time.<br>
+               Please contact support for assistance or try a different request.",
+      html_safe: true,
+      json_error: 'Unprocessable entity',
+    },
+    internal_server_error: {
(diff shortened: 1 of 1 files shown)


app/controllers/errors_controller.rb
```diff
@@ -9,6 +9,65 @@ class ErrorsController < ApplicationController
...
+      header: 'Not acceptable',
+      message: "Unfortunately, we cannot fulfill your request as it is not in a format we can accept.<br>
+               Please contact support for assistance or try a different request.",
+      html_safe: true,
+      json_error: 'Not acceptable',
+    },
+    not_implemented: {
+      status: :not_implemented,
+      header: 'Not implemented',
+      message: "We're sorry, but the requested action is not supported by our server at this time.<br>
```
willfish, line 54: I wonder if we can just localise all of this?

willfish (approved):

app/controllers/errors_controller.rb
neilmiddleton: we could when we need to?

PR #2879, opened 2026-03-26
Add page_context to OTT survey links and remove noreferrer from feedback entry points

shjohnson:
This change makes survey page context capture robust and consistent for the feedback links introduced in the recent OTT survey rollout. For non-subscription pages, feedback_link_url now appends a URL-encoded page_context query parameter using the current request path, so Alchemer receives explicit page context without relying only on browser referrer behavior. In the same scoped entry points, feedback links opened in a new tab now use rel="noopener" (not noopener noreferrer) so referrer is not suppressed. Subscription pages continue to use the existing in-app feedback_path flow unchanged.

### DIFF
--- a/.trufflehog-exclude-paths.txt
+++ b/.trufflehog-exclude-paths.txt
@@ -1,0 +1,5 @@
+^node_modules/
+^public/packs/
+^terraform/\.terraform/
+^vendor/bundle/
+^tmp/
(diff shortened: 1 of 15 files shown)


vertism (dismissed):

.trufflehog-exclude-paths.txt
```diff
@@ -0,0 +1,5 @@
+^node_modules/
```
willfish, line 1: Doesn't trufflehog just need .gitignore?

willfish (dismissed):

willfish (approved):

PR #2886, opened 2026-03-31
OTTIMP-474: Live issues status sorting

montychristo:
### Jira link

[OTTIMP-474](https://transformuk.atlassian.net/browse/OTTIMP-474)

### What?

Server-side sorting for the Live Issues log: default order is **Active** issues first, then **Resolved**, with **newest `updated_at` first** within each group. The **Status** column header links between that order and the inverse (**Resolved** first, still newest-within-group for Active). Uses `?sort=desc` to toggle; no client-side table sorting.

### Why?

So users can see open issues at the top by default and switch to resolved-first when they need it

### DIFF
--- a/app/controllers/live_issues_controller.rb
+++ b/app/controllers/live_issues_controller.rb
@@ -5,34 +5,10 @@ class LiveIssuesController < ApplicationController
 
   def index
     @sort_direction = sort_direction
-    @live_issues = sorted_live_issues
+    @live_issues = LiveIssue.sorted_by_status(@sort_direction)
   end
 
 private
-
-  def sorted_live_issues
-    LiveIssue.all.sort_by do |live_issue|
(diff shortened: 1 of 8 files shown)


app/controllers/live_issues_controller.rb
```diff
@@ -4,6 +4,37 @@ class LiveIssuesController < ApplicationController
...
 
   def index
-    @live_issues = LiveIssue.all
+    @sort_direction = sort_direction
+    @live_issues = sorted_live_issues
+  end
+
+private
+
+  def sorted_live_issues
```
willfish: How do you feel about this living in the model?

willfish (dismissed):

app/controllers/live_issues_controller.rb
montychristo: Yes let's do that 👍

willfish (approved):

PR #2892, opened 2026-04-02
MYOTT-594 Improve normalisation of upload data

vertism:
### Jira link

[MYOTT-594](https://transformuk.atlassian.net/browse/MYOTT-594)

### What?

Some values were being loaded as floats e.g. 9876543210.0 and the code was converting to a string and stripping the decimal point which incorrectly left an 11 digit commodity code.

The value type is now checked and converted to an integer first.

### DIFF
--- a/app/services/commodity_codes_extraction_service.rb
+++ b/app/services/commodity_codes_extraction_service.rb
@@ -29,8 +29,17 @@ class CommodityCodesExtractionService
       end
 
     rows.filter_map do |value|
-      code = value.to_s.gsub(/[^0-9]/, '')
+      code = normalize_code_value(value)
       code if code.present? && code.length.between?(1, 20)
+    end
+  end
+
+  def normalize_code_value(value)
+    case value
+    when Numeric
+      value.floor.to_s if value.finite?
+    else
+      value.to_s.sub(/\.0+\z/, '').gsub(/[^0-9]/, '') # Remove trailing .0 and any non-numeric characters
     end
   end
 end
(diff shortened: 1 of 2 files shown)


neilmiddleton:
Clean fix for a fiddly problem. A few notes:

**`value.floor` vs `value.round` for the `Numeric` branch**

`floor` is safe here because commodity codes are integers and 10-digit values fit comfortably within IEEE 754 double precision (< 2^53), so there's no risk of `9876543210.0` being stored as `9876543209.999...`. Worth a comment though — future readers might wonder why `floor` rather than `to_i` or `round`.

**`value.finite?` guard**

`nil` is returned when the value is `Infinity` or `NaN`, which flows through to `code.present?` and gets filtered. Correct, but it's a subtle interaction. No action needed, just noting it's doing two jobs quietly.

**`else` branch still mangles non-integer decimals**

`"1234567890.5"` → `.sub(/\.0+\z/, '')` doesn't match → `.gsub(/[^0-9]/, '')` → `"12345678905"` (11 digits). This is pre-existing behaviour and non-integer commodity codes are invalid anyway, so not a regression — just worth being aware that this path only fully fixes the `.0` suffix case.

**Tests**

Cover both the CSV and Excel paths for the reported bug, which is what matters. 

Happy with this — the `when Numeric` dispatch is the right approach rather than trying to be clever with string regexes on float-serialised values.

neilmiddleton (approved):

willfish (approved):

PR #2908, opened 2026-04-10
Cache repeated association access in measure collection partial

neilmiddleton:
## Why

`ApiEntity#has_many` reconstructs its collection on every call — it reads the raw instance variable, optionally wraps it, and optionally filters it, returning a new object each time. `_measure.html.erb` is rendered once per measure in a collection (20–50 per page), and previously accessed `measure.geographical_area`, `measure.measure_type`, and `measure.footnotes` multiple times each per render, producing 100–200 redundant allocations per request.

## What

Assign the three associations to locals at the top of the partial:

```erb
<% geo_area     = measure.geographical_area %>
<% measure_type = measure.measure_type %>
<% footnotes    = measure.footnotes %>
```

All subsequent uses within the template reference the local variable. The accessor is now called exactly once per measure render with no change to behaviour or output.

### DIFF
--- a/app/views/measures/_measure.html.erb
+++ b/app/views/measures/_measure.html.erb
@@ -69,7 +69,7 @@
   </td>
 
   <td class="footnotes-col govuk-table__cell govuk-table__header--numeric">
-    <% if measure.has_measure_footnotes? %>
+    <% if footnotes.any? %>
       <div data-controller="anchor">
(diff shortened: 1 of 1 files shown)


app/views/measures/_measure.html.erb
```diff
@@ -68,7 +71,7 @@
   <td class="footnotes-col govuk-table__cell govuk-table__header--numeric">
     <% if measure.has_measure_footnotes? %>
```
vertism:
```suggestion
    <% if footnotes.any? %>
```

vertism:
Is this PR still necessary after https://github.com/trade-tariff/trade-tariff-frontend/pull/2912?

PR #2910, opened 2026-04-10
Eliminate duplicate iteration in rules of origin and permutation group partials

neilmiddleton:
## Summary

- In the three rules-of-origin tab partials (`legacy/_tab`, `_tab_uk`, `_tab_xi`), `flat_map(&:links)` was being called twice on `rules_of_origin_schemes` — once for the `.any?` guard and a second time to iterate with `.uniq(&:id).each`. Each call traverses all schemes and their links collections. This is fixed by extracting the result to a `scheme_links` local before the guard, so the traversal happens once.

- In `measure_conditions/_permutation_group.html.erb`, `permutation_group.permutations` was accessed twice and the guidance-condition chain (`flat_map(&:measure_conditions).select(&:has_guidance?).uniq(&:resource_id)`) was inlined at the render call site. This is fixed by extracting `permutations` and `conditions_with_guidance` locals at the top of the partial so both chains execute exactly once per render.

### DIFF
--- a/app/views/rules_of_origin/_tab_uk.html.erb
+++ b/app/views/rules_of_origin/_tab_uk.html.erb
@@ -40,8 +40,9 @@
 
     <nav role="navigation">
       <ul class="govuk-list govuk-list-s">
-        <% if rules_of_origin_schemes.flat_map(&:links).any? %>
-          <% rules_of_origin_schemes.flat_map(&:links).uniq(&:id).each do |link| %>
+        <% scheme_links = rules_of_origin_schemes.flat_map(&:links).uniq(&:id) %>
+        <% if scheme_links.any? %>
+          <% scheme_links.each do |link| %>
             <p>
               <li>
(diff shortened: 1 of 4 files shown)


app/views/rules_of_origin/_tab_uk.html.erb
```diff
@@ -40,8 +40,9 @@
 
     <nav role="navigation">
       <ul class="govuk-list govuk-list-s">
-        <% if rules_of_origin_schemes.flat_map(&:links).any? %>
-          <% rules_of_origin_schemes.flat_map(&:links).uniq(&:id).each do |link| %>
+        <% scheme_links = rules_of_origin_schemes.flat_map(&:links).uniq(&:id) %>
+        <% if scheme_links.any? %>
```
vertism, line 44: `if scheme_links.any?` isn't needed

vertism (approved):

PR #2912, opened 2026-04-10
Memoize ApiEntity has_many and has_one association getters

neilmiddleton:
Every call to a `has_many` or `has_one` accessor was constructing a new wrapper object and re-running the filter against the raw collection on each invocation. Because the cost was invisible at the definition site, call sites had been patched individually — capturing accessor results in local variables in views and presenters — to avoid repeated work.

The root fix is in `lib/api_entity.rb`:

- **`has_many` getter**: caches the wrapped-and-filtered result in `@_<name>` (distinct from `@<name>`, which holds the raw array populated by the setter). On subsequent calls the cached value is returned immediately.
- **`has_many` setter** and **`add_<singular>` method**: both invalidate `@_<name>` so the cache is never stale if data is written after first access.
- **`has_one` getter**: same pattern — caches the entity in `@_<name>` on first access.
- **`has_one` setter**: invalidates `@_<name>` on write.

Existing call-site local-variable captures remain valid as defensive coding but are no longer load-bearing. No behaviour change is observable through the public API.

Note, now that ApiEntity memoizes associations at the source, the local-variable patches in the earlier PRs (#2907, #2908, #2909) remain valid as defensive coding but are no longer strictly necessary for performance. They don't hurt though.

### DIFF
--- a/lib/api_entity.rb
+++ b/lib/api_entity.rb
@@ -178,7 +178,12 @@ private
 
       relationships << association
 
-      attr_reader association.to_sym
+      define_method(association) do
+        cache_ivar = "@_#{association}"
+        return instance_variable_get(cache_ivar) if instance_variable_defined?(cache_ivar)
+
+        instance_variable_set(cache_ivar, instance_variable_get("@#{association}"))
+      end
 
       define_method("#{association}=") do |attributes|
         entity = if attributes.nil?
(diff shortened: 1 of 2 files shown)


vertism:
Nice, fixing the issue at the source. Would be good to add some tests for the invalidation on the setters.

neilmiddleton:
Specs added

vertism (approved):

PR #2913, opened 2026-04-10
Move pure display logic from CommoditiesHelper into presenters

neilmiddleton:
`CommodityPresenter` and `HeadingPresenter` were near-empty `SimpleDelegator` wrappers while `commodities_helper.rb` held ~30 methods, many of which are pure display logic that takes a commodity/declarable as an argument and uses only its attributes. These belong on the presenter, not a helper.

`TradeTariffFrontend::Presenter` uses `SimpleDelegator`, so presenter instances already respond to all model methods. Adding display methods directly to the presenter means views can call `commodity.leaf_position` or `declarable.footnote_heading` instead of the helper form.

## What moved

**To `DeclarablePresenter`** (shared by Commodity and Heading):
- `footnote_heading` — calls `I18n.t` with `self.id` and `self.model_name`
- `format_full_code` — builds the full segmented code HTML from `self.code`
- `format_commodity_code` — formats the display short code with non-breaking spaces
- `format_commodity_code_based_on_level` — conditionally renders the tree code based on indent and producline suffix
- `abbreviate_commodity_code` — returns full code for declarables, abbreviated for non-declarables

**To `CommodityPresenter`** (commodity-specific):
- `leaf_position` — returns `' last-child'` when `last_child?` is true
- `commodity_level(initial_indent = nil)` — returns the CSS level class from normalised indent

## What stayed in the helper

Methods that require Rails view context (`tag`, `content_tag`, `safe_join`, `render`, `url_options`, `concat`) remain in the helper unchanged. String-based utilities that take a raw code string rather than a commodity (`divide_commodity_code`, `abbreviate_code`, `sanitize_quotes`) also remain.

The helper's public commodity-display methods now delegate to the object (`commodity.format_full_code`, etc.), preserving backward compatibility for views that work with raw model objects from `heading.commodities` or search results. The `_measures.html.erb` view is updated to call `declarable.footnote_heading` directly since `declarable` is guaranteed to be a presenter there.

Presenter specs added for all moved methods in `spec/presenters/declarable_presenter_spec.rb` and the new `spec/presenters/commodity_presenter_spec.rb`.


app/helpers/commodities_helper.rb
```diff
@@ -1,12 +1,10 @@
...
   def footnote_heading(declarable)
-    t('tabs.footnote.heading', goods_nomenclature_item_id: declarable.id, declarable_type: declarable.model_name.singular)
+    declarable.footnote_heading
   end
 
   def leaf_position(commodity)
-    if commodity.last_child?
-      ' last-child'
-    end
+    ' last-child' if commodity.last_child?
```
vertism: Should this be using the presenter method?
```suggestion
    commodity.leaf_position
```

app/presenters/declarable_presenter.rb
```diff
@@ -6,4 +6,74 @@ def to_s
   def self.model_name
     name.chomp('Presenter').constantize.model_name
   end
+
+  def footnote_heading
+    I18n.t('tabs.footnote.heading', goods_nomenclature_item_id: id, declarable_type: model_name.singular)
+  end
+
+  def format_full_code
```
vertism, line 14: How are these new methods being used?

app/helpers/commodities_helper.rb
neilmiddleton: https://github.com/trade-tariff/trade-tariff-frontend/pull/2913/commits/64e6f0f79f45c3aab8e53c90ca816118a319ab64

app/presenters/declarable_presenter.rb
neilmiddleton: https://github.com/trade-tariff/trade-tariff-frontend/pull/2913/changes/f3d9e3f9d6b478e16672493187e99a9a88763484

app/helpers/commodities_helper.rb
```diff
@@ -1,12 +1,10 @@
...
   end
 
   def leaf_position(commodity)
-    if commodity.last_child?
-      ' last-child'
-    end
+    as_commodity_presenter(commodity).leaf_position
   end
 
   def commodity_level(commodity, initial_indent)
```
vertism, line 10: Is this missing a call to the presenter?

PR #2914, opened 2026-04-13
Build hash index for JSONAPI included resources to fix O(n²) parse timeout

neilmiddleton:
`find_included` was doing an O(n) linear scan of the `included` array on every call. `parse_resource` calls `find_and_parse_included` for every relationship of every resource, so for a commodity response with 50+ measures each having ~10 relationships and hundreds of included objects the total work scaled as O(n²). This matches the `execution expired` `Timeout::Error` seen in production.

**Fix 1 — O(1) index for included resources**

A hash index (`included_index`) is built once per parser instance, keyed by `[type, id]`. Every call to `find_included` now does a single hash lookup instead of a full array scan.

**Fix 2 — memoize parsed included resources**

`find_and_parse_included` now stores its result in `parsed_included_cache` (also keyed by `[type, id]`) before returning. Subsequent calls for the same resource return the cached object immediately without re-parsing. Shared resources such as the erga omnes geographical area (referenced by every measure) are now parsed exactly once rather than once per reference. The cache check uses `key?` rather than a nil/falsy check so that a legitimately missing resource (cached as `{}`) is returned from cache correctly on repeat calls.

### DIFF
--- a/lib/tariff_jsonapi_parser.rb
+++ b/lib/tariff_jsonapi_parser.rb
@@ -85,14 +85,21 @@ class TariffJsonapiParser
   def find_and_parse_included(name, id, type)
     return nil if id.blank? || type.blank?
 
+    key = [type.to_s, id.to_s]
+    return parsed_included_cache[key] if parsed_included_cache.key?(key)
+
     found_resource = find_included(id, type)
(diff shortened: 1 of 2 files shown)


willfish (approved):
Probably a good idea to replicate this in the admin app, too.

Nice solve

PR #2922, opened 2026-04-14
Remove redundant requires

vertism:
### What?

Classes in the lib directory are autoloaded (https://github.com/trade-tariff/trade-tariff-frontend/blob/main/config/application.rb#L70) so it is not necessary to require them when used.

Validated this works by running `bundle exec rails zeitwerk:check`


spencerldixon (approved):
LGTM 👍

PR #2924, opened 2026-04-14
[HMRC-2185] Remove tariff last updated

spencerldixon:
### Jira link

[HMRC-2185](https://transformuk.atlassian.net/browse/HMRC-2185)

### What?

When reviewing the apps for performance optimisations, I noticed the `/updates/latest` endpoint was called on every request to set the @tariff_last_updated timestamp. This should not be necessary when the timestamp should only change after an import (once a day).

More investigating shows that we removed this in an earlier PR: [OTTIMP-439](https://transformuk.atlassian.net/browse/OTTIMP-439) so this is no longer needed.

This PR removed mentions of `tariff_last_updated` almost entirely, but leaves the model intact should we wish to bring this back.

### Why?

We currently make a request to `/updates/latest` on every request but don't actually do anything with this.
Removing this means way fewer requests, faster service, and cleaner code.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -10,7 +10,6 @@ class ApplicationController < ActionController::Base
   before_action :maintenance_mode_if_active
 
   before_action :set_cache
-  before_action :set_last_updated
   before_action :set_path_info
   before_action :set_search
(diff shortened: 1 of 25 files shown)


willfish (approved):
Yay

PR #2925, opened 2026-04-14
Remove unnecessary gems

vertism:
### What?

I have removed the following from the Gemfile:
- `csv` - not used directly and is pulled in as a dependency of roo anyway
- `net-http-persistent` - already included as a dependency of `faraday-net_http_persistent`
- `plek` - is not used anywhere but is loaded as a transitive dependency anyway
- `aws-sdk-rails` - not required since actionmailer delivery was changed to :ses_v2
- `rubocop-capybara` - already included as a dependency of `govuk-rubocop` 
-  `rack-test` - not used

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,7 +9,6 @@ end
 
 gem 'rails', '~> 8.1'
 
-gem 'csv'
 gem 'faraday'
 gem 'faraday-http-cache'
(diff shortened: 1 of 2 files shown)


spencerldixon (approved):
LGTM 👍

PR #2927, opened 2026-04-16
Remove unused Duty Calculator errors controller

vertism:
### What?

The errors controller defined in the Duty Calculator module is not being used so can be removed.


spencerldixon (approved):
LGTM 👍

PR #2934, opened 2026-04-23
OTTIMP-524 Create GroupedMeasuresPresenter

vertism:
### Jira link
[OTTIMP-524](https://transformuk.atlassian.net/browse/OTTIMP-524)

### What?

Extracts logic for which measure details to show based on UK or XI tariff into a new presenter.
This simplifies the views and removes the need for separate uk and xi navigation partials.


willfish (approved):
Exquisite

PR #2937, opened 2026-04-27
OTTIMP-525 Stop hiding legal base column on UK tariff

vertism:
### Jira link

[OTTIMP-525](https://transformuk.atlassian.net/browse/OTTIMP-525)

### What?

Legal base column had been unhidden selectively. It can now be unhidden everywhere.

### Why?

- Traders need to verify the legal basis for tariff measures affecting their imports and exports.
- On the UK front end this information is plumbed in but currently hidden, which forces users to rely on alternative sources and undermines confidence in OTT’s authoritative guidance. The reason it was hidden no longer applies (post-brexit UK legislation not yet formed)


spencerldixon (approved):
LGTM 👍

PR #2966, opened 2026-05-07
Fix CSP violations for GTM script and blob connections

vertism:
### What?

Updates CSP-related frontend scripts so GTM and inline page scripts are nonce-compliant and compatible with es-module-shims.

- Added nonce support to the GTM head snippet by rendering it through a nonce-enabled Rails script helper.
- Added nonce support to the search autocomplete inline script.
- Added nonce to the es-module-shims options script tag while keeping it as raw JSON (not wrapped in Rails javascript helper output), so it remains valid for parsing.
- Added a connect-src directive to allow self, https, and blob connections.

### Why?

I am doing this because CSP report-only logs were showing violations for inline scripts and blob connections, and one attempted nonce fix caused es-module-shims JSON parsing to fail.

This change:
- Removes noisy CSP violations for known inline scripts by ensuring matching nonces are present.
- Preserves strict script policy without enabling unsafe-inline or unsafe-eval.
- Fixes the runtime error where es-module-shims rejected invalid JSON content.
- Keeps GTM functioning while aligning implementation with planned CSP enforcement.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -29,7 +29,7 @@
   <%= yield :head %>
 
   <script async src="/javascripts/es-module-shims.js?v=<%= CURRENT_REVISION %>"></script>
-  <script type="esms-options">
+  <script type="esms-options" nonce="<%= content_security_policy_nonce %>">
     {
       "noLoadEventRetriggers": true
     }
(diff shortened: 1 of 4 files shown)


neilmiddleton:
One thing worth confirming: the `esms-options` script tag uses `nonce="<%= content_security_policy_nonce %>"` directly rather than going through `javascript_tag`. The nonce value itself is correct, but worth checking whether the browser treats a `type="esms-options"` script with a nonce the same as a `type="text/javascript"` script for CSP purposes. The spec says the nonce applies to any script element regardless of type, so this should be fine, but it's worth verifying in the CSP violation logs after this lands.

neilmiddleton (approved):

willfish (approved):
cc @anupamc

PR #2975, opened 2026-05-08
HMRC-2209: Bump ECS service tf module 

AigulDj:
### Jira link

[HMRC-2209](https://transformuk.atlassian.net/browse/HMRC-2209)

### What?

I have
- Updated ECS service terraform module version from 2.0.1 to 3.0.1
- Explicitly pin ECS autoscaling settings in prod to current values

- Temporarily increase cpu_alarm_threshold from 70 to 85.

### Why?

I am doing this because...

- This ensures there is no behaviour change in production following the ECS module upgrade.
- Current prod autoscaling uses a CPU target of 50% with zero scale-in and scale-out cooldowns; Configurable cooldowns are introduced to stabilise scaling behaviour and avoid unnecessary churn.

- This temporary adjustment reduces alert noise while we evaluate and determine the appropriate long-term threshold values.

Ref
[HMRC-1766: Configure scaling policies](https://github.com/trade-tariff/trade-tariff-platform-terraform-modules/pull/84)

Deployment risk
🟠 Medium – infrastructure configuration changes affecting autoscaling logic.

Reason for rating:
This change modifies Terraform autoscaling logic and resource creation conditions. While production behaviour is explicitly preserved and changes are first applied to non-production environments, the work alters infrastructure behaviour and requires team awareness

### DIFF
--- a/terraform/config_production.tfvars
+++ b/terraform/config_production.tfvars
@@ -5,3 +5,7 @@ memory        = 4096
 service_count = 4
 min_capacity  = 3
 max_capacity  = 16
+
+# Temporarily pinned autoscaling cooldown values in Production to preserve existing behaviour while testing changes in dev and staging
+scale_in_cooldown  = 0
+scale_out_cooldown = 0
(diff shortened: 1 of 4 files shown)


terraform/config_production.tfvars
```diff
@@ -5,3 +5,7 @@ memory        = 4096
 service_count = 4
 min_capacity  = 3
 max_capacity  = 16
+
+# Temporarily pinned autoscaling cooldown values in Production to preserve existing behaviour while testing changes in dev and staging
```
willfish, line 9: 🙌

willfish (approved):

terraform/config_production.tfvars
AigulDj: Following your advice :)

PR #2997, opened 2026-05-12
HMRC:-2201: ActionView::Template::Error in frontend

rasikasri:
## What:
`DutyCalculator::Steps::Base: import_date_step_path` — builds `import_date_path` with day/month/year from user_session.import_date when present.
`ImportDestination` and `Stopping`: `previous_step_path` now uses import_date_step_path.
`ImportDateController`: day / month / year now come from query params if present, else user_session.import_date, else today — so returning from import-destination keeps the chosen date and does not overwrite the session with today.
`ConfirmationDecorator#import_date_path`: edit link includes the same date query params when the session has an import date.

## Why:
Duty Calculator back link omits the session import date, and ImportDateController#show defaults day/month/year to today when params are absent, overwriting the session and triggering commodity fetches with default date which is commodity would not be valid.

This scenario cause 404 response from backend

## Ticket:
https://transformuk.atlassian.net/browse/HMRC-2201

## Risk:

**Risk level:** 🟢 

**Reason for rating:**
Fixing an existing bug

Rate the overall risk of deploying this change:

🟢 Green  – Low risk. Good to go, standard review applies.

🟠 Amber  – Medium risk. Socialise with the team before merging.

🔴 Red    – High risk. Requires explicit approval from Thor or Neil before merging.

🟢 GREEN – things that are typically low risk:

- Copy or content changes to GOV.UK Frontend components (labels, hint text, page titles)
- Adding or updating GOV.UK Design System components with no behaviour change
- New tests or improved test coverage with no production code changes
- Dependency bumps with no API changes (minor/patch gems, npm packages)
- Config/env var additions that are purely additive and have safe defaults
- Refactors with full test coverage and no behaviour change
- Adding or updating CloudWatch alarms or dashboards (read-only observability)
- Terraform formatting or variable renaming with no resource recreation
- Static asset changes (images, fonts, stylesheets) with no layout impact

🟠 AMBER – things that need a team conversation first:

- Changes to page layout or navigation that affect all or many user journeys
- Modifications to how commodity data or tariff information is presented to traders
- New or modified API calls to the backend or admin services
- Adding or changing feature flags that affect live user journeys
- Changes to error pages, accessibility markup, or GOV.UK service header/footer
- Search UI changes (autocomplete behaviour, result rendering, filter logic)
- Significant Sass/CSS changes that could affect rendering across browsers or screen sizes
- Infrastructure changes that alter networking, security groups, or IAM permissions in non-production first
- Terraform changes that will cause a resource replacement (check plan output carefully)
- Changes to CI/CD pipeline steps or deployment order dependencies
- Removing or deprecating a route, controller action, or view partial still in use elsewhere

🔴 RED – requires explicit approval from Thor or Neil:

- Changes to how legally significant content is surfaced (trade remedies, prohibitions, licensing, sanctions)
- Modifications to the declarable goods flow or any trader-facing regulatory journey
- Any change to how the service integrates with the identity/authentication layer
- Changes to production AWS infrastructure that cannot be easily rolled back
- Secrets rotation or changes to how credentials are stored, scoped, or accessed
- Significant architectural shifts (e.g. new service boundaries, caching topology changes)
- Changes that affect GOV.UK service compliance, accessibility standards (WCAG 2.2 AA), or government design system conformance

### DIFF
--- a/app/controllers/duty_calculator/steps/import_date_controller.rb
+++ b/app/controllers/duty_calculator/steps/import_date_controller.rb
@@ -58,15 +58,27 @@ module DutyCalculator
       end
 
       def day
-        params[:day] || now.day
+        reference_date_components[0].to_s
       end
 
       def month
-        params[:month] || now.month
+        reference_date_components[1].to_s
       end
 
       def year
-        params[:year] || now.year
+        reference_date_components[2].to_s
+      end
+
+      def reference_date_components
+        @reference_date_components ||=
+          if params[:day].present? && params[:month].present? && params[:year].present?
+            [params[:day].to_i, params[:month].to_i, params[:year].to_i]
+          elsif user_session.import_date.present?
+            d = user_session.import_date
+            [d.day, d.month, d.year]
+          else
+            [now.day, now.month, now.year]
+          end
       end
 
       def now
--- a/app/decorators/duty_calculator/confirmation_decorator.rb
+++ b/app/decorators/duty_calculator/confirmation_decorator.rb
@@ -42,7 +42,13 @@ module DutyCalculator
   private
 
     def import_date_path
-      super(commodity_code: user_session.commodity_code)
+      params_hash = { commodity_code: user_session.commodity_code }
+      if user_session.import_date.present?
+        params_hash[:day] = user_session.import_date.day
+        params_hash[:month] = user_session.import_date.month
+        params_hash[:year] = user_session.import_date.year
+      end
+      super(**params_hash)
     end
 
     def additional_code_path
--- a/app/models/duty_calculator/steps/base.rb
+++ b/app/models/duty_calculator/steps/base.rb
@@ -37,7 +37,22 @@ module DutyCalculator
         UserSession.get
       end
 
+      def import_date_step_path
+        import_date_path(import_date_path_params_for_session)
+      end
+
       private
+
+      def import_date_path_params_for_session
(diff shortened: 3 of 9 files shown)


neilmiddleton (dismissed):
One observation: the `stopping_spec.rb` change adds `:with_import_date` to the session factory and now asserts the path includes `day: 1, month: 1, year: 2025`. That's correct, but the existing `stopping_spec` was previously asserting against `import_date_path(commodity_code: user_session.commodity_code)` without a date - which means that test was previously passing without the date params, implying the old behaviour was tested as correct. Worth confirming there isn't a separate test case somewhere that should be retained to cover the no-import-date-in-session path for `Stopping#previous_step_path`, equivalent to what was added for `ImportDestination`.

rasikasri:
> One observation: the `stopping_spec.rb` change adds `:with_import_date` to the session factory and now asserts the path includes `day: 1, month: 1, year: 2025`. That's correct, but the existing `stopping_spec` was previously asserting against `import_date_path(commodity_code: user_session.commodity_code)` without a date - which means that test was previously passing without the date params, implying the old behaviour was tested as correct. Worth confirming there isn't a separate test case somewhere that should be retained to cover the no-import-date-in-session path for `Stopping#previous_step_path`, equivalent to what was added for `ImportDestination`.

added the default test back again now

neilmiddleton (approved):

PR #3023, opened 2026-05-20
AI-334: Show exact guided search matches as results

willfish:
### Jira link

[AI-334](https://transformuk.atlassian.net/browse/AI-334)

### What?

- [x] Increase guided search client-side and server-side frontend validation from 500 to 1000 characters
- [x] Update the guided search validation error copy to mention 1000 characters
- [x] Render guided search exact matches as the interactive results page instead of redirecting directly to the commodity page
- [x] Update form, controller, model, and JavaScript specs for the new behaviour

### Why?

Guided search needs to accept longer user search inputs, and exact matches should be shown as a single guided search answer so the results journey remains consistent.

### Risk

**Risk level:** 🟠

**Reason for rating:** This changes search UI result rendering for guided search exact matches. It is scoped to guided search, leaves classic search redirects unchanged, and is covered by focused controller, form, model, and JavaScript specs.

### DIFF
--- a/app/forms/interactive_search_form.rb
+++ b/app/forms/interactive_search_form.rb
@@ -2,7 +2,7 @@ class InteractiveSearchForm
   include ActiveModel::Model
   include ActiveModel::Attributes
 
-  MAX_QUERY_LENGTH = 500
+  MAX_QUERY_LENGTH = 1000
   MIN_QUERY_LENGTH = 2
 
   attribute :q, :string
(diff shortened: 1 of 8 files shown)


app/forms/interactive_search_form.rb
```diff
@@ -2,7 +2,7 @@ class InteractiveSearchForm
   include ActiveModel::Model
   include ActiveModel::Attributes
 
-  MAX_QUERY_LENGTH = 500
+  MAX_QUERY_LENGTH = 1000
```
HWallenberg, line 5: 1000? is that enough?

HWallenberg (approved):

PR #3024, opened 2026-05-20
Add LaunchDarkly feature flag DSL

neilmiddleton:
## What:
Integrates LaunchDarkly as the feature flag backend with a thin DSL that keeps LD details out of application code. Adds `feature_gate`, `feature_enabled?`, and `feature_disabled?` helpers to every controller and view. All flags are declared in a central `REGISTRY` constant with safe defaults.

## Why:
Replaces ad-hoc `ENV['FLAG'].to_s == 'true'` checks scattered across `lib/trade_tariff_frontend.rb` with a proper feature flag system that supports dynamic toggling without deploys, environment/service targeting, and percentage rollouts.

## Ticket:
N/A

## Risk:

**Risk level:** 🟢

**Reason for rating:** Purely additive — no existing behaviour changes. The SDK runs in offline mode when `LAUNCHDARKLY_SDK_KEY` is absent, returning the same `false` defaults as the ENV-based flags would. Existing `TradeTariffFrontend.green_lanes_enabled?` methods are untouched.

## Usage

**Declare a flag** (once, in `REGISTRY`):
```ruby
# lib/trade_tariff_frontend/feature_flag.rb
REGISTRY = {
  my_new_feature: false,
}.freeze
```

**Check in controllers/views:**
```erb
<% if feature_enabled?(:my_new_feature) %>…<% end %>
```

**Gate an entire controller or specific actions:**
```ruby
class MyController < ApplicationController
  feature_gate :my_new_feature               # all actions
  feature_gate :my_new_feature, only: :show  # specific action
end
```

**In tests:**
```ruby
stub_feature_flag(:my_new_feature)                 # enabled
stub_feature_flag(:my_new_feature, enabled: false) # disabled
```

## Evaluation context

Every flag evaluation sends a **multi-context** to LaunchDarkly:

- **`application` context** — carries `service` (uk/xi) and `environment`. Use for targeting by deployment, e.g. "only on staging".
- **`user` context** — carries an anonymous UUID from `session[:ld_anonymous_id]`, stable for the session lifetime. Use for percentage rollouts — LaunchDarkly hashes the key consistently so the same visitor always lands in the same bucket.

## Test plan

- [ ] 23 specs across `feature_flag_spec.rb` and `feature_flaggable_spec.rb` cover: enabled/disabled, string vs symbol, unknown flags, offline defaults, single vs multi-context, anonymous user key generation and reuse, `feature_gate` with `only:`, and multi-flag independence
- [ ] 696 controller examples pass unchanged
- [ ] Without `LAUNCHDARKLY_SDK_KEY`: app boots in offline mode, all flags return REGISTRY defaults (false)
- [ ] With `LAUNCHDARKLY_SDK_KEY`: LD client connects on boot, flag changes take effect without a deploy
- [ ] `stub_feature_flag` can be called multiple times in one example without stubs interfering with each other

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -131,6 +131,7 @@ GEM
     aws-sigv4 (1.12.1)
       aws-eventstream (~> 1, >= 1.0.2)
     base64 (0.3.0)
+    benchmark (0.5.0)
     bigdecimal (4.1.2)
     bootsnap (1.24.3)
       msgpack (~> 1.2)
@@ -164,6 +165,7 @@ GEM
     date (3.5.1)
     diff-lcs (1.6.2)
     docile (1.4.1)
+    domain_name (0.6.20240107)
     dotenv (3.2.0)
     dotenv-rails (3.2.0)
       dotenv (= 3.2.0)
@@ -265,6 +267,11 @@ GEM
     html-attributes-utils (1.0.2)
       activesupport (>= 6.1.4.4)
     htmlentities (4.4.2)
+    http (6.0.3)
+      http-cookie (~> 1.0)
+      llhttp (~> 0.6.1)
+    http-cookie (1.1.6)
+      domain_name (~> 0.5)
     i18n (1.14.8)
       concurrent-ruby (~> 1.0)
     importmap-rails (2.2.3)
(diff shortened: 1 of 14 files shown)


Gemfile.lock
```diff
@@ -599,6 +623,7 @@ GEM
     xpath (3.2.0)
       nokogiri (~> 1.8)
     zeitwerk (2.7.5)
+    zlib (3.2.3)
```
willfish, line 626: I'm assuming this is just a wrapper around the C lib zlib so we need to get this into the Dockerfile?

neilmiddleton: No Dockerfile changes needed. `zlib` is a Ruby bundled gem — part of the Ruby standard library, extracted as an explicit gem in newer Ruby versions. It comes pre-compiled with Ruby and doesn't require a separate build step.

The `libz` C library is already present at runtime in the `ruby:4.0.3-alpine3.23` base image as a hard dependency of the `ruby` Alpine package (Ruby itself uses zlib for gem installation and other internals). Both the builder and production stages inherit it from the base image without any explicit `apk add`.

willfish: Okay I tested this myself locally and you definitely need zlib-dev in the build phase:

```dockerfile
RUN apk add --no-cache \
    build-base \
    linux-headers \
    git \
    yarn \
    tzdata \
    yaml-dev \
    openssl-dev \
    zlib-dev \ # << You need this in the build phase
    && \
    cp /usr/share/zoneinfo/Europe/London /etc/localtime && \
    echo "Europe/London" > /etc/timezone
```

You can run `make build` without it and it fails

neilmiddleton:
You're right @willfish, sorry — I was wrong about zlib. `zlib-dev` is needed in the builder stage so the gem's C extension can compile against the zlib headers. Fixed in 780ec59.

PR #3029, opened 2026-05-22
BAU: Reuse expanded internal search query

willfish:
## What:

- [x] Store `expanded_query` on `Search` and include it in internal search requests/cache keys.
- [x] Expose `expanded_query` from `Search::InternalSearchResult`.
- [x] Add the expanded query as a hidden field on the interactive question form.
- [x] Preserve the expanded query on blocking-guidance redirects.
- [x] Stop rendering the extra feedback useful banner divider for interactive search pages.

## Why:

Guided search follow-up submissions repeat the internal search request for each answered question. Carrying the expanded query through the form lets the backend reuse it instead of running query expansion for the same original search term on every iteration.

The loading state also had a visual mismatch: first submissions showed one feedback separator, while follow-up submissions showed two. Removing the interactive-page divider keeps the separator count consistent.

## Ticket:

N/A

## Risk:

**Risk level:** 🟢

**Reason for rating:** This affects the guided search flow, which is not enabled in production yet. The API parameter is additive and the UI separator change is limited to interactive search pages.

### DIFF
--- a/app/controllers/concerns/interactive_searchable.rb
+++ b/app/controllers/concerns/interactive_searchable.rb
@@ -90,7 +90,11 @@ module InteractiveSearchable
 
   def redirect_to_interactive_blocking
     redirect_to perform_search_path(
-      search_params.to_h.merge(interactive_search: 'true', request_id: @search.request_id),
+      search_params.to_h.merge(
+        interactive_search: 'true',
+        request_id: @search.request_id,
+        expanded_query: @results.expanded_query,
+      ).compact,
     )
   end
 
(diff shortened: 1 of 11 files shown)


HWallenberg (approved):

app/controllers/concerns/interactive_searchable.rb
montychristo, line 134: Is it worth adding `expanded_query` into this meta too? That keeps it available when answer validation fails and we re-render the question, so the next submit can still reuse the cached expansion.

montychristo (approved):
One minor nice-to-have but not a blocker

PR #3043, opened 2026-05-27
Use govuk_list helper

vertism:
## What:
Use govuk_list helper. 
Instances where the use is more complex, such as where partials are used, and where lists are used in a static HTML page have not been updated.

## Why:
Replaces raw HTML with standardised helper

## Ticket:
N/A

## Risk:

**Risk level:** 🟢

**Reason for rating:**
Like for like change


HWallenberg (approved):
lgtm

PR #3045, opened 2026-05-28
OTTIMP-615: Implement live issues card filtering layout

montychristo:
## What:

Implements the approved Option 1 live issues log layout for `/live_issues`:

- replaces the previous table/list presentation with responsive GOV.UK
  summary-card style issue cards
- uses `govuk_summary_card`, `govuk_summary_list`, and `govuk_details`
  for the card and filter panel markup
- adds a filter and sort panel for last-updated sorting and active/resolved
  status filters
- applies filters through GET params so filtered states are linkable and reload
  safely
- shows active filter chips for non-default sorting and selected statuses, with
  `Clear all` links
- treats `Last updated (newest)` as the baseline default sort, so it is selected
  in the panel but not shown as an active filter chip
- keeps the filter panel closed after applying filters via the normal page reload
- adds accessible filter headings, a unique page title, GOV.UK pagination
  ellipses, and safer card rendering
- handles blank recommendations, missing dates, blue links, and commodity
  disclosure/linking via route helpers
- moves new user-facing copy into I18n
- replaces controller-spec coverage with request-level coverage for the live
  issues endpoint

## Why:

OTTIMP-615 asks for the MyOTT updates/news experience to use the approved
Option 1 card display with filtering panel. This should make tariff updates
more scannable across desktop and mobile.

## Ticket:

https://transformuk.atlassian.net/browse/OTTIMP-615

## Risk:

**Risk level:** 🟠

**Reason for rating:**
This changes a trader-facing content presentation and filtering flow for live
issues. The change is scoped to `/live_issues`, uses GOV.UK components and
existing backend API data, and is covered by focused specs plus CI.

## Validation:

Latest local validation on head `6151f593a3959501491db2ef245625486ed97114`:

- focused live issue specs: `16 examples, 0 failures`
- RuboCop on touched Ruby app/spec files: no offences
- `bin/rails assets:precompile`: passed
- PR line-count check input is now 579 changed lines against the 600 threshold

Previously, before the latest review-feedback commits:

- GitHub Actions `ci`: passed
- CodeQL Advanced: passed

The latest GitHub Actions run is in progress after the line-count cleanup.

## Manual verification:

Run locally and verify `/live_issues` against the approved Option 1 screens and
review comments:

- default desktop layout shows 20 total results with 4 cards per page
- pagination shows skipped pages using an ellipsis
- opening `Filter and sort` shows newest selected by default
- the filter panel shows status checkboxes, `Apply`, and `Clear all`
- applying `Last updated (oldest)` plus `Active` shows active chips for both
  sort and status
- applying only the default newest sort does not show a default sort chip
- mobile layout stacks filter actions and card summary rows without overlap


app/views/live_issues/_commodities.html.erb
```diff
@@ -1,17 +1,19 @@
...
-<%= govuk_details(summary_text: "#{live_issue.commodities.count} commodities affected") do %>
-  <% live_issue.commodities.each do |commodity_code| %>
-    <p class="govuk-body">
-      <%= govuk_link_to(commodity_code, "/commodities/#{commodity_code}") %>
-    </p>
+  <%= govuk_details(summary_text: t('live_issues.commodities.count', count: live_issue.commodities.count)) do %>
+    <ul class="govuk-list govuk-!-margin-bottom-0 live-issues__commodity-list">
+      <% live_issue.commodities.each do |commodity_code| %>
+        <li>
+          <%= govuk_link_to(commodity_code, "/commodities/#{commodity_code}") %>
```
willfish: Can't we use the commodities_path here?

app/views/live_issues/_commodities.html.erb
```diff
@@ -1,17 +1,19 @@
...
+      <% end %>
+    </ul>
   <% end %>
-<% end %>
 <% else %>
-<p class="govuk-body">
-  <%= govuk_link_to(live_issue.commodities.first, "/commodities/#{live_issue.commodities.first}") %>
-</p>
+  <p class="govuk-body">
+    <%= govuk_link_to(live_issue.commodities.first, "/commodities/#{live_issue.commodities.first}") %>
```
willfish: And here

app/assets/stylesheets/src/_live_issues.scss
```diff
@@ -0,0 +1,66 @@
...
+}
+
+.app-c-filter-panel__content {
+  background: govuk-colour("black", $variant: "tint-95");
+  padding: govuk-spacing(4);
+}
+
+.live-issues__filter-actions.govuk-button-group {
+  .govuk-button {
+    margin-bottom: 24px;
```
willfish: Is it not better to use the screen-adapting govuk-spacing, here, rather than a fixed value?

```suggestion
    margin-bottom: govuk-spacing(4);
```

app/assets/stylesheets/src/_live_issues.scss
```diff
@@ -0,0 +1,66 @@
...
+  }
+}
+
+@include govuk-media-query($from: tablet) {
+  .live-issues__filter-actions.govuk-button-group {
+    .govuk-button {
+      margin-bottom: 0;
+      margin-right: 24px;
```
willfish: And here

app/controllers/live_issues_controller.rb
```diff
@@ -1,15 +1,37 @@
 class LiveIssuesController < ApplicationController
+  LIVE_ISSUES_PER_PAGE = 4
```
willfish, line 2: Should this be 20 as per the PR or is the PR description incorrect?

spec/controllers/live_issues_controller_spec.rb
```diff
@@ -2,10 +2,10 @@
   subject(:perform_request) { get :index, params: params }
```
willfish: Can we use a request spec instead, here, or migrate in a separate PR?

app/assets/stylesheets/src/_live_issues.scss
```diff
@@ -0,0 +1,66 @@
+.live-issues__result-count {
```
willfish, line 1: If we're adding an MoJ/other gov service component (with this details component filtering) I'm wondering if we're not better off having this file be more abstract to the component and then adding a govuk helper method for the new component. I assume the idea was stolen from, here: https://design-patterns.service.justice.gov.uk/components/filter/, https://design-patterns.service.justice.gov.uk/patterns/filter-a-list/?

We could add a method like:

```ruby
  def app_filter(title:, action:, result_count: nil, clear_path: nil, selected_filters: [], &block)
    render(
      'shared/components/filter',
      title:,
      action:,
      result_count:,
      clear_path:,
      selected_filters:,
      form_content: capture(&block),
    )
  end
```

```erb
  <details class="app-c-filter govuk-details">
    <summary class="govuk-details__summary">
      <span class="govuk-details__summary-text"><%= title %></span>
      <% if result_count.present? %>
        <span class="app-c-filter__count"><%= result_count %></span>
      <% end %>
    </summary>

    <div class="govuk-details__text app-c-filter__content">
      <% if selected_filters.present? %>
        <div class="app-c-selected-filters">
          <h2 class="govuk-heading-s">Selected filters</h2>

          <ul class="govuk-list app-c-selected-filters__list">
            <% selected_filters.each do |filter| %>
              <li>
                <%= link_to filter[:remove_path], class: 'app-c-selected-filters__tag govuk-link' do %>
                  <span aria-hidden="true">×</span>
                  <span class="govuk-visually-hidden">Remove this filter: </span>
                  <%= filter[:label] %>
                <% end %>
              </li>
            <% end %>
          </ul>

          <% if clear_path.present? %>
            <%= govuk_link_to 'Clear all filters', clear_path, no_visited_state: true %>
          <% end %>
        </div>
      <% end %>

      <%= form_with url: action, method: :get, local: true do %>
        <%= form_content %>
        <%= submit_tag 'Apply filters', class: 'govuk-button' %>
      <% end %>
    </div>
  </details>
```

app/views/live_issues/_card.html.erb
```diff
@@ -0,0 +1,43 @@
+<div class="govuk-summary-card govuk-!-margin-bottom-4 live-issues__card">
```
vertism: It would be good to use the govuk helper for summary cards - https://govuk-components.x-govuk.org/components/summary-list/#summary-cards

app/views/live_issues/_filters.html.erb
```diff
@@ -0,0 +1,67 @@
+<details class="govuk-details govuk-!-margin-bottom-4 live-issues__filter-panel">
```
vertism: You can use the [govuk_details](https://govuk-components.x-govuk.org/components/details/) helper

app/controllers/live_issues_controller.rb
montychristo: Ah yes, slight PR wording ambiguity. It should be 20 total results, with 4 cards per page, which is why pagination shows 1 … 5. I’ve updated the PR description/manual verification

app/assets/stylesheets/src/_live_issues.scss
montychristo: Agree the filter pattern could be reusable. For this PR I’ve kept it scoped to live issues and switched the markup to `govuk_details`, because extracting a new shared app filter helper/component feels like a broader scope than this ticket. 

How about we extract it if we end up getting a confirmed second requirement for this?

willfish: The problem is - no one will reach for a specific implementation if they come to use this filter pattern again. They'll probably just recreate and forget about the duplication

willfish (approved):

vertism (approved):

app/assets/stylesheets/src/_live_issues.scss
montychristo: Fair point - only concern I have is the reusable component may push this PR over the 600 line limit.

I suggest we merge this, and then I create a followup PR to refactor this out into a reusable shared component?

PR #3048, opened 2026-05-28
Revert Chapter headings table conversion

vertism:
## What:
Converting this table to use govuk_table was causing issues where a partial was being used to render each heading.
This change has been reverted.
Also fixes a use of govuk_link_to that had arguments the wrong way round

## Why:
Fixes a rendering issue

## Ticket:
N/A

## Risk:

**Risk level:** 🟢

**Reason for rating:**
Fixes a bug


HWallenberg (approved):
lgtm 🙌

PR #3056, opened 2026-06-01
BAU: Handle news item not-found responses

willfish:
### Jira link

BAU

### What?

- [x] Add request coverage for backend not-found responses on news index and show pages.
- [x] Rescue `Faraday::ResourceNotFound` in `NewsItemsController`.
- [x] Redirect news not-found cases to the existing not-found route.

### Why?

Production logs showed frontend requests where backend 404s for news item paths were still being surfaced through exception handling. This makes expected not-found responses explicit in the owning controller.

### Risk

**Risk level:** 🟢

**Reason for rating:** Only handles backend not-found exceptions in the news controller and uses the existing not-found route.

### Validation

- [x] `bundle exec rspec spec/requests/news_items_controller_spec.rb`

### DIFF
--- a/app/controllers/news_items_controller.rb
+++ b/app/controllers/news_items_controller.rb
@@ -1,6 +1,7 @@
 class NewsItemsController < ApplicationController
   before_action :disable_search_form,
                 :disable_switch_service_banner
+  rescue_from Faraday::ResourceNotFound, with: :handle_resource_not_found
 
   def index
     @news_collections = News::Collection.all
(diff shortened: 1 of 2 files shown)


willfish:
Closing this because `Faraday::ResourceNotFound` is already mapped to `:not_found` in `config/initializers/exceptions.rb`, so the explicit news controller rescue is not needed for the 500s under investigation. The sampled news 500 was caused by malformed `cookies_policy` JSON during analytics helper rendering, which is covered by #3057.

PR #3098, opened 2026-06-09
Add page title assertions for duty calculator import destination

vertism:
## What:
- Adds a request spec for the duty calculator import destination step.
- Asserts the page title on the successful GET render.
- Asserts the page title is still present when POST validation fails and the page is re-rendered.

## Why:
- Page titles are part of the user-facing behaviour in this journey.
- Gives direct coverage for a core duty calculator step without expanding controller specs.

## Ticket:
- N/A

## Risk:

**Risk level:** 🟢

**Reason for rating:**
This is test-only coverage and does not change production runtime behaviour.


vertism:
Superseded by #3097, which now includes this coverage.

PR #3112, opened 2026-06-11
Remove unused moving_date param

vertism:
## What:
Removing incorrect parameter used when calling Green Lanes API.
VCR cassette has been regenerated.

## Why:
Green Lanes API does not have a `moving_date` filter. The correct parameter is `as_of` which is already being sent.

## Ticket:
BAU

## Risk:

**Risk level:** 🟢

**Reason for rating:**
Removes unused parameter so should have no impact.


willfish (approved):
Yes!

PR #3136, opened 2026-06-22
Group Sass imports by layer

willfish:
## What:

- Groups `application.sass.scss` imports by Sass layer: vendor, settings, tools, base, GOV.UK overrides, utilities, patterns, components, and journeys.
- Moves Sass partials from the flat `src/` directory into matching layer directories.
- Moves shared feature-panel colour tokens into `settings/_colors.scss` so components do not depend on another component being imported first.
- Removes the duplicate `a-z-index` import.
- Deletes the unimported `_forms.scss` partial.
- Updates `print.sass.scss` and the step-by-step nav vendor import path for the new file locations.

## Why:

The Sass entrypoint was an append-only flat list, which made ownership and import order hard to reason about. This keeps styling ownership explicit by layer, with shared prerequisites loaded from settings before component and journey styles.

## Ticket:

N/A

## Risk:

**Risk level:** 🟢

**Reason for rating:**

Low risk. This is a Sass filesystem/import-path refactor with representative before/after visual comparisons showing no rendered change across affected surfaces.


amberstarlight (approved):
Amazing stuff 🎉

PR #3159, opened 2026-06-24
Migrate Sass imports to namespaced modules

willfish:
## What:

Migrate the application and print Sass entrypoints from deprecated `@import` usage to namespaced Sass modules with `@use`.

This also adds small Sass facade modules for shared settings/tools/GOV.UK helpers, moves print-only rules into `src/print`, and makes the previous cross-file `@extend` selectors explicit where module scoping would otherwise change the compiled output.

## Why:

Sass `@import` is deprecated and keeps all partials in a shared global scope. Moving to `@use` makes each partial's dependencies explicit, reduces order-sensitive coupling, and keeps the stylesheet structure aligned with the CSS architecture work.

## Ticket:

BAU

## Risk:

**Risk level:** 🟢

**Reason for rating:**

This is a Sass source-structure refactor with no intended visual, content, journey, accessibility markup, or backend behaviour change. Before/after validation covered 328/328 touched Sass selector tokens from 66 changed Sass files across desktop, mobile, and print-CSS contexts with 0 computed-style differences.


shjohnson (approved):
LGTM

PR #3161, opened 2026-06-24
Extract shared view component helpers

willfish:
## What:

Extract repeated view structures behind shared helpers for related navigation, card lists, app cards, feature panels, inset wrappers, measure insets, and downloadable rules-of-origin documents.

Ordinary GOV.UK inset variants now delegate to `govuk_inset_text`. The duty calculator measures inset remains on its existing custom wrapper.

## Why:

These CSS component structures were duplicated across templates, which made future styling cleanup dependent on editing many near-identical ERB blocks.

Putting the repeated structures behind named helpers gives each component pattern a single Rails entry point while preserving the existing markup and behaviour.

## Ticket:

BAU

## Risk:

**Risk level:** 🟢

**Reason for rating:**

This is a covered view refactor with no intended content, journey, backend, or layout behaviour change. Before/after validation rendered 36/36 changed view surfaces on the stacked base branch and helper branch, with the same 415-example suite passing on both.


shjohnson (approved):
LGTM

PR #3170, opened 2026-06-26
Revert "HMRC-2401 Preserve MyOTT subscription return URL"

rasikasri:
Reverts trade-tariff/trade-tariff-frontend#3130


willfish (approved):
Worth a try

PR #3187, opened 2026-07-01
feat: feature flag opt-in page

neilmiddleton:
## What:
Adds a hidden feature flag opt-in page at `/feature-flags` where internal users can enable experimental features for their browser session.

## Why:
Allows the team to test Flagsmith-backed features in deployed environments without manually editing identity overrides. Users visit the page, see the features registered as opt-in capable in frontend config, and toggle them on/off.

Opting in writes a trait to the current anonymous Flagsmith identity. The page reads and writes through the Flagsmith core API (`http://flagsmith.tariff.internal:8000`) using the existing `FLAGSMITH_ENVIRONMENT_KEY`, so it does not need a separate management token.

## Ticket:
N/A

## Risk:

**Risk level:** 🟢

**Reason for rating:**
Purely additive — new route, controller, service, and view. No changes to any existing flag evaluation path or user-facing journey. The page is effectively hidden unless Flagsmith is configured for the environment.

### DIFF
--- a/app/controllers/feature_flags_controller.rb
+++ b/app/controllers/feature_flags_controller.rb
@@ -1,0 +1,41 @@
+class FeatureFlagsController < ApplicationController
+  before_action :require_management_client
+
+  def index
+    flags = Current.flagsmith_flags ||= FlagsmithClient.instance.get_flags_for(Current.flagsmith_identity)
+
+    @optin_features = flags.all_flags
+      .select { |f| f.feature_name.end_with?('_optin') && f.enabled? }
+      .map { |optin_flag|
+        feature_name = optin_flag.feature_name.delete_suffix('_optin')
+        { name: feature_name, enabled: flags.get_flag(feature_name).enabled? }
+      }
+      .sort_by { |f| f[:name] }
+  rescue StandardError => e
+    Rails.logger.error("FeatureFlagsController#index: Flagsmith unavailable: #{e.class}: #{e.message}")
+    @optin_features = []
+    flash.now[:alert] = 'Feature flags could not be loaded. Flagsmith may be unavailable.'
+  end
+
+  def update
+    flag_name = params[:id]
+    enabled = params[:enabled] == 'true'
+
+    FlagsmithManagementClient.instance.set_trait(
+      Current.flagsmith_identity.identifier,
+      flag_name,
+      enabled,
+    )
+
+    redirect_to feature_flags_path, notice: "#{flag_name} #{enabled ? 'enabled' : 'disabled'}."
+  rescue StandardError => e
+    Rails.logger.error("FeatureFlagsController#update: failed to set trait #{params[:id]}: #{e.class}: #{e.message}")
+    redirect_to feature_flags_path, alert: 'Could not save your preference. Flagsmith may be unavailable.'
+  end
+
+  private
+
+  def require_management_client
+    render plain: 'Not found', status: :not_found unless FlagsmithManagementClient.configured?
+  end
+end
--- a/app/controllers/feature_flags_controller.rb
+++ b/app/controllers/feature_flags_controller.rb
@@ -3,7 +3,7 @@ class FeatureFlagsController < ApplicationController
   before_action :disable_search_form
 
   def index
-    flags = Current.flagsmith_flags ||= FlagsmithClient.instance.get_flags_for(Current.flagsmith_identity)
+    flags = FlagsmithManagementClient.instance.get_flags_for(Current.flagsmith_identity)
 
     @optin_features = optin_flag_names.map { |name|
       { name: name, enabled: flags.get_flag(name).enabled? }
--- a/app/controllers/feature_flags_controller.rb
+++ b/app/controllers/feature_flags_controller.rb
@@ -1,16 +1,13 @@
 class FeatureFlagsController < ApplicationController
   before_action :require_management_client
+  before_action :disable_search_form
 
   def index
     flags = Current.flagsmith_flags ||= FlagsmithClient.instance.get_flags_for(Current.flagsmith_identity)
 
-    @optin_features = flags.all_flags
-      .select { |f| f.feature_name.end_with?('_optin') && f.enabled? }
-      .map { |optin_flag|
-        feature_name = optin_flag.feature_name.delete_suffix('_optin')
-        { name: feature_name, enabled: flags.get_flag(feature_name).enabled? }
-      }
-      .sort_by { |f| f[:name] }
+    @optin_features = optin_flag_names.map { |name|
+      { name: name, enabled: flags.get_flag(name).enabled? }
+    }.sort_by { |f| f[:name] }
   rescue StandardError => e
     Rails.logger.error("FeatureFlagsController#index: Flagsmith unavailable: #{e.class}: #{e.message}")
     @optin_features = []
@@ -19,9 +16,8 @@ class FeatureFlagsController < ApplicationController
 
   def update
     flag_name = params[:id]
-    flags = Current.flagsmith_flags ||= FlagsmithClient.instance.get_flags_for(Current.flagsmith_identity)
 
-    unless flags.get_flag("#{flag_name}_optin").enabled?
+    unless optin_flag_names.include?(flag_name)
       return redirect_to feature_flags_path, alert: 'That feature is not available for opt-in.'
     end
 
@@ -43,5 +39,12 @@ class FeatureFlagsController < ApplicationController
 
   def require_management_client
     render plain: 'Not found', status: :not_found unless FlagsmithManagementClient.configured?
+  end
+
+  def optin_flag_names
+    TradeTariffFrontend::Config.registered_flags
+      .values
+      .select { |f| f[:optin] }
+      .map { |f| f[:name] }
   end
 end
--- a/app/controllers/feature_flags_controller.rb
+++ b/app/controllers/feature_flags_controller.rb
@@ -29,6 +29,9 @@ class FeatureFlagsController < ApplicationController
       enabled,
     )
 
+    session[:flagsmith_optin_traits] ||= {}
+    session[:flagsmith_optin_traits][flag_name] = enabled
+
     redirect_to feature_flags_path, notice: "#{flag_name.humanize} #{enabled ? 'enabled' : 'disabled'}."
   rescue StandardError => e
     Rails.logger.error("FeatureFlagsController#update: failed to set trait #{params[:id]}: #{e.class}: #{e.message}")
--- a/app/controllers/feature_flags_controller.rb
+++ b/app/controllers/feature_flags_controller.rb
@@ -19,6 +19,12 @@ class FeatureFlagsController < ApplicationController
 
   def update
     flag_name = params[:id]
+    flags = Current.flagsmith_flags ||= FlagsmithClient.instance.get_flags_for(Current.flagsmith_identity)
+
+    unless flags.get_flag("#{flag_name}_optin").enabled?
+      return redirect_to feature_flags_path, alert: 'That feature is not available for opt-in.'
+    end
+
     enabled = params[:enabled] == 'true'
 
     FlagsmithManagementClient.instance.set_trait(
--- a/app/controllers/feature_flags_controller.rb
+++ b/app/controllers/feature_flags_controller.rb
@@ -29,7 +29,7 @@ class FeatureFlagsController < ApplicationController
       enabled,
     )
 
-    redirect_to feature_flags_path, notice: "#{flag_name} #{enabled ? 'enabled' : 'disabled'}."
+    redirect_to feature_flags_path, notice: "#{flag_name.humanize} #{enabled ? 'enabled' : 'disabled'}."
   rescue StandardError => e
     Rails.logger.error("FeatureFlagsController#update: failed to set trait #{params[:id]}: #{e.class}: #{e.message}")
     redirect_to feature_flags_path, alert: 'Could not save your preference. Flagsmith may be unavailable.'
--- a/app/services/flagsmith_management_client.rb
+++ b/app/services/flagsmith_management_client.rb
@@ -1,13 +1,36 @@
-# Thin wrapper for writing identity traits directly to the Flagsmith core API.
+# Wrapper for the Flagsmith core API — writes identity traits and reads
+# identity flags directly from the core server.
 #
-# Uses the SDK trait endpoint (POST /api/v1/traits/) with the X-Environment-Key
-# header — identical schema to the standard SDK, but sent to the Flagsmith core
-# server via Cloud Map (http://flagsmith.tariff.internal:8000) rather than the
-# Edge Proxy. The Edge Proxy is read-through and does not persist trait writes.
-#
-# Only set_trait is implemented — the read path uses the existing FlagsmithClient.
+# Both reads and writes use the SDK endpoints with the X-Environment-Key header,
+# sent to the Flagsmith core server via Cloud Map
+# (http://flagsmith.tariff.internal:8000) rather than the Edge Proxy.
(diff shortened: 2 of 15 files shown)


app/controllers/feature_flags_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+    Rails.logger.error("FeatureFlagsController#index: Flagsmith unavailable: #{e.class}: #{e.message}")
+    @optin_features = []
+    flash.now[:alert] = 'Feature flags could not be loaded. Flagsmith may be unavailable.'
+  end
+
+  def update
+    flag_name = params[:id]
+    enabled = params[:enabled] == 'true'
+
+    FlagsmithManagementClient.instance.set_trait(
```
willfish, line 26: This lets anyone PATCH any trait key. Need to re-check the `_optin` allowlist here.

config/initializers/flagsmith.rb
```diff
@@ -1,30 +1,41 @@
...
 if flagsmith_environment_key.present? && flagsmith_api_url.present?
   Rails.application.config.to_prepare do
     FlagsmithClient.configure(
       environment_key: flagsmith_environment_key,
       api_url: flagsmith_api_url,
     )
+
+    if flagsmith_management_api_token.present?
+      FlagsmithManagementClient.configure(
+        api_url: flagsmith_api_url,
```
willfish: This wires writes to the edge URL. Edge Proxy does not persist traits.

app/services/flagsmith_management_client.rb
```diff
@@ -0,0 +1,42 @@
...
+      @instance = new(api_url:, api_token:)
+    end
+  end
+
+  def initialize(api_url:, api_token:)
+    @connection = Faraday.new(url: api_url.to_s.delete_suffix('/')) do |f|
+      f.request :json
+      f.response :json
+      f.options.timeout = 2
+      f.headers['Authorization'] = "Token #{api_token}"
```
willfish: This looks like the wrong auth scheme. Admin API wants `Authorization: Api-Key ...`.

app/services/flagsmith_management_client.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+  end
+
+  # Set a trait on the given Flagsmith identity.
+  # trait_value should be a boolean, string, integer, or float.
+  def set_trait(identifier, trait_key, trait_value)
+    @connection.post('/api/v1/traits/', {
```
willfish, line 75: Faraday will not raise on 4xx/5xx here. We need to check the response before showing success.

app/views/feature_flags/index.html.erb
```diff
@@ -0,0 +1,66 @@
+<% content_for :title, "Feature flags | #{default_title}" %>
+
+<div class="govuk-width-container">
+  <main class="govuk-main-wrapper" id="main-content" role="main">
```
willfish: The layout already has `<main>`. This adds a second main landmark.

app/views/feature_flags/index.html.erb
```diff
@@ -0,0 +1,66 @@
...
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+        <h1 class="govuk-heading-xl">Feature flags</h1>
+
+        <p class="govuk-body">
+          Opt into experimental features for your current browser session.
+          These features are under development and may change or be removed.
+        </p>
+
+        <% if flash[:notice] %>
```
willfish: Shared flashes already render this. This duplicates the banner and ID.

app/controllers/feature_flags_controller.rb
neilmiddleton: The `update` action now fetches the flags collection and checks that `{flag_name}_optin` is enabled before writing the trait — any name not in the allowlist gets a redirect with an error notice and no trait is written.

config/initializers/flagsmith.rb
neilmiddleton: Added FLAGSMITH_MANAGEMENT_API_URL as a separate env var pointing at the core API server — the management client now uses that URL instead of the Edge Proxy URL, which doesn't persist traits.

app/services/flagsmith_management_client.rb
neilmiddleton: Changed to `Api-Key` — the header is now `Authorization: Api-Key <token>`.

app/services/flagsmith_management_client.rb
neilmiddleton: Added `f.response :raise_error` to the Faraday stack so 4xx/5xx responses raise `Faraday::Error`, which the controller rescues and redirects with an alert.

app/views/feature_flags/index.html.erb
neilmiddleton: Removed the duplicate `<main>` wrapper — the layout already provides it at line 81 of application.html.erb.

app/views/feature_flags/index.html.erb
neilmiddleton: Removed the flash blocks from the view — the layout renders `shared/flashes` at line 82 already.

app/services/flagsmith_management_client.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def initialize(api_url:, api_token:)
+    @connection = Faraday.new(url: api_url.to_s.delete_suffix('/')) do |f|
+      f.request :json
+      f.response :json
+      f.options.timeout = 2
+      f.headers['Authorization'] = "Token #{api_token}"
+    end
+  end
```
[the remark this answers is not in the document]
neilmiddleton: The management client now uses a separate `FLAGSMITH_MANAGEMENT_API_URL` env var (not the Edge Proxy URL), so there is no path doubling. The `f.response :raise_error` middleware is also in place in the current code, so non-2xx responses raise `Faraday::Error`.

app/controllers/feature_flags_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+      Current.flagsmith_identity.identifier,
+      flag_name,
+      enabled,
+    )
```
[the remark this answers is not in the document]
neilmiddleton: The `update` action checks that `#{flag_name}_optin` is enabled before writing any trait — if the optin flag is not enabled the request redirects with an error notice and no trait is written.

config/initializers/flagsmith.rb
```diff
@@ -1,30 +1,45 @@
...
 #
-# Wrapped in to_prepare so the autoloaded FlagsmithClient constant resolves
-# correctly: initializers run before eager loading, and referencing an app/
-# constant at the top level here raises NameError during asset precompilation.
+# Wrapped in to_prepare so the autoloaded constants resolve correctly:
+# initializers run before eager loading, and referencing an app/ constant at
+# the top level here raises NameError during asset precompilation.
 flagsmith_environment_key = ENV['FLAGSMITH_ENVIRONMENT_KEY']
 flagsmith_api_url = TradeTariffFrontend.flagsmith_api_url
+flagsmith_management_api_url = ENV['FLAGSMITH_MANAGEMENT_API_URL']
```
willfish: I don't think this env var exists anywhere. Can we avoid adding it?

config/initializers/flagsmith.rb
```diff
@@ -1,30 +1,45 @@
...
 if flagsmith_environment_key.present? && flagsmith_api_url.present?
   Rails.application.config.to_prepare do
     FlagsmithClient.configure(
       environment_key: flagsmith_environment_key,
       api_url: flagsmith_api_url,
     )
+
+    if flagsmith_management_api_url.present? && flagsmith_management_api_token.present?
+      FlagsmithManagementClient.configure(
+        api_url: flagsmith_management_api_url,
```
willfish: This can be the fixed Cloud Map URL: `http://flagsmith.tariff.internal:8000`. No need to configure it per env.

app/services/flagsmith_management_client.rb
```diff
@@ -0,0 +1,44 @@
...
+      f.options.timeout = 2
+      f.headers['Authorization'] = "Api-Key #{api_token}"
+    end
+  end
+
+  # Set a trait on the given Flagsmith identity.
+  # trait_value should be a boolean, string, integer, or float.
+  # Raises Faraday::Error on non-2xx responses.
+  def set_trait(identifier, trait_key, trait_value)
+    @connection.post('/api/v1/traits/', {
```
willfish, line 75: This is still the SDK trait endpoint. Admin token needs an Admin identity-trait endpoint.

willfish (commented):
Couple of config/API bits still need tidying.

config/initializers/flagsmith.rb
neilmiddleton: Removed `FLAGSMITH_MANAGEMENT_API_URL` — the management client now hardcodes `http://flagsmith.tariff.internal:8000` (Cloud Map) as a constant inside the class.

config/initializers/flagsmith.rb
neilmiddleton: Done — the URL is now a `MANAGEMENT_API_URL` constant set to `http://flagsmith.tariff.internal:8000` in the class, with no env var override needed.

app/services/flagsmith_management_client.rb
neilmiddleton: Fixed — the endpoint is now `/api/v1/environments/{environment_key}/traits/` (the admin identity-trait endpoint), constructed using the SDK environment key passed to `configure`.

app/services/flagsmith_management_client.rb
```diff
@@ -0,0 +1,52 @@
...
+      f.options.timeout = 2
+      f.headers['Authorization'] = "Api-Key #{api_token}"
+    end
+  end
+
+  # Set a trait on the given Flagsmith identity.
+  # trait_value should be a boolean, string, integer, or float.
+  # Raises Faraday::Error on non-2xx responses.
+  def set_trait(identifier, trait_key, trait_value)
+    @connection.post("/api/v1/environments/#{@environment_key}/traits/", {
```
willfish: This should be the SDK trait endpoint on core, not edge and not Admin API. `POST /api/v1/traits/` with `X-Environment-Key` matches the schema and persists via core.

willfish (commented):
One more clarification on the Flagsmith endpoint.

app/services/flagsmith_management_client.rb
neilmiddleton: Switched to `POST /api/v1/traits/` with `X-Environment-Key` on `http://flagsmith.tariff.internal:8000` (Cloud Map core). Also removed `FLAGSMITH_MANAGEMENT_API_TOKEN` — the management client now shares the SDK environment key and is configured alongside `FlagsmithClient`.

app/controllers/feature_flags_controller.rb
```diff
@@ -0,0 +1,47 @@
+class FeatureFlagsController < ApplicationController
+  before_action :require_management_client
+
+  def index
+    flags = Current.flagsmith_flags ||= FlagsmithClient.instance.get_flags_for(Current.flagsmith_identity)
+
+    @optin_features = flags.all_flags
+      .select { |f| f.feature_name.end_with?('_optin') && f.enabled? }
```
willfish: Can we avoid the `_optin` companion flag here? I’d use registered Flagsmith features with an opt-in marker, then toggle the trait for that same feature name. Something like `flagsmith_flag :interactive_search_enabled?, name: :interactive_search, allow_optin: true` would keep the opt-in config next to the registered flag.

app/controllers/feature_flags_controller.rb
```diff
@@ -0,0 +1,47 @@
+class FeatureFlagsController < ApplicationController
+  before_action :require_management_client
```
willfish, line 2: Can we hide the shared search and service switcher on this page? Looks noisy here; existing pages call `disable_search_form` and `disable_switch_service_banner` for this.

app/controllers/feature_flags_controller.rb
neilmiddleton: Replaced the `_optin` companion flag with an `optin:` metadata option on `flagsmith_flag` registrations (e.g. `flagsmith_flag :interactive_search_enabled?, name: :interactive_search, services: %i[uk], optin: true`). The controller now reads `TradeTariffFrontend::Config.registered_flags` to derive the opt-in list and toggles the same flag name as a trait — no companion flags in Flagsmith needed.

app/controllers/feature_flags_controller.rb
neilmiddleton: Added `before_action :disable_search_form` to the controller.

trade-tariff-infrastructure-continuity (dismissed):

app/controllers/feature_flags_controller.rb
willfish: Amazing

app/controllers/feature_flags_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+  before_action :disable_search_form
+
+  def index
+    flags = Current.flagsmith_flags ||= FlagsmithClient.instance.get_flags_for(Current.flagsmith_identity)
```
willfish: For this page we should read these flags from core too. At the moment we write traits to core, then read state back via the existing Edge-backed `FlagsmithClient`, so the redirect can say success while still showing the old value.

willfish: image

neilmiddleton: Added `get_flags_for` to `FlagsmithManagementClient` so the index action reads flags from the core API (same Faraday connection used for trait writes) rather than the Edge Proxy. `Current.flagsmith_flags` is no longer used on this page, so the Edge-backed cache is unaffected.

trade-tariff-infrastructure-continuity (dismissed):

app/controllers/feature_flags_controller.rb
neilmiddleton: Good spot — the raw flag name was leaking into the flash notice. Fixed with `.humanize` so it now reads "Interactive search enabled." instead of "interactive_search enabled."

trade-tariff-infrastructure-continuity (dismissed):

app/controllers/feature_flags_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+    flag_name = params[:id]
+
+    unless optin_flag_names.include?(flag_name)
+      return redirect_to feature_flags_path, alert: 'That feature is not available for opt-in.'
+    end
+
+    enabled = params[:enabled] == 'true'
+
+    FlagsmithManagementClient.instance.set_trait(
+      Current.flagsmith_identity.identifier,
```
willfish, line 27: Writing to core is right. The missing bit is the read path: the app evaluates through Edge without passing the current opt-in trait value, so guided search stays disabled after opt-in.

```ruby
traits = FlagsmithManagementClient.instance.traits_for(Current.flagsmith_identity)
  .slice(:interactive_search)

@sdk_client.get_identity_flags(
  identity.identifier,
  false,
  **traits,
)
```

Or equivalent: read the current identity traits from core, then pass those values into `FlagsmithClient#get_flags_for` / `get_identity_flags` when evaluating via Edge.

neilmiddleton: Good catch. Writing to core was right but the Edge Proxy never saw the trait on subsequent requests. Fixed by storing opt-in choices in the session after each write and passing them as explicit traits to `get_identity_flags` — the Edge Proxy then applies them immediately without waiting for its cache to sync from core (51144bb83).

trade-tariff-infrastructure-continuity (approved):

willfish (approved):
Validated in development. Disable hides guided search; enable shows it for the same browser identity.

trade-tariff-infrastructure-continuity (approved):

PR #3188, opened 2026-07-02
feat: default all feature flags to enabled when Flagsmith is unconfigured

neilmiddleton:
## What:
Replace the `raise` in `FlagsmithClient.instance` with a `NullClient` that reports every flag as enabled. When `FLAGSMITH_ENVIRONMENT_KEY` is absent the app now boots and serves requests with all Flagsmith-registered feature flags on.

## Why:
Without Flagsmith credentials the app previously raised `RuntimeError` on the first flag evaluation (caught by `rescue StandardError`, but only in the Flagsmith-backed config path). The practical effect was that features defaulted to their Ruby-computed values, which could be `false` in some environments and made local development confusing. Defaulting all flags to enabled is the least-surprising behaviour when the flag service is absent — developers get every feature available without needing service credentials.

## Ticket:
AI-928

## Risk:

**Risk level:** 🟢

**Reason for rating:**
Purely additive change to local/unconfigured behaviour. In environments where `FLAGSMITH_ENVIRONMENT_KEY` is set the code path is identical to before — the real `FlagsmithClient` is used. The `NullClient` only activates when the key is absent, which is never the case in staging or production.

## Test commands
```
bundle exec rspec spec/services/flagsmith_client_spec.rb spec/lib/
```

### DIFF
--- a/app/services/flagsmith_client.rb
+++ b/app/services/flagsmith_client.rb
@@ -3,34 +3,21 @@
 # Evaluation (get_flags_for) uses the SDK environment key to call the
 # Flagsmith API and returns a Flagsmith::Flags::Collection.
 #
-# When Flagsmith is not configured (no FLAGSMITH_ENVIRONMENT_KEY), instance
-# returns a NullFlagsmithClient that reports all flags as enabled, so the app
-# starts and all features are on by default in local development.
+# When Flagsmith is not configured (no FLAGSMITH_ENVIRONMENT_KEY), configured?
+# returns false and FlagsmithBackedConfig falls back to each flag's own
+# Ruby-computed default rather than attempting an API call.
 #
 # In tests, FlagsmithClient.instance is replaced with a TestFlagsmithClient
 # double (see spec/support/flagsmith.rb). Use FlagsmithClient.instance= to
 # inject any replacement.
 class FlagsmithClient
-  # Returned by get_flag when Flagsmith is unconfigured. Reports enabled so
-  # flag.is_default is false and flag.enabled? is true.
-  class AllEnabledFlag
-    def is_default = false
-    def enabled? = true
-  end
-
-  # Returned by get_flags_for when Flagsmith is unconfigured. All flags on.
-  class AllEnabledFlags
-    def get_flag(_name) = AllEnabledFlag.new
-  end
-
-  # Used when FLAGSMITH_ENVIRONMENT_KEY is absent. All flags enabled.
-  class NullClient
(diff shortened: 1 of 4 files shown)


config/initializers/flagsmith.rb
```diff
@@ -27,7 +27,7 @@
     )
   end
 elsif flagsmith_environment_key.blank?
-  Rails.logger.warn('Flagsmith environment key is not configured; guided search is disabled')
+  Rails.logger.warn('Flagsmith environment key is not configured; all feature flags default to enabled')
```
willfish: Shouldn't we rely on the config decorator dsl for what the default is? I suspect in some cases we'll default to on and sometimes off depending on what each flag specifies

neilmiddleton: Agreed — swapped the NullClient (which blanket-enabled everything) for a `FlagsmithClient.configured?` guard in `flagsmith_config_flag`. When the client isn't set up we now return early with each flag's own Ruby-computed default, so the per-flag DSL declarations are respected.

app/services/flagsmith_client.rb
```diff
@@ -3,13 +3,34 @@
...
+# starts and all features are on by default in local development.
+#
 # In tests, FlagsmithClient.instance is replaced with a TestFlagsmithClient
 # double (see spec/support/flagsmith.rb). Use FlagsmithClient.instance= to
 # inject any replacement.
 class FlagsmithClient
+  # Returned by get_flag when Flagsmith is unconfigured. Reports enabled so
+  # flag.is_default is false and flag.enabled? is true.
+  class AllEnabledFlag
+    def is_default = false
```
morganism, line 17: why not  the more Ruby-like way ? which is in line with 'enabled?'

def default? = false

neilmiddleton: The `AllEnabledFlag` class and its `is_default` method were removed entirely as part of addressing the previous review — we replaced the NullClient approach with a `FlagsmithClient.configured?` guard, so there's nothing left here to rename.

willfish (approved):
Approved. I checked the final diff and the configured production path still uses the real Flagsmith client; the new fallback only applies when Flagsmith is unconfigured. I also verified the focused specs on the PR ref:

`BUNDLE_PATH=.bundle bundle exec rspec spec/services/flagsmith_client_spec.rb spec/lib/trade_tariff_frontend_spec.rb spec/models/search_spec.rb spec/controllers/application_controller_spec.rb`

Result: 105 examples, 0 failures.

PR #3190, opened 2026-07-02
refactor: extract shared pagination concern from changes controllers

neilmiddleton:
## What:
Extract the duplicated `handle_next_page` method from `GreenLanes::MovingRequirementsController` and `GreenLanes::ApplicableExemptionsController` into a shared concern: `GreenLanes::Concerns::NextPageNavigation`.

## Why:
Both controllers contained an identical structural method that parses a wizard next-page URL, merges controller-specific query params, and appends a cache-busting timestamp. Duplicated logic creates a maintenance hazard — a change to the URL-building logic would need to be applied in two places. The concern gives this a single home.

The `change_path` methods in the three `ChangesController` subclasses were also listed as candidates, but those are intentional polymorphic template method overrides of the abstract base class — not duplication — so they were left as-is.

## Ticket:
BAU

## Risk:

**Risk level:** 🟢

**Reason for rating:** Pure refactor. No behaviour change — the URI parsing, param merging, and timestamp logic are identical to what was in both controllers. Full test suite (974 examples) passes green.

### DIFF
--- a/app/controllers/green_lanes/applicable_exemptions_controller.rb
+++ b/app/controllers/green_lanes/applicable_exemptions_controller.rb
@@ -3,6 +3,7 @@ module GreenLanes
     include GreenLanesHelper
 
     include Concerns::ExpirableUrl
+    include Concerns::NextPageNavigation
 
     before_action :disable_switch_service_banner,
                   :disable_search_form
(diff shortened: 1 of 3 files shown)


HWallenberg (approved):
lgtm

PR #3191, opened 2026-07-02
refactor: extract ApplicationController responsibilities into concerns

neilmiddleton:
## What:
Extract four ActiveSupport::Concern modules from `ApplicationController` to give each responsibility its own file. The controller shrinks from 269 lines to ~140 lines of tightly-scoped code.

## Why:
`ApplicationController` was a catch-all that made it hard to locate and reason about individual responsibilities. Separating them into concerns makes each piece easier to find, test, and change in isolation.

## Ticket:
BAU

## Risk:

**Risk level:** 🟠

**Reason for rating:**
Pure extraction with no logic changes, but this touches the base controller inherited by every controller in the app. All 974 request/controller specs pass.

## Concerns created

| File | Responsibility |
|------|----------------|
| `app/controllers/concerns/cache_control.rb` | `set_cache` — sets the three no-cache response headers |
| `app/controllers/concerns/flagsmith_setup.rb` | Anonymous Flagsmith identity cookie creation and `Current.flagsmith_identity` assignment |
| `app/controllers/concerns/bot_protection.rb` | `bots_no_index_if_historical` — sets `X-Robots-Tag: none` for historical search dates |
| `app/controllers/concerns/error_handling.rb` | All seven `rescue_from` handlers and their supporting private methods |

`before_action` and `rescue_from` declarations remain in `ApplicationController` in their original order to preserve callback sequencing.

## Test commands

```sh
bundle exec rspec spec/requests/ spec/controllers/ --format progress
# 974 examples, 0 failures
```


app/controllers/application_controller.rb
```diff
@@ -6,6 +6,11 @@ class ApplicationController < ActionController::Base
   include CacheHelper
   include TradeTariffFrontend::Config::RegisteredFlags
 
+  include CacheControl
```
willfish, line 9: This is an excellent pattern 👍

willfish (dismissed):
Bit of a lint failure but otherwise this is awesome

willfish (approved):

app/controllers/application_controller.rb
neilmiddleton: Thanks! No changes needed — glad the pattern reads well.

PR #3192, opened 2026-07-02
refactor: replace method_missing with explicit delegate in presenters

neilmiddleton:
## What:
Remove `method_missing` delegation from three presenter files and replace with explicit `delegate` declarations or (where already redundant) simply remove the overrides.

## Why:
`method_missing` delegation kills IDE autocomplete, makes debugging harder, and requires a `rubocop:disable Style/MissingRespondToMissing` comment on each file. Explicit delegation makes the contract clear at a glance.

- **ChapterPresenter** and **SectionPresenter** both inherit from `TradeTariffFrontend::Presenter < SimpleDelegator`. `SimpleDelegator` already forwards all unknown methods to the wrapped object automatically, so the `method_missing` overrides were entirely redundant. They have been removed.
- **ChangePresenter** does not use `SimpleDelegator`, so its `method_missing` has been replaced with `delegate :id, :operation, :operation_date, to: :change` — the three `Change` attributes that were actually accessed via delegation.

## Ticket:
BAU

## Risk:

**Risk level:** 🟢

**Reason for rating:**
Pure internal refactor with no behaviour change. Full test suite (1382 examples) passes. No user-facing output affected.

## Test commands:
```
bundle exec rails assets:precompile
bundle exec rspec spec/presenters/ spec/views/ spec/features/ --format progress
```

### DIFF
--- a/app/presenters/change_presenter.rb
+++ b/app/presenters/change_presenter.rb
@@ -2,6 +2,7 @@ class ChangePresenter
   attr_reader :change
 
   delegate :record, to: :change, prefix: true
+  delegate :id, :operation, :operation_date, to: :change
 
   def initialize(change)
     @change = change
@@ -30,11 +31,4 @@ class ChangePresenter
 
   def anchor_link; end
 
-  private
-
-  # rubocop:disable Style/MissingRespondToMissing
-  def method_missing(*args, &block)
-    @change.send(*args, &block)
-  end
-  # rubocop:enable Style/MissingRespondToMissing
 end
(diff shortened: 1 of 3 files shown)


app/presenters/change_presenter.rb
```diff
@@ -2,6 +2,7 @@ class ChangePresenter
   attr_reader :change
 
   delegate :record, to: :change, prefix: true
```
willfish, line 4: I'm actually curious about who even uses this implementation since AFAIK its using a largely dead change implementation

willfish (approved):

app/presenters/change_presenter.rb
HWallenberg: last commit was 3 years ago, IDK what this is for either. https://github.com/trade-tariff/trade-tariff-frontend/commit/60fdb54627eb1fd3c25d0280d98b0f74311b398a

PR #3199, opened 2026-07-07
HMRC-2326: Route public search between classic and internal implement…

rasikasri:
## What:
Integrate non-interactive internal search results with classic search, using FlagSmith to control user routing.

## Why:
The public search experiment needs internal search to return direct results so it can be compared with classic search.

## Ticket:
[HMRC-2326](https://transformuk.atlassian.net/browse/HMRC-2326)

## Risk:

**Risk level:** 🟠

**Reason for rating:**

- This change commodity data presentation slightly for the users when flag is turned on
- Use modified API calls to the backend to route search to internal search


app/controllers/concerns/hybrid_searchable.rb
```diff
@@ -0,0 +1,25 @@
...
+    @results = @search.perform
+
+    respond_to do |format|
+      format.html { route_hybrid_results }
+      format.json { render json: SearchPresenter.new(@search, @results) }
+      format.atom
+    end
+  end
+
+  def route_hybrid_results
```
willfish, line 16: Can we keep the existing blank-search redirect here? With the flag on, an empty query renders an empty results page instead of returning to the finder.

spec/controllers/search_controller_search_spec.rb
```diff
@@ -312,6 +312,78 @@
       end
     end
 
+    context 'with hybrid search enabled' do
```
willfish, line 315: Can we cover this in a request spec please? The ticket asks for request-level routing coverage, and it would catch the JSON and blank-search cases.

willfish (commented):
A few things to tidy before this goes in.

app/models/search/hybrid_outcome.rb
```diff
@@ -0,0 +1,52 @@
+class Search
+  class HybridOutcome
+    include ApiEntity
+
+    attr_accessor :entry, :type
+
+    def initialize(parsed_data)
+      @results = Array(parsed_data).map { |attrs| build_model(attrs) }
+      @type = exact_match? ? 'exact_match' : 'fuzzy_match'
+    end
```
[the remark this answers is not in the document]
willfish: This breaks JSON searches. The exact and fuzzy presenters expect data this outcome does not set, so exact matches error and fuzzy matches return no results. Can we preserve the existing JSON response shape here?

app/models/search/hybrid_outcome.rb
```diff
@@ -0,0 +1,52 @@
...
+      @results.first
+    end
+
+    def commodities
+      @results
+        .select { |r| r.is_a?(Commodity) && r.declarable }
+        .first(TradeTariffFrontend.hybrid_results_to_show)
+    end
+
+    delegate :any?, :none?, to: :@results
```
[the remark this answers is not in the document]
willfish: This can render a blank page when the backend only returns headings or chapters. `any?` is true but `commodities` is empty, so we show neither results nor the no-results message.

willfish:
I think once we've proven this the risk level goes down? But worth raising it in the ops meeting anyway

spec/controllers/search_controller_search_spec.rb
rasikasri: Done

app/controllers/concerns/hybrid_searchable.rb
rasikasri: Done

app/models/search/hybrid_outcome.rb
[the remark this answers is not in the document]
rasikasri: Done

PR #3267, opened 2026-07-30
fix: return 404 instead of 500 for invalid goods nomenclature code lengths

neilmiddleton:
## What:
Return a 404 (not a 500) when an invalid goods nomenclature code length reaches the URL helper or a route constraint.

Two changes:
- In `goods_nomenclature_path`, replace the implicit `else → commodity_path` with an explicit `when 10` for the commodity case and an `else` that raises `ActionController::RoutingError`. Rails middleware converts routing errors to 404.
- Add `rescue_from ActionController::UrlGenerationError, with: :raise_not_found` in `ApplicationController` as a safety net for any other route helper that may receive an invalid ID — consistent with the existing handling of `URI::InvalidURIError` and `ActionController::UnknownFormat`.

## Why:
We are seeing frequent errors of the form:

```
No route matches {action: "show", controller: "commodities", id: "104"},
possible unmatched constraints: [:id]
```

The `else` branch in `goods_nomenclature_path` called `commodity_path` for any code length that wasn't 2, 4, or 13. For a 3-digit code like `104` this triggered the `\d{10}` route constraint and raised `ActionController::UrlGenerationError`, which was not rescued and became a 500. It should be a 404.

The `GoodsNomenclatureContext` concern already validates codes as a before-action, but this change adds defence-in-depth so the app responds correctly even if an invalid code reaches a route helper by another path.

## Ticket:
BAU

## Risk:

**Risk level:** 🟢

**Reason for rating:** Error-handling only. Converts an unhandled 500 to a 404 for invalid input. No change to any commodity data path, user journey, or API call. Full test coverage added.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -23,6 +23,7 @@ class ApplicationController < ActionController::Base
 
   rescue_from Faraday::TooManyRequestsError, with: :handle_too_many_requests_error
   rescue_from URI::InvalidURIError, with: :handle_invalid_uri_error
+  rescue_from ActionController::UrlGenerationError, with: :raise_not_found
   rescue_from ActionController::UnknownFormat, with: :handle_unknown_format
   rescue_from Faraday::ConnectionFailed, with: :handle_connection_failed
   rescue_from Faraday::TimeoutError, with: :handle_timeout_error
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -23,7 +23,6 @@ class ApplicationController < ActionController::Base
 
   rescue_from Faraday::TooManyRequestsError, with: :handle_too_many_requests_error
   rescue_from URI::InvalidURIError, with: :handle_invalid_uri_error
-  rescue_from ActionController::UrlGenerationError, with: :raise_not_found
   rescue_from ActionController::UnknownFormat, with: :handle_unknown_format
   rescue_from Faraday::ConnectionFailed, with: :handle_connection_failed
   rescue_from Faraday::TimeoutError, with: :handle_timeout_error
--- a/app/helpers/goods_nomenclature_helper.rb
+++ b/app/helpers/goods_nomenclature_helper.rb
@@ -15,8 +15,10 @@ module GoodsNomenclatureHelper
       heading_path(path_opts)
     when Subheading::FULL_CODE_LENGTH
       subheading_path(path_opts)
-    else
+    when 10
       commodity_path(path_opts)
+    else
+      raise ActionController::RoutingError, "No route for goods nomenclature code #{current_goods_nomenclature_code.inspect}"
     end
   end
 
--- a/app/helpers/goods_nomenclature_helper.rb
+++ b/app/helpers/goods_nomenclature_helper.rb
@@ -15,7 +15,7 @@ module GoodsNomenclatureHelper
       heading_path(path_opts)
     when Subheading::FULL_CODE_LENGTH
       subheading_path(path_opts)
-    when 10
+    when Commodity::FULL_CODE_LENGTH
       commodity_path(path_opts)
     else
       raise ActionController::RoutingError, "No route for goods nomenclature code #{current_goods_nomenclature_code.inspect}"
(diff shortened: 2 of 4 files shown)


spec/helpers/goods_nomenclature_helper_spec.rb
```diff
@@ -74,6 +74,14 @@
     it_behaves_like 'a goods_nomenclature_path', '1901', '/headings/1901?country=AR&day=01&month=01&year=2021#export'
     it_behaves_like 'a goods_nomenclature_path', '19', '/chapters/19?country=AR&day=01&month=01&year=2021#export'
     it_behaves_like 'a goods_nomenclature_path', nil, '/find_commodity?country=AR&day=01&month=01&year=2021'
+
+    context 'when the code length does not match any known goods nomenclature type' do
+      let(:goods_nomenclature_code) { '104' }
+
+      it 'raises a routing error rather than a URL generation error' do
+        expect { helper.goods_nomenclature_path }.to raise_error(ActionController::RoutingError)
+      end
```
willfish, line 83: Fine for the helper path. If we keep the app-wide rescue, a request-level check that the page actually returns 404 would help prove it end to end.

app/controllers/application_controller.rb
```diff
@@ -23,6 +23,7 @@ class ApplicationController < ActionController::Base
 
   rescue_from Faraday::TooManyRequestsError, with: :handle_too_many_requests_error
   rescue_from URI::InvalidURIError, with: :handle_invalid_uri_error
+  rescue_from ActionController::UrlGenerationError, with: :raise_not_found
```
willfish: **This rescue is broader than the helper fix.**

It turns any `UrlGenerationError` into a soft 404, including genuine coding mistakes in other helpers/templates. Those stop showing up as 500s.

Happy with the helper change on its own. Not sure we still need this now `#3266` is on main.

app/helpers/goods_nomenclature_helper.rb
```diff
@@ -15,8 +15,10 @@ def goods_nomenclature_path(path_opts = {})
       heading_path(path_opts)
     when Subheading::FULL_CODE_LENGTH
       subheading_path(path_opts)
-    else
+    when 10
       commodity_path(path_opts)
+    else
+      raise ActionController::RoutingError, "No route for goods nomenclature code #{current_goods_nomenclature_code.inspect}"
```
willfish: **This bit is solid.** Explicit commodity length + raise is the real fix for the fallthrough.

Doesn't duplicate `#3266` — that blocks bad optional context earlier. This covers anything that still reaches the helper.

willfish (commented):
Helper change looks good. Main question is whether we still want the app-wide `UrlGenerationError` rescue after `#3266`.

spec/helpers/goods_nomenclature_helper_spec.rb
neilmiddleton: Removed — with #3266 guarding the boundary and the helper now raising `RoutingError` for unknown lengths, no request-level spec is needed here. Rails handles `RoutingError` as 404 natively.

app/controllers/application_controller.rb
neilmiddleton: Agreed and removed in the latest commit. #3266 now blocks malformed codes at the boundary, and the helper raises `RoutingError` (not `UrlGenerationError`) for unrecognised lengths — which Rails handles as 404 without an explicit rescue. Keeping the catch-all would silently swallow genuine URL helper bugs elsewhere in the app.

app/helpers/goods_nomenclature_helper.rb
```diff
@@ -15,8 +15,10 @@ def goods_nomenclature_path(path_opts = {})
       heading_path(path_opts)
     when Subheading::FULL_CODE_LENGTH
       subheading_path(path_opts)
-    else
+    when 10
       commodity_path(path_opts)
```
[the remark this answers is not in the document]
neilmiddleton: Added `Commodity::FULL_CODE_LENGTH = 10` to `app/models/commodity.rb` and updated the `when` branch accordingly — now consistent with `Chapter::SHORT_CODE_LENGTH`, `Heading::SHORT_CODE_LENGTH`, and `Subheading::FULL_CODE_LENGTH`.

app/controllers/application_controller.rb
```diff
@@ -23,6 +23,7 @@ class ApplicationController < ActionController::Base
...
   rescue_from ActionController::UnknownFormat, with: :handle_unknown_format
```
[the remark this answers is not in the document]
neilmiddleton: Removed — see the reply above. The narrowing approach (matching on error message) would work, but it's fragile against Rails internalising that string. Dropping the rescue entirely is cleaner given #3266 already handles the boundary case.

amberstarlight (approved):

PR #3282, opened 2026-08-11
fix: guided search back button freezes on 'Collecting information' throbber

HWallenberg:
## What:
Adds a bfcache (browser back-forward cache) restore handler to both `guided_search_validation_controller.js` (the initial search form) and `interactive_question_controller.js` (each guided search question). When the browser restores a page from bfcache after Back is pressed, the handler puts the visible form/header back and hides the "Collecting information..." throbber.

## Why:
Answering a guided search question hides the question form and shows a "Collecting information..." throbber client-side (via JS classes), then submits a real form POST to move to the next question. That means the page the browser freezes into its back-forward cache — the snapshot it restores when you press Back — is frozen mid-throbber, not showing the form.

Browsers normally skip re-running page JS (`connect()` etc.) on a bfcache restore, since nothing actually reloads — it's the exact same in-memory DOM being shown again. So without a handler for this specific case, that frozen throbber state was permanent: pressing Back left the user stuck with no way to answer the question or go back further, matching the ticket's report of a second Back press skipping past `/find-commodity` entirely.

The fix listens for the browser's `pageshow` event and checks `event.persisted` (true only on a bfcache restore, not a normal page load) to reset the DOM back to the form-visible state whenever this happens. This mirrors an existing pattern already used for the same class of problem in `search-autocomplete.js`.

## Ticket:
[AI-1150](https://transformuk.atlassian.net/browse/AI-1150)

## Risk:

**Risk level:** 🟠

**Reason for rating:** This changes browser back/forward navigation behaviour within the guided search journey (Search UI change, per the Amber criteria in the PR template). It's additive — a new `pageshow` listener plus DOM class toggles that only run on a bfcache restore — and doesn't touch the form submission or search request logic itself, but navigation behaviour changes are easy to get subtly wrong across browsers, so it's worth a second pair of eyes before merging.

## Test plan
- [ ] `yarn jest spec/javascript/controllers/guided_search_validation_controller_spec.js spec/javascript/controllers/interactive_question_controller_spec.js` — 30/30 passing, including new bfcache-restore cases and a case confirming a normal (non-bfcache) `pageshow` is a no-op
- [ ] Manually confirmed on a local dev server (`bin/dev`): start a guided search, answer at least one question, then press browser Back — lands back on the previous question/search page with the form visible, not the throbber. Forward navigation also confirmed working.
- [ ] Accessibility impact: none expected — no new interactive elements, only visibility toggles on existing GOV.UK components using the same `govuk-!-display-none` convention already used elsewhere on these pages

### DIFF
--- a/app/javascript/controllers/guided_search_validation_controller.js
+++ b/app/javascript/controllers/guided_search_validation_controller.js
@@ -6,6 +6,15 @@ const MIN_QUERY_LENGTH = 2
 export default class extends Controller {
   static targets = ['hiddenField', 'textarea', 'formGroup', 'formContent']
 
+  connect() {
+    this.boundRestoreFromBfcache = this.#restoreFromBfcache.bind(this)
+    window.addEventListener('pageshow', this.boundRestoreFromBfcache)
+  }
+
+  disconnect() {
+    window.removeEventListener('pageshow', this.boundRestoreFromBfcache)
+  }
+
   async validateAndSubmit(event) {
     if (this.hiddenFieldTarget.value !== 'true') return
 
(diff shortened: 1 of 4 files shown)


anupamc (approved):
LGTM

PR #3288, opened 2026-08-13
[HMRC-2644] -  Fix a 404 when clicking rules of origin proof document link

punitcse:
## What:

Fix a 404 when clicking a rules of origin proof document link, and add regression tests for it.        

 
## Why:
`RulesOfOrigin::ProofsController#index` requested schemes without including the `origin_reference_document` relation. Proof content can embed a literal `<a href="{ord_url}">...</a>` placeholder that only resolves to the real asset path when that relation is present. With it missing,  the raw `{ord_url}` string was left in the rendered link's `href`, so clicking it 404'd.

Added:                                                                                                 
  - A request spec asserting the controller requests `include: 'proofs,origin_reference_document'`, so this can't silently regress.                                                                           
  - An end-to-end request spec that renders a proof containing the real `{ord_url}` pattern, follows the rendered link, and asserts it resolves (200) rather than 404.                                          
  - A view spec asserting the `origin_reference_document.document_path` link resolves correctly.

## Ticket:

https://transformuk.atlassian.net/jira/software/c/projects/HMRC/boards/155?selectedIssue=HMRC-2644

## Risk:

**Risk level:** 🟢  

**Reason for rating:**

Additive change to an existing API call (adds one relation to an already-used `include` param) plus new test coverage; no behaviour change to any other journey.

### DIFF
--- a/app/controllers/rules_of_origin/proofs_controller.rb
+++ b/app/controllers/rules_of_origin/proofs_controller.rb
@@ -3,7 +3,7 @@ module RulesOfOrigin
     before_action :disable_switch_service_banner, :disable_search_form
 
     def index
-      @schemes = RulesOfOrigin::Scheme.all(include: 'proofs')
+      @schemes = RulesOfOrigin::Scheme.all(include: 'proofs,origin_reference_document')
     end
   end
 end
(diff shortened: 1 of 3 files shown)


HWallenberg (approved):
nice. 👍 😄

